PR #1, opened 2016-08-07
Regex stringify proposal

delvedor:
JSON by default does not support RegExp objects or strings, I propose to add `RegExp` type to the schema types.
```js
reg: {
  type: 'RegExp'
}
```
In this way *fast-json-stringify* will transform regex object in escaped strings and not in `'{}'` as *JSON.stringify* does.

Example:
```js
// Before
/"([^"]|\\")*"/
// After
'"\\"([^\\"]|\\\\\\\\\\")*\\""'

// Then
/"([^"]|\\")*"/.source === new RegExp(JSON.parse('"\\"([^\\"]|\\\\\\\\\\")*\\""')).source // true
```

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,9 @@ const stringify = fastJson({
     age: {
       description: 'Age in years',
       type: 'integer'
+    },
+    reg: {
+      type: 'string'
     }
   }
 })
@@ -41,7 +44,8 @@ const stringify = fastJson({
 console.log(stringify({
   firstName: 'Matteo',
   lastName: 'Collina',
-  age: 32
+  age: 32,
+  reg: /"([^"]|\\")*"/
 }))
 ```
 
@@ -61,8 +65,15 @@ Supported types:
  * `'boolean'`
  * `'null'`
 
-And nested ones, too.
-`Date` instances are serialized with `toISOString()`.
+And nested ones, too.  
+
+*Specific use cases:*
+
+| Instance   | Serialized as                               |
+| -----------|---------------------------------------------|
+| `Date`     | `string` <small>via `toISOString()`</small> |
+| `RegExp`   | `string`                                    |
+
 
 ## Acknowledgements
 
--- a/example.js
+++ b/example.js
@@ -17,6 +17,9 @@ const stringify = fastJson({
     },
     now: {
       type: 'string'
+    },
+    reg: {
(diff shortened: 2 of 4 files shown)


mcollina:
type: 'RegExp' is not part ot jsonschema. How about adding this behavior to type: 'string'?

delvedor:
Yeah, you are right.
But there's a little drawback:
If the variable is an instance of RegExp it's easy transform it in an escaped string, but if it's a string that represents a regex, how could you know it?

For example:
```js
// $asString function
str = /"([^"]|\\")*"/
if (str instanceof RegExp) {
  $asRegExp(str)
}
// $asString code
```
If the users pass `'"([^"]|\\\\")*"'`, that is the string that represents the above regex, the `\`will not be escaped, so this string becomes `'"\\"([^\\"]|\\\\\\")*\\""'`.
If so  `new RegExp(JSON.parse('"\\"([^\\"]|\\\\\\")*\\""').source` is not the same of the original regex.

The only solution I see here, is to escape `\\` in $asString* as well.

mcollina:
This is json-schema, it's a spec :/. So, +1 for making it a part of $asString. Open a separate issue for escaping, because I fear that might have bugs in general (for / as an example).

mcollina:
or \ maybe :).

delvedor:
Sorry, made a little mess with commit messages :P

Is better now?
If the user pass a regex object, the regex will be parsed into a string.

mcollina:
That's ok. I'll merge when I get back. You can do a git rebase to squash all commits into one.. :)

delvedor:
Done, thanks!

mcollina:
Can you please add some notes in the README about this?

delvedor:
Sure, which one you prefer?

`Date` instances are serialized with `toISOString()`.
`RegExp`instances are serialized as `string`.

or

| Instance   | Serialized as                               |
| -----------|---------------------------------------------|
| `Date`     | `string` <small>via `toISOString()`</small> |
| `RegExp`   | `string`                                    |

In my opinion the second is better, is more clean.

mcollina:
Agreed, do the 2nd.
Il giorno mar 9 ago 2016 alle 14:54 Tomas Della Vedova <
notifications@github.com> ha scritto:

PR #2, opened 2016-08-19
Required fields support

delvedor:
Added `required` field support.

Example:
```javascript

const stringify = fastJson({
  title: 'schema with required field',
  type: 'object',
  properties: {
    str: {
      type: 'string',
      required: true
    },
    num: {
      type: 'number'
    }
  }
})
```

If the required field is not provided, `fast-json-stringify` will throw an error.

Required feedback 😄

### DIFF
--- a/index.js
+++ b/index.js
@@ -202,6 +202,12 @@ function buildArray (schema, code, name) {
 function nested (laterCode, name, key, schema) {
   var code = ''
   var funcName
+  if (schema.required) {
+    code += `
+      if (!obj.hasOwnProperty('${key.slice(1)}')) {
+        throw new Error('${key} is required!')
+      }`
+  }
   const type = schema.type
   switch (type) {
     case 'null':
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -202,6 +202,12 @@ function buildArray (schema, code, name) {
 function nested (laterCode, name, key, schema) {
   var code = ''
   var funcName
+  if (schema.required) {
+    code += `
+      if (!obj.hasOwnProperty('${key.slice(1)}')) {
+        throw new Error('${key} is required!')
+      }`
+  }
```
mcollina: what happens if a key is not `required`, but it is not present?

mcollina:
Good job, one nit :)

index.js
delvedor: This: 
```javascript
const stringify = fastJson({
  title: 'schema with required field',
  type: 'object',
  properties: {
    str: {
      type: 'string',
      required: true
    },
    num: {
      type: 'number'
    }
  }
})

console.log(stringify({
  str: 'string'
}) // '{"str":"string","num":null}'
```

mcollina:
THx! :)

PR #3, opened 2016-08-22
Support for missing parameters

delvedor:
Using example.js as schema, this is the generated code:
```javascript
function $main(obj) {
      var json = '{'

      if (obj.firstName !== undefined) {
        json += '"firstName":'
        json += $asString(obj.firstName)
        json += ','
      }

      if (obj.lastName !== undefined) {
        json += '"lastName":'
        json += $asString(obj.lastName)
        json += ','
      }

      if (obj.age !== undefined) {
        json += '"age":'
        json += $asNumber(obj.age)
        json += ','
      }

      if (obj.now !== undefined) {
        json += '"now":'
        json += $asString(obj.now)
        json += ','
      } else {
        throw new Error('now is required!')
      }

      if (obj.reg !== undefined) {
        json += '"reg":'
        json += $asString(obj.reg)
      }

      json += '}'
      return json
    }
```
If a parameter is required, there will be the `else` branch.

As you can see I'm not using `obj.hasOwnProperty(prop)` for testing the existence of the key (is commented inside the code), because I've noticed a big difference of performances.
When using `obj.key !== undefined` I get around *5 million ops/sec*, instead if I use `obj.hasOwnProperty(prop)` I get *~2 million ops/sec*.

The problem is that `obj.key !== undefined` is not strict as `obj.hasOwnProperty(prop)`.
Opinions?

### DIFF
--- a/index.js
+++ b/index.js
@@ -140,9 +140,12 @@ function buildObject (schema, code, name) {
   var laterCode = ''
 
   Object.keys(schema.properties).forEach((key, i, a) => {
+    /* code += `
+      if (obj.hasOwnProperty('${key}')) {
+        json += '${$asString(key)}:'`*/
     code += `
-      json += '${$asString(key)}:'
-    `
+      if (obj.${key} !== undefined) {
+        json += '${$asString(key)}:'`
 
     const result = nested(laterCode, name, '.' + key, schema.properties[key])
 
@@ -150,7 +153,20 @@ function buildObject (schema, code, name) {
     laterCode = result.laterCode
 
     if (i < a.length - 1) {
-      code += 'json += \',\''
+      code += `
+        json += \',\'`
+    }
+
+    if (schema.properties[key].required) {
+      code += `
+      } else {
+        throw new Error('${key} is required!')
+      }
+      `
+    } else {
+      code += `
+      }
+      `
     }
   })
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -140,17 +140,33 @@ function buildObject (schema, code, name) {
...
 
   Object.keys(schema.properties).forEach((key, i, a) => {
+    /* code += `
+      if (obj.hasOwnProperty('${key}')) {
+        json += '${$asString(key)}:'`*/
     code += `
-      json += '${$asString(key)}:'
-    `
+      if (obj.${key} !== undefined) {
+        json += '${$asString(key)}:'`
```
mcollina: please close the '`'  in the new line

index.js
```diff
@@ -202,48 +218,36 @@ function buildArray (schema, code, name) {
...
-      `
+        json += ${funcName}(obj${key})`
       break
     case 'array':
       funcName = (name + key).replace(/[-.\[\]]/g, '')
       laterCode = buildArray(schema, laterCode, funcName)
       code += `
-        json += ${funcName}(obj${key})
-      `
+        json += ${funcName}(obj${key})`
```
mcollina: same in all the above :)

mcollina:
I agree on using `!== undefined`, if you are using a schema, you want that type. On the other end, I see why people would want that behavior. Put a comment to this PR in that place, something like "using !== for perf reasons, see https://github.com/mcollina/fast-json-stringify/pull/3 for discussion".

I'm 100% open to have a configuration switch to use `hasOwnProperty` but I don't think it's important at this stage of the project. So, if someone wants to implement it and send a PR, I'm happy to review it and merge.

Anyway, this is a major difference from `JSON.stringify`, so we should write it in the `README.md`.

index.js
delvedor: I did that change to improve generated code readability, if I close the '`' in the newline the generated code will be:
```javascript
if (obj.lastName !== undefined) {
        json += '"lastName":'

        json += $asString(obj.lastName)

        json += ','

      }
```
instead of 
```javascript
if (obj.lastName !== undefined) {
        json += '"lastName":'
        json += $asString(obj.lastName)
        json += ','
      }
```
Is ok to you?

mcollina: every generated code was put it in this way so that he could easily copied and pasted around, without worrying about backticks :). It's optimized to be read in the "template" form, rather than the generated one: I'm selfish and lazy :D.

delvedor: lol, I cannot fight with *"I'm selfish and lazy :D."*
I will update it :)

delvedor:
Totally agree with you, in the next commit I'll add the README and code changes.

Issue #5, opened 2016-09-18
Pattern and Additional properties


delvedor:
Below the proposal to how implement pattern and additional properties.

#### patternProperties:
Must be an object, where the key is a valid regex and the value is an object.
Example:
```javascript
"patternProperties": {
    "^S_": { "type": "string" },
    "^I_": { "type": "integer" }
  }
```
`patternProperties` will work only for the properties that are not explicitly listed in the properties object.

#### additionalProperties:
Must be a boolean or an object.
If the is setted to false and there are properties that are not explicitly listed in the properties object we should throw an error.
If `additionalProperties` is used in conjunction with `patternProperties` the pecking order will be:
`properties -> patternProperties -> additionalProperties`
Example: 
```javascript
"additionalProperties": {
  "type": "string"
}
```

#### Discussion:
The json schema is very bad documented, so before start to implement these functionalities we need to discuss about few things.
- **additionalProperties**
  If is setted to `{ type: 'string' }` and any additional property is not a string we have two possible cases:
  1. throw an error
  2. stringify the property as string

- **patternProperties**
  Basically the same as above, if a property matches to a pattern with `{ type: 'number' }` but the property is not a number we have two cases:
  1. throw an error
  2. stringify the property as number

If I forgot or I didn't see some edge case please write it below.

mcollina:
Use http://npm.im/fast-safe-stringify to stringify the property, we can even be gentle and not have a strong type check for now, we can add that later (maybe).

delvedor:
Ok but we should stringify it as is or as the pattern/additional property says?

Example:
*schema*
```javascript
properties: {...}
additionalProperties: {
  type: 'string'
}
```
*object to stringify*
```javascript
{ addProp: 42 } // addProp is a number and not a string
```
The result will be
```javascript
'{"addProp":42}'   // original type
                   // or 
'{"addProp":"42"}' // additionaProperty wanted type
```

mcollina:
We should try to coerce and serialize. If not possible, throw.
Il giorno lun 19 set 2016 alle 14:02 Tomas Della Vedova <
notifications@github.com> ha scritto:

delvedor:
Edge case:
Schema:
```javascript
{
  title: 'title',
  type: 'object',
  properties: {...}
  patternProperties: {
    foo: {
      type: 'string'
    }
  }
}
```
if the user tries to stringify `objfoo: {a: true}`what we should do?
1) Stringify as `[object Object]`
[...]

PR #7, opened 2016-09-22
Pattern properties support

delvedor:
Added support for `patternProperties`.
See discussion in #5 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,7 @@ Build a `stringify()` function based on
 
 Supported types:
 
+ * `'string'`
  * `'integer'`
  * `'number'`
  * `'array'`
@@ -106,10 +107,10 @@ const stringify = fastJson({
       type: 'string'
     },
     mail: {
-      type: 'string',
-      required: true
+      type: 'string'
     }
-  }
+  },
+  required: ['mail']
 })
 
 const obj = {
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -119,6 +120,74 @@ const obj = {
...
+      type: 'string'
+    }
+  },
+  patternProperties: {
+    'foo': {
+      type: 'string'
+    },
+    'num': {
+      type: 'number'
+    }
```
mcollina: We should support regexps:

```js
    "patternProperties": {
        "^(/[^/]+)+$": {}
    },
```

delvedor: I'll update this with a more complex example.

index.js
```diff
@@ -131,11 +132,66 @@ function $asRegExp (reg) {
...
+  } else if (type === 'boolean') {
+    return Boolean(value)
+  } else if (type === 'object') {
+    return {}
+  } else if (type === 'array') {
+    return []
+  } else {
+    throw new Error('Cannot coerce ' + value + ' to ' + type)
+  }
+}
```
mcollina: why are you doing a single function, when `type` is known at startup time?

test.js
```diff
@@ -316,3 +316,112 @@ test('missing values', (t) => {
...
+    properties: {},
+    patternProperties: {
+      foo: {
+        type: 'array'
+      }
+    }
+  })
+
+  obj = { foo: 'true', ofoo: 0, arrfoo: [1, 2], objfoo: { tyrion: 'lannister' } }
+  t.equal('{"foo":[],"ofoo":[],"arrfoo":[],"objfoo":[]}', stringify(obj))
```
mcollina: You can split each of those tests in its own `test`. They are hard to read.
You can also split each feature in its own file.
See https://github.com/mcollina/pino as an example.

delvedor: Yeah, you are right, I'll update it.

README.md
```diff
@@ -119,6 +120,74 @@ const obj = {
...
+  type: 'object',
+  properties: {
+    field1: {
+      type: 'string'
+    },
+    field2: {
+      type: 'number'
+    }
+  }
+})
```
mcollina: I don't think this is correct, it is missing the  `obj` property.

README.md
```diff
@@ -119,6 +120,74 @@ const obj = {
...
+  nickname: 'nick',
+  matchfoo: 42,
+  otherfoo: 'str'
+  matchnum: 3
+}
+
+console.log(stringify(obj)) // '{"nickname":"nick","matchfoo":"42","otherfoo":"str","matchnum":3}'
+```
+`Object` and `Array` will be stringified as `{}` and `[]`.  
+Since `fast-json-stringify` is faster than `JSON.stringify` because works on a **fixed schema**, we strongly advice you to not pass complex objects or array as object values, but decompose them in their individual properties and stringify them.
```
mcollina: to not pass -> not to pass

mcollina: Also, this sentence is probably not ok. It's perfectly fine to pass complex objects, as long as the schema stays fixed.

delvedor: I'll remove all this part, since in the next pr I'll fix the nested object properties, as they works now in case of the array.

delvedor:
I've updated the docs and fixed the nested array/objects stringify.
In the near future we must add to the docs an explanation of why `fast-json-stringify` needs a fixed schema to stringify properties.

I think could be a good idea throw if object or array has not properties/items fields.

mcollina:
Great job :).

What happens if one of the properties defined in `properties`  is also matched by `patternProperties`?

delvedor:
`patternProperties` does not overwrite `properties`, you can check it [here](https://github.com/mcollina/fast-json-stringify/blob/0fd2d0c42271d58c2462dfb9ac2fa6bd54eeba01/index.js#L138) and [here](https://github.com/mcollina/fast-json-stringify/blob/0fd2d0c42271d58c2462dfb9ac2fa6bd54eeba01/test.js#L341-L360) :)

index.js
```diff
@@ -4,7 +4,7 @@ function build (schema) {
   /*eslint no-new-func: "off"*/
   var code = `
     'use strict'
-
+    const properties = ${JSON.stringify(schema.properties)}
```
mcollina: Nice one, I completely missed this. Can you please explicit why this is needed with a comment?

PR #11, opened 2016-10-04
Additional properties support

delvedor:
Added support for `additionalProperties`.
See discussion in #5.

*Ps: I took the occasion to add a table of contents to the docs.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,20 @@ JSON.stringify obj x 1,763,980 ops/sec ±1.30% (88 runs sampled)
 fast-json-stringify obj x 5,085,148 ops/sec ±1.56% (89 runs sampled)
 ```
 
+#### Table of contents:
+- <a href="#example">`Example`</a>
+- <a href="#api">`API`</a>
+ - <a href="#fastJsonStringify">`fastJsonStringify`</a>
+ - <a href="#specific">`Specific use cases`</a>
+ - <a href="#required">`Required`</a>
+ - <a href="#missingFields">`Missing fields`</a>
+ - <a href="#patternProperties">`Pattern Properties`</a>
+ - <a href="#additionalProperties">`Additional Properties`</a>
+- <a href="#acknowledgements">`Acknowledgements`</a>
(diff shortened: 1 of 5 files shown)


mcollina:
We should add a case for `additionalProperties: true` and  `additionalProperties: false`. If `additionalProperties: true` is passed, we use fast-safe-stringify to JSONify them (to avoid circular references).

delvedor:
In case of `additionalProperties: false` all the values that are not present in `properties`/`patternProperties` are ignored.

Regarding `additionalProperties: true`, I never see in the json docs this case, only `false` or `object`. That's why I wrote [this](https://github.com/mcollina/fast-json-stringify/blob/0d4d02d0c11b9fceefe9931378d84da727ef0b3b/index.js#L254-L256).

mcollina:
Look at this: https://spacetelescope.github.io/understanding-json-schema/reference/object.html.

> The additionalProperties keyword is used to control the handling of extra stuff, that is, properties whose names are not listed in the properties keyword. By default any additional properties are allowed.

I think we should "bend the rule here", and support `additionalProperties`  as being false by default. If someone wants to add random properties, I think we should support `additionalProperties: true`.

delvedor:
For me is ok support `additionalProperties: true`, but how?
You say to use `fast-safe-stringify`, but we can't use require in the generated code.
I saw the source code of `fast-safe-stringify` and it's composed by three different functions and it's exposing only one of them, so I can't `fast-safe-stringify.toString()` as we do with the other functions.

Ideas?

mcollina:
You can pass it in as an argument of https://github.com/mcollina/fast-json-stringify/blob/master/index.js#L55.

delvedor:
If this implementation is ok for you I'll update the docs to describe this behaviour :)

mcollina:
The implementation is ok.

mcollina:
Does this slows down benchmarks? I don't think so, but please check.

Anyway, LGTM.

delvedor:
I'll add more benchmarks in the future.
In the meantime I merge this :P

PR #12, opened 2016-10-07
$ref support

delvedor:
As titled.
This implementation supports both *internal* and *external* definitions.

**Caveats:**
If the user is using an external json file for his definitions it must use `__dirname` to get the correct path.
See the following example (the same of the test).
```javascript
const schema = {
  title: 'object with $ref',
  type: 'object',
  properties: {
    obj: {
      $ref: __dirname + '/ref.json#/definitions/def'
    }
  }
}
```
If everything is ok to you I can update the docs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -141,13 +141,17 @@ function $asRegExp (reg) {
   return '"' + reg + '"'
 }
 
-function addPatternProperties (pp, ap) {
+function addPatternProperties (schema) {
+  var pp = schema.patternProperties
   let code = `
       var keys = Object.keys(obj)
       for (var i = 0; i < keys.length; i++) {
         if (properties[keys[i]]) continue
   `
   Object.keys(pp).forEach((regex, index) => {
+    if (pp[regex]['$ref']) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -241,26 +250,41 @@ function additionalProperty (ap) {
...
   `
 }
 
+function refFinder (ref, schema) {
+  // Split file from walk
+  ref = ref.split('#')
+  // If external file
+  if (ref[0]) {
+    schema = require(ref[0])
+  }
```
mcollina: Never do this, it's sync and done in a place where it might cause to bugs. We should not load files into Node. However, we should use an approach similar to https://www.npmjs.com/package/is-my-json-valid#external-schemas.

index.js
```diff
@@ -241,26 +250,41 @@ function additionalProperty (ap) {
...
+  const walk = ref[1].split('/')
+  let code = 'return schema'
+  for (let i = 1; i < walk.length; i++) {
+    code += `['${walk[i]}']`
+  }
+  return (new Function('schema', code))(schema)
```
mcollina: Nice!!

PR #30, opened 2017-04-27
Support for long integers

### LINKED ISSUE #28: Support for long integers
Support for 64bit integers as numbers instead of strings would be very useful. The use case being 3rd-party APIs that enforce it.

From my understanding, the `$asNumber` function could simply not convert the input to a `Number` and it would pretty much work as it (since the source long would be a string anyway). However, I don't know if that would introduce a backward compatibility issue.

Alternatively, the string could be validated with regex, but that might be too slow.

I thought about adding a `long` type but that doesn't exist in the JSON Schema specification.

What do you think?

rochdev:
closes #28

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ fast-json-stringify obj x 4,976,369 ops/sec ±1.00% (89 runs sampled)
  - <a href="#patternProperties">`Pattern Properties`</a>
  - <a href="#additionalProperties">`Additional Properties`</a>
  - <a href="#ref">`Reuse - $ref`</a>
+ - <a href="#long">`Long integers`</a>
 - <a href="#acknowledgements">`Acknowledgements`</a>
 - <a href="#license">`License`</a>
 
@@ -292,6 +293,30 @@ const externalSchema = {
 }
 
 const stringify = fastJson(schema, { schema: externalSchema })
+```
+
+<a name="long"></a>
+#### Long integers
+Long integers (64-bit) are supported using the [long](https://github.com/dcodeIO/long.js) module.
+Example:
+```javascript
+const Long = require('long')
+
+const stringify = fastJson({
+  title: 'Example Schema',
+  type: 'object',
+  properties: {
+    id: {
+      type: 'integer'
+    }
+  }
+})
+
+const obj = {
+  id: Long.fromString('18446744073709551615', true)
+}
+
+console.log(stringify(obj)) // '{"id":18446744073709551615}'
 ```
 
 <a name="acknowledgements"></a>
--- a/index.js
+++ b/index.js
@@ -2,6 +2,13 @@
 
 const fastSafeStringify = require('fast-safe-stringify')
 
+let isLong
+try {
+  isLong = require('long').isLong
+} catch (e) {
+  isLong = null
+}
+
 function build (schema, options) {
   options = options || {}
   /* eslint no-new-func: "off" */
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -294,6 +295,30 @@ const externalSchema = {
...
+Example:
+```javascript
+const Long = require('long')
+
+const stringify = fastJson({
+  title: 'Example Schema',
+  type: 'object',
+  properties: {
+    id: {
+      type: 'integer'
```
mcollina: I do not think this is part of the json-schema spec, or is it?

index.js
```diff
@@ -62,6 +84,14 @@ function $asNull () {
   return 'null'
 }
 
+function $asInteger (i) {
+  if (isLong && isLong(i)) {
+    return i.toString()
+  } else {
+    return $asNumber(i)
+  }
+}
```
mcollina: those ifs are really not good for performance if long is not in the dependency tree. We should use two predeclared functions, and select one base on long availability.

test/long.test.js
```diff
@@ -0,0 +1,22 @@
...
+    type: 'integer'
+  }
+
+  const validate = validator(schema)
+  const stringify = build(schema)
+  const output = stringify(Long.fromString('18446744073709551615', true))
+
+  t.equal(output, '18446744073709551615')
+  t.ok(validate(JSON.parse(output)), 'valid schema')
+})
```
mcollina: you should tests this both inside object values and arrays.

README.md
rochdev: I cannot find actual documentation on the data types on the official website, but there are examples in the specification that are using the `integer` type. It is also already used in the current [README.md](https://github.com/fastify/fast-json-stringify/blob/master/README.md#api)

index.js
rochdev: This is already the case.

See https://github.com/fastify/fast-json-stringify/pull/30/files#diff-168726dbe96b3ce427e7fedce31bb0bcR32

test/long.test.js
rochdev: Agreed. I will do that tomorrow.

rochdev:
It can be decoded from the JS side with a custom `JSON.parse`. If using `long`, you most probably need a custom `JSON.parse` anyway. Parsing seems to be outside the scope of this library though.

Also, when integrating with 3rd-parties, you don't necessarily have to parse. As an example, I'm building a library for Datadog APM and only have to send JSON to the agent. It requires 64-bit integers for IDs and is built in Go, so it doesn't have any problem parsing longs.

README.md
delvedor: Yes, `integer` is part of the spec, I think @mcollina was talking about `long` numbers.

mcollina:
I'm +1 with the missing tests.

README.md
rochdev: I don't think the specification has an opinion on this. It's basically implementation dependant.

rochdev:
I just thought that `long` could be required as a transitive dependency. Is that ok or would you rather only add support if explicitly declared as a direct dependency?

mcollina:
I'm ok with it being a transitive dependency.
Another option is to enable it explicity when calling `build` inside an option?

rochdev:
There are no options right now if I'm not mistaken though. In any case most libraries probably don't require `long` so it should be fine. From what I can tell it seems libraries using it are mostly binary protocols.

PR #31, opened 2017-04-30
adds benchmarks with es6 syntax

danielsan:
It also adds more benchmarks with strings declared with var, let and const.
Strings "numbered" and not "numbered" (passed through Number call)


mcollina:
CI is failing because of linting: https://travis-ci.org/fastify/fast-json-stringify/jobs/227266790.

Also, why do you think we need this benchmark with ES6 syntax?

delvedor:
I agree with @mcollina, what kind of results you are searching for?

delvedor:
Close due to inactivity.

Issue #24, opened 2017-04-12
Nested additionalProperties throw at runtime


delvedor:
As titled, test to reproduce:
```js
test('nested additionalProperties', (t) => {
  t.plan(1)
  const stringify = build({
    title: 'additionalProperties',
    type: 'object',
    properties: {
      ap: {
        type: 'object',
        additionalProperties: { type: 'string' }
      }
    }
  })

  let obj = { ap: { value: 'string' } }
  t.equal('{"ap":{"value":"string"}}', stringify(obj))
})
```
The output:
```
  nested additionalProperties
  not ok Cannot convert undefined or null to object
    stack: |
      buildObject (index.js:267:10)
      nested (index.js:381:19)
      Object.keys.forEach (index.js:279:20)
      Array.forEach (native)
      buildObject (index.js:267:34)
      build (index.js:28:14)
      Test.test (test/additionalProperties.test.js:186:21)
    at:
      line: 267
      column: 10
      file: index.js
      function: buildObject
    type: TypeError
    test: additionalProperties
    source: |
      Object.keys(schema.properties).forEach((key, i, a) => {
```

delvedor:
I found that the above issue is related to, and is very easy to solve, already done it.
The real issue that was nehind of this, is the following:
```js 
 const stringify = build({
    title: 'additionalProperties',
    type: 'array',
    items: {
      type: 'object',
      properties: {
        ap: {
          type: 'object',
          additionalProperties: { type: 'string' }
        }
      }
    }
  })
```
The above code, with
```js
 [{ ap: { value: 'string' } }]
```
will throw the following error
```
Cannot read property 'value' of undefined
```

I'll continue to investigate and update here when I'll have more results.

delvedor:
Found it, the problem is [here](https://github.com/fastify/fast-json-stringify/blob/master/index.js#L14), if properties is `undefined` the lib will not work. It's an easy fix, so will land the pr.

@mcollina we should think about a refactor of this library, maybe use template strings e find a way to log better error messages :)

SimenB:
@delvedor I don't understand the solution here. You're example in the OP still throws.

I would like to do the exact same thing (serialise all keys of a nested object), but I'm unable to make it work. Do I have to use an array wrapper?

PR #35, opened 2017-05-20
Allow specifying deep additionalProperties=true

SimenB:
Fixes my problem: https://github.com/fastify/fast-json-stringify/issues/24#issuecomment-302866856

### DIFF
--- a/index.js
+++ b/index.js
@@ -74,10 +74,24 @@ function build (schema, options) {
     ;
     return ${main}
   `
-  if (schema.additionalProperties === true) {
+  if (hasAdditionalPropertiesTrue(schema)) {
     return (new Function('fastSafeStringify', code))(fastSafeStringify)
   }
   return (new Function(code))()
+}
+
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -74,12 +74,24 @@ function build (schema, options) {
...
     return ${main}
   `
-  if (schema.additionalProperties === true) {
+  if (hasAdditionalPropertiesTrue(schema)) {
     return (new Function('fastSafeStringify', code))(fastSafeStringify)
   }
   return (new Function(code))()
 }
 
+function hasAdditionalPropertiesTrue (schema) {
```
SimenB: This is just an adapted http://stackoverflow.com/a/15523704/1850276.

SimenB: Can also do this:
```js
function hasAdditionalPropertiesTrue (schema) {
  const stringifiedSchema = fastSafeStringify(schema)

  return stringifiedSchema.includes('"additionalProperties":true')
}
```

That safeguards against circular objects, and thus infinite recursion

PR #22, opened 2017-04-07
Better errors

delvedor:
If we throw an `${type} unsupported` error, is damn hard understand which schema is failing if you have more than one.
This pr fixes that by logging the wrong schema.

Issue #15, opened 2016-10-21
Regression when stringifying arrays


mcollina:
the array case in the benchmarks has become slower than `JSON.stringify`.

PR #36, opened 2017-05-20
Add ability to uglify generated code

SimenB:
@mcollina Something like this you had in mind?

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,14 +7,18 @@ advantages reduces on large payloads.
 Benchmarks:
 
 ```
-JSON.stringify array x 3,679 ops/sec ±1.01% (85 runs sampled)
-fast-json-stringify array x 4,618 ops/sec ±1.64% (87 runs sampled)
-JSON.stringify long string x 13,303 ops/sec ±1.01% (89 runs sampled)
-fast-json-stringify long string x 13,489 ops/sec ±0.88% (90 runs sampled)
-JSON.stringify short string x 4,974,749 ops/sec ±1.14% (86 runs sampled)
-fast-json-stringify short string x 11,030,700 ops/sec ±0.82% (89 runs sampled)
-JSON.stringify obj x 1,774,593 ops/sec ±1.07% (90 runs sampled)
-fast-json-stringify obj x 4,976,369 ops/sec ±1.00% (89 runs sampled)
+JSON.stringify array x 3,288 ops/sec ±5.18% (82 runs sampled)
+fast-json-stringify array x 1,813 ops/sec ±10.21% (71 runs sampled)
+fast-json-stringify-uglified array x 2,106 ops/sec ±3.23% (83 runs sampled)
+JSON.stringify long string x 12,933 ops/sec ±1.27% (87 runs sampled)
+fast-json-stringify long string x 12,221 ops/sec ±3.31% (84 runs sampled)
+fast-json-stringify-uglified long string x 13,256 ops/sec ±0.95% (92 runs sampled)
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -2,6 +2,7 @@
 
 const fastSafeStringify = require('fast-safe-stringify')
 
+var uglify
```
SimenB: Seems inconsistent between `let` and `var` in the code base. Which do you want?

index.js
```diff
@@ -2,6 +2,7 @@
 
 const fastSafeStringify = require('fast-safe-stringify')
 
+var uglify = null
```
SimenB: Seems inconsistent between `let` and `var` in the code base. Which do you want?

delvedor: `var` is better since is a little faster than `let`, thanks! :)

bench.js
```diff
@@ -35,9 +35,9 @@ const obj = {
 
 const multiArray = []
 
-const stringify = require('.')(schema)
-const stringifyArray = require('.')(arraySchema)
-const stringifyString = require('.')({ type: 'string' })
+const stringify = require('.')(schema, { uglify: true })
+const stringifyArray = require('.')(arraySchema, { uglify: true })
+const stringifyString = require('.')({ type: 'string', uglify: true })
```
mcollina: Can you include benchmarks for both uglified and non-uglified versions?

mcollina:
This is also conflicting for the other two PRs, can you rebase as well?

mcollina:
@delvedor I think we can bump v1.0.0 when this lands, it has been pretty stable API-wise for the last releases.

delvedor:
@mcollina I agree, but before I think we should close/improve https://github.com/fastify/fast-json-stringify/pull/22, then we are ready!

SimenB:
Rebased and separate benchmark added plus updated readme.

README.md
```diff
@@ -7,14 +7,18 @@ advantages reduces on large payloads.
...
-JSON.stringify array x 3,679 ops/sec ±1.01% (85 runs sampled)
-fast-json-stringify array x 4,618 ops/sec ±1.64% (87 runs sampled)
-JSON.stringify long string x 13,303 ops/sec ±1.01% (89 runs sampled)
-fast-json-stringify long string x 13,489 ops/sec ±0.88% (90 runs sampled)
-JSON.stringify short string x 4,974,749 ops/sec ±1.14% (86 runs sampled)
-fast-json-stringify short string x 11,030,700 ops/sec ±0.82% (89 runs sampled)
-JSON.stringify obj x 1,774,593 ops/sec ±1.07% (90 runs sampled)
-fast-json-stringify obj x 4,976,369 ops/sec ±1.00% (89 runs sampled)
+JSON.stringify array x 3,288 ops/sec ±5.18% (82 runs sampled)
+fast-json-stringify array x 1,813 ops/sec ±10.21% (71 runs sampled)
```
SimenB: This is worse, but that's #15, right?

delvedor: Yeah, we must definitely fix that issue :S

package.json
```diff
@@ -29,7 +29,8 @@
     "long": "^3.2.0",
     "pre-commit": "^1.1.3",
     "standard": "^10.0.0",
-    "tap": "^10.3.0"
+    "tap": "^10.3.0",
+    "uglify-es": "^3.0.9"
```
SimenB: We need the es variant to support `const` in the code, btw

mcollina:
@SimenB would you like to help maintaining this library and join our team?

SimenB:
@mcollina I'd love to join up! 🙂

mcollina:
I've sent you an invitation!

PR #37, opened 2017-05-29
If a value is undefined, it shouldn't be passed to fastSafeStringify

SimenB:


SimenB:
@mcollina could this get a merge and release? 👼

SimenB:
Thanks!

PR #42, opened 2017-08-15
Note that uglify is unnecessary on node 8.3.0

SimenB:
❤️ 8.3.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,22 +6,39 @@ advantages reduces on large payloads.
 
 Benchmarks:
 
+Node 6.11.2:
+
 ```
-JSON.stringify array x 3,343 ops/sec ±1.28% (86 runs sampled)
-fast-json-stringify array x 3,031 ops/sec ±1.38% (88 runs sampled)
-fast-json-stringify-uglified array x 3,222 ops/sec ±1.12% (87 runs sampled)
-JSON.stringify long string x 12,400 ops/sec ±1.25% (88 runs sampled)
-fast-json-stringify long string x 12,151 ops/sec ±1.16% (84 runs sampled)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -331,6 +348,9 @@ console.log(stringify(obj)) // '{"id":18446744073709551615}'
 If you want to squeeze a little bit more performance out of the serialisation, at the cost of readability in the generated code, you can pass `uglify: true` as an option.
 Note that you have to manually install `uglify-es` in order for it to work. Only version 3 is supported.
 Example:
+
+Note that if you are using Node 8.3.0 or newer, performance gains from using uglify are negligible.
+
```
mcollina: I would note this is not needed at all. Since turbofan these optimizations are kicked in by counting AST nodes, not the function size. See

https://www.nearform.com/blog/node-js-is-getting-a-new-v8-with-turbofan/

SimenB: So instead of "negligible" just say "no performance gains" and link to that blogpost?

mcollina: Yes.

SimenB: Updated

README.md
```diff
@@ -6,6 +6,8 @@ advantages reduces on large payloads.
 
 Benchmarks:
 
+Node 6.11.0:
+
```
mcollina: Can you please rerun those as well? People will make a comparison.

SimenB: Sure! The I'll run it on my beefier work machine (15" 4 core instead of 13" 2 core)

mcollina: As long as they are all run on the same machine.

SimenB: yup, both are now, I updated 8.3 one as well

mcollina:
Good job!

PR #62, opened 2017-12-13
Edit grammar and syntax in readme

bendrucker:
Hey there, thanks for building this! Noticed a few sentences with awkward syntax or grammar in the readme. Went through and touched it up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,6 @@
 # fast-json-stringify&nbsp;&nbsp;[![Build Status](https://travis-ci.org/fastify/fast-json-stringify.svg?branch=master)](https://travis-ci.org/fastify/fast-json-stringify)
 
-__fast-json-stringify__ is two times faster than `JSON.stringify()`.
-It is particularly suited if you are sending small JSON payloads, the
-advantages reduces on large payloads.
+__fast-json-stringify__ is significantly faster than `JSON.stringify()` for small payloads. Its performance advantage shrinks as your payload grows.
 
 Benchmarks:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,8 +1,6 @@
 # fast-json-stringify&nbsp;&nbsp;[![Build Status](https://travis-ci.org/fastify/fast-json-stringify.svg?branch=master)](https://travis-ci.org/fastify/fast-json-stringify)
 
-__fast-json-stringify__ is two times faster than `JSON.stringify()`.
-It is particularly suited if you are sending small JSON payloads, the
-advantages reduces on large payloads.
+__fast-json-stringify__ is significantly faster than `JSON.stringify()` for small payloads. Its performance advantage shrinks as your payload grows.
```
mcollina: Can you explain change a bit more? Have you experience a degradation in throughput?

bendrucker: I was trying to just rephrase the existing idea based on the benchmarks. The most significant change was removing "two times faster." I thought it was clearer to focus on the performance characteristics instead of using a rough average of the performance improvement that might not apply to many users.

bendrucker:
> A clear part that we are missing here is that this should be combined with http://npm.im/flatstr.

What do you mean?

mcollina:
See https://github.com/fastify/fastify/pull/172.

README.md
mcollina: 👍  ok.

bendrucker:
Gotcha, so you're suggesting that the readme should recommend using flatstr in combination with fast-json-stringify?

mcollina:
Yes.

bendrucker:
Added

Issue #64, opened 2017-12-19
Is allOf a possible keyword in schema ?


keldorus:
Hi,

I got a schema like that for a route on fastify.
```json
{
  "200": {
    "type": "object",
    "allOf": [
      {
        "type": "object",
        "required": [
          "name"
        ],
        "properties": {
          "name": {
            "type": "string"
          },
          "tag": {
            "type": "string"
          }
        }
      },
      {
        "required": [
          "id"
        ],
        "type": "object",
        "properties": {
          "id": {
            "type": "integer"
          }
        }
      }
    ]
  },
  "default": {
    "type": "object",
    "required": [
      "code",
      "message"
    ],
    "properties": {
      "code": {
        "type": "integer"
      },
      "message": {
        "type": "string"
      }
    }
  }
}
```

And sadly, my validation function for the 200 (`responseSchemaDef[statusCode]`) look like this : 
```js
$main(obj) {
      var json = '{'
      var addComma = false
  
      json += '}'
      return json
}
```

I'm guessing that `allOf` is not available, but it's a [JSON schema ](http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.7.1) official keywork.

Can you add it?

Same for `oneOf` ?

Regards

mcollina:
Would you mind sending us a PR? We are pretty busy wrapping up 1.0.0 of fastify these days.

keldorus:
I try something

keldorus:
So can I close this one ? :)

mcollina:
let's close this, it was released on 0.16.0.

PR #65, opened 2017-12-19
Add a forEach to test each schema when there is an allOf

keldorus:
Respond to issue #64 

Feel free to comment and argue, I don't open PR often.

### DIFF
--- a/test/allof.test.js
+++ b/test/allof.test.js
@@ -3,11 +3,11 @@
 const test = require('tap').test
 const build = require('..')
 
-test('object with multiple types field', (t) => {
+test('object with allOf and multiple schema on the allOf', (t) => {
   t.plan(2)
 
   const schema = {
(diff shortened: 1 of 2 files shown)


test/allof.test.js
```diff
@@ -0,0 +1,62 @@
...
+const test = require('tap').test
+const build = require('..')
+
+test('object with multiple types field', (t) => {
+  t.plan(2)
+
+  const schema = {
+    title: 'object with multiple types field',
+    type: 'object',
+    allOf: [
```
mcollina: can you add a test for an empty `allOf` and one with one element.

keldorus:
Hum maybe I can add a last test for the required part, no?

mcollina:
Yes

keldorus:
ok done.

PR #69, opened 2018-01-21
Fix null integer

allevo:
According to https://tools.ietf.org/html/rfc7159#section-6, a `NaN` value cannot be represented in JSON format. Infinity too.

This PR should fix https://github.com/fastify/fastify/issues/693

I don't know if this a breaking change.


StarpTech:
LGTM but does the error message contains the property name of the violation or we only got the message `'Cannot coerce to number'` ?

index.js
```diff
@@ -146,8 +146,8 @@ function $asInteger (i) {
 
 function $asNumber (i) {
   var num = Number(i)
-  if (isNaN(num)) {
-    return 'null'
+  if (!isFinite(num)) {
```
StarpTech: `!isFinite(null)` will evaluate to `false` and will return `"0"` IMO only strings can be coerced to a number.

allevo: `isFinite` matches `NaN` and `Infinity`. Both are not used in JSON.

With this implementation, only numeric string and boolean can be coerced to number

StarpTech:
@allevo I already fixed it [here](https://github.com/mcollina/tinysonic/blob/master/tinysonic.js) this is the correct behaviour

```js
function isOrCanBeNumber(str) {
  if (!isNaN(str)) {
    var number = parseFloat(str);
    if (!isNaN(number)) {
      return number;
    }
  }
  throw new Error('Cannot coerce to number')
}

isOrCanBeNumber('22')
isOrCanBeNumber('0b0001')
isOrCanBeNumber(22)
isOrCanBeNumber(22.4)
isOrCanBeNumber(0b0001)

//Throw
isOrCanBeNumber(null)
isOrCanBeNumber(true)
isOrCanBeNumber(false)
```

allevo:
@StarpTech 

The problem is not about which value can be considered number or not. JSON doesn't allow Infinity value. This is a serialization problem.

```js
> isOrCanBeNumber(Infinity)
Infinity
```
and in JSON context (only during the serialization) it is wrong.

And this is your choice:
```js
isOrCanBeNumber(true)
isOrCanBeNumber(false)
```
In many languages, the coercion here is valid (C, PHP, ...)

For the boolean, we can consider them unallowed if you prefer.

StarpTech:
@Please check it again I updated the code

index.js
StarpTech: Yes, but `null` should not be considered as number or fallback to 0. Please look at my updated code. It covers all.

allevo:
We can discuss how we should treat the boolean coercion.

The code you linked above is slower than the one implemented here
```js
'use strict'

var Benchmark = require('benchmark')

var suite = new Benchmark.Suite()

function isOrCanBeNumber (str) {
  if (!isNaN(str) && isFinite(str)) {
    var number = parseFloat(str)
    if (!isNaN(number)) {
      return number
    }
  }
  throw new Error('Cannot coerce to number')
}

function $asNumber (i) {
  var num = Number(i)
  if (!isFinite(num)) {
    throw new Error('Cannot coerce to number')
  } else {
    return '' + num
  }
}

suite
.add('isOrCanBeNumber', function () {
  isOrCanBeNumber('44')
})
.add('$asNumber', function () {
  $asNumber('44')
})
.on('cycle', function (event) {
  console.log(String(event.target))
})
.on('complete', function () {
  console.log('Fastest is ' + this.filter('fastest').map('name'))
})
.run({ 'async': true })
```

```
$ node -v
v8.9.0
```

**Output:**
```
$ node benchmark.js 
isOrCanBeNumber x 36,834,073 ops/sec ±4.14% (75 runs sampled)
$asNumber x 59,180,212 ops/sec ±1.32% (79 runs sampled)
Fastest is $asNumber
```

Anyway I don't like to discuss which implementation is the best: I need to know how to treat the boolean during the number coercion.

StarpTech:
These values should throw `Infinity`, `NaN`, `null`, `false`, `true`

allevo:
> These values should throw Infinity, NaN, null, false, true

Oki. I'll update the code asap

allevo:
@StarpTech code updated. Test added for each values

README.md
```diff
@@ -391,6 +391,20 @@ const stringify = fastJson({
...
 ```
 
+
+<a name="type-coerce"></a>
+## Type coerce
+
+When the object you're serializing does not strictly match the given schema,
+`fast-json-stringify` applies a type coercion as describe below:
+
+- if the schema describes an integer and a Infinity, NaN, null or a boolean was given,
```
StarpTech: I would say "number" because we cover integer and floats

package.json
```diff
@@ -35,6 +35,7 @@
   },
   "dependencies": {
     "ajv": "^6.0.0",
-    "fast-safe-stringify": "^1.2.1"
+    "fast-safe-stringify": "^1.2.1",
+    "microtime": "^2.1.7"
```
delvedor: Is this needed?

allevo: No at all. I'll remove it

allevo:
I'm convinced that `$asBoolean` suffers the same problem. Should I update it too?

https://github.com/fastify/fast-json-stringify/blob/60fc96faae51bea13570b946d1a43f7d8ef670b3/index.js#L155-L159

README.md
allevo: done

package.json
allevo: done

README.md
```diff
@@ -391,6 +391,20 @@ const stringify = fastJson({
...
 ```
 
+
+<a name="type-coerce"></a>
+## Type coerce
+
+When the object you're serializing does not strictly match the given schema,
+`fast-json-stringify` applies a type coercion as describe below:
+
+- if the schema describes a number and a Infinity, NaN, null or a boolean was given,
```
StarpTech: "if the schema describes a number but a .."

delvedor:
> I'm convinced that $asBoolean suffers the same problem. Should I update it too?

Write a test, and if it breaks update it as well :)

README.md
allevo: Done

allevo:
This PR has a drawback: the performance goes down because `$asNumber` and `$asBoolean` throw error. This makes v8 slower

benchmark.js 
```js
'use strict'

var Benchmark = require('benchmark')

var suite = new Benchmark.Suite()

function $asNumber (i) {
  if (i === 0) return '0'
  if (!i || i === true) {
    throw new Error('Cannot coerce to number')
  }
  if (i.constructor === Number) {
    if (!isFinite(i)) {
      throw new Error('Cannot coerce to number')
    } else {
      return '' + i
    }
  }
  if (typeof i === 'object') {
    throw new Error('Cannot coerce to number')
  }
  var num = Number(i)
  if (!isFinite(num)) {
    throw new Error('Cannot coerce to number')
  } else {
    return '' + num
  }
}

function $asNumberOld (i) {
  var num = Number(i)
  if (isNaN(num)) {
    return 'null'
  } else {
    return '' + num
  }
}

function $asBoolean (bool) {
  if (bool === true) return 'true'
  if (bool === false) return 'false'
  if (bool === 'true') return 'true'
  if (bool === 'false') return 'false'
  throw new Error('Cannot coerce to boolean')
}

function $asBooleanOld (bool) {
  return bool && 'true' || 'false' // eslint-disable-line
}

suite
.add('$asNumber', function () {
  $asNumber('44')
})
.add('$asNumberOld', function () {
  $asNumberOld('44')
})
.add('$asBoolean', function () {
  $asBoolean('true')
})
.add('$asBooleanOld', function () {
  $asBooleanOld('true')
})
.on('cycle', function (event) {
  console.log(String(event.target))
})
.on('complete', function () {
  console.log('Fastest is ' + this.filter('fastest').map('name'))
})
.run({ 'async': true })
```

```
$ node benchmark.js 
$asNumber x 59,177,745 ops/sec ±1.61% (78 runs sampled)
$asNumberOld x 82,933,541 ops/sec ±2.86% (73 runs sampled)
$asBoolean x 301,351,695 ops/sec ±2.55% (72 runs sampled)
$asBooleanOld x 308,681,657 ops/sec ±2.38% (78 runs sampled)
```

If I comment the `throw new Error('Cannot coerce to number')` line, the benchmarks are pretty the same

allevo:
Substituting all throws declaration with `return null`, the benchmarks are:

```
$ node benchmark.js 
$asNumber x 145,607,906 ops/sec ±2.03% (78 runs sampled)
$asNumberOld x 143,132,312 ops/sec ±2.10% (75 runs sampled)
$asBoolean x 541,179,174 ops/sec ±1.74% (78 runs sampled)
$asBooleanOld x 529,253,781 ops/sec ±1.61% (77 runs sampled)
```

```
$ node -v
v8.9.0
```

StarpTech:
👍

index.js
```diff
@@ -145,16 +145,34 @@ function $asInteger (i) {
...
     return '' + num
   }
 }
 
 function $asBoolean (bool) {
-  return bool && 'true' || 'false' // eslint-disable-line
+  if (bool === true) return 'true'
+  if (bool === false) return 'false'
+  if (bool === 'true') return 'true'
+  if (bool === 'false') return 'false'
```
mcollina: I'm __very__ -1 on rendering a string of `'true'` as boolean `true`.

index.js
```diff
@@ -145,16 +145,34 @@ function $asInteger (i) {
 }
 
 function $asNumber (i) {
+  if (i === 0) return '0'
+  if (!i || i === true) {
+    throw new Error('Cannot coerce to number')
+  }
+  if (i.constructor === Number) {
```
mcollina: can you do a `typeof` and save the value somewhere?

index.js
```diff
@@ -145,16 +145,34 @@ function $asInteger (i) {
 }
 
 function $asNumber (i) {
+  if (i === 0) return '0'
+  if (!i || i === true) {
+    throw new Error('Cannot coerce to number')
```
mcollina: can you move this `throw` in its own function? Maybe we can avoid the slowdown.

index.js
```diff
@@ -145,16 +145,34 @@ function $asInteger (i) {
 }
 
 function $asNumber (i) {
+  if (i === 0) return '0'
+  if (!i || i === true) {
```
mcollina: this check would be very slow to perform on turbofan.

allevo:
The performance drops down. This is not acceptable.  I'm working on a huge refactor in order to avoid the throw declarations inside the code.

A possible solution could be return `undefined` value (which is not serializable in JSON) and on the `$main` function check the `json` value for throwing the error.

But I need to know if you are ok with this. So please let me know what you think.

Comment if other options are available.

index.js
```diff
@@ -145,16 +145,34 @@ function $asInteger (i) {
 }
 
 function $asNumber (i) {
```
mcollina: generically, I _prefer_ if we firstly check if `i` is valid and return that, and then handle everything else.

index.js
allevo: Truly I replicate the *same* approach used for number: 
- "55" => 55
- "true" => true

Let me know if I'll remove those `if`s

index.js
allevo:
```js
const serializer = build({
  type: 'integer'
})
serializer(new Number(55)) // 55
```

This is the only case in which an instance of Object *is* a number.

The most correct form is the `instanceof` check but it could be expensive

index.js
allevo: Good point

index.js
allevo: I'll do my best

index.js
allevo: The benchmarks don't change. Probably because v8 applies the inline function here.

```
$ node benchmark.js 
$asNumber x 59,491,275 ops/sec ±2.15% (76 runs sampled)
$asNumberOld x 88,423,066 ops/sec ±1.88% (77 runs sampled)
$asBoolean x 310,544,231 ops/sec ±1.98% (77 runs sampled)
$asBooleanOld x 311,224,120 ops/sec ±1.67% (76 runs sampled)
```

allevo: Remove this `if`.

index.js
allevo: Remove this `if`

index.js
allevo: done

allevo:
New benchmark:

```
$ node benchmark.js 
$asNumber x 65,311,200 ops/sec ±2.13% (77 runs sampled)
$asNumberOld x 91,749,540 ops/sec ±2.82% (74 runs sampled)
$asBoolean x 341,353,240 ops/sec ±1.79% (75 runs sampled)
$asBooleanOld x 346,044,324 ops/sec ±1.90% (77 runs sampled)
```

index.js
mcollina: I'm not 100% sure we are doing the right thing.

@delvedor would you prefer this to be strict about the type?

index.js
```diff
@@ -145,16 +145,22 @@ function $asInteger (i) {
...
-  var num = Number(i)
-  if (isNaN(num)) {
-    return 'null'
-  } else {
-    return '' + num
+  if (i === 0) return '0'
+  if (i !== '' && i !== null && i !== undefined && i !== true && i !== false && (i.constructor === Number || typeof i !== 'object')) {
```
mcollina: Can you get away with:

```
const type = typeof i
if (type === 'number' && isFinite(i)) {
  return '' + i
} else if (typeof === 'string') {
  const num = Number(i)
  if (isFinite(num)) {
    return '' + num
  }
  throw new Error()
}
```

allevo:
> Another alternative option is to skip fields that are not matching their type, and use null.

This breaks the interface.

I can rewrite that `if` using other formulas but the problem persiste: the presence of `throw` inside the function slows the performance down.

Probably the only solution is to rewrite the code in order to return `undefined` or `null` and in the `$main` function throw the error.

index.js
mcollina: @allevo have you seen this?

allevo: That implementation allow the empty string `''` to be coerced to `0`.

The benchmark:
```
$asNumber x 62,418,169 ops/sec ±2.23% (75 runs sampled)
$asNumberOld x 92,619,961 ops/sec ±2.14% (78 runs sampled)
```

As you can see that implementation doesn't impact the speed.

As already said, the problem here is the `throw` statement.

allevo:
@mcollina last commit introduces a different approach: an `error` variable is used to store the error if something goes wrong. This allows us to avoid the `throw`s increasing the performance.

No error will be thrown: the serializer function now returns a string or an instance of Error.

The bench script has this result:


**master**
```
$ node bench.js
JSON.stringify array x 2,234 ops/sec ±4.35% (78 runs sampled)
fast-json-stringify array x 4,251 ops/sec ±3.53% (81 runs sampled)
fast-json-stringify-uglified array x 4,243 ops/sec ±1.54% (83 runs sampled)
JSON.stringify long string x 8,886 ops/sec ±1.67% (82 runs sampled)
fast-json-stringify long string x 9,068 ops/sec ±1.90% (83 runs sampled)
fast-json-stringify-uglified long string x 9,207 ops/sec ±1.60% (83 runs sampled)
JSON.stringify short string x 3,067,865 ops/sec ±1.55% (80 runs sampled)
fast-json-stringify short string x 13,849,216 ops/sec ±1.68% (82 runs sampled)
fast-json-stringify-uglified short string x 13,584,504 ops/sec ±1.94% (80 runs sampled)
JSON.stringify obj x 1,222,879 ops/sec ±2.16% (79 runs sampled)
fast-json-stringify obj x 3,893,722 ops/sec ±2.55% (81 runs sampled)
fast-json-stringify-uglified obj x 3,932,748 ops/sec ±1.21% (85 runs sampled)
```
**fix-null-integer**
```
$ node bench.js
JSON.stringify array x 2,154 ops/sec ±4.66% (76 runs sampled)
fast-json-stringify array x 2,878 ops/sec ±2.91% (80 runs sampled)
fast-json-stringify-uglified array x 2,780 ops/sec ±2.72% (79 runs sampled)
JSON.stringify long string x 8,930 ops/sec ±1.97% (83 runs sampled)
fast-json-stringify long string x 8,975 ops/sec ±1.81% (78 runs sampled)
fast-json-stringify-uglified long string x 9,212 ops/sec ±1.45% (83 runs sampled)
JSON.stringify short string x 3,067,814 ops/sec ±1.40% (82 runs sampled)
fast-json-stringify short string x 13,911,871 ops/sec ±1.85% (80 runs sampled)
fast-json-stringify-uglified short string x 14,087,323 ops/sec ±1.67% (82 runs sampled)
JSON.stringify obj x 1,207,085 ops/sec ±2.10% (78 runs sampled)
fast-json-stringify obj x 2,976,231 ops/sec ±2.68% (84 runs sampled)
fast-json-stringify-uglified obj x 3,050,007 ops/sec ±1.16% (85 runs sampled)
```

This PR introduces some validations. That's why the performance is lower.

As always *any* suggestions are appreciated.

mcollina:
I'm traveling for all of this week, I'd like to have a closer look at this issue and see what's going on - if anything can be done at all.

Can we hold off until I come back?

allevo:
@mcollina sure

I continue updating this PR if I'll find a better implementation

mcollina:
Closing for #70

PR #70, opened 2018-01-31
skip non-coerceable numbers

mcollina:
This is an alternative to #69. Instead of throwing, it skips.

**master**:

```
JSON.stringify array x 3,379 ops/sec ±2.12% (86 runs sampled)
fast-json-stringify array x 6,406 ops/sec ±2.28% (84 runs sampled)
fast-json-stringify-uglified array x 6,283 ops/sec ±2.68% (83 runs sampled)
JSON.stringify long string x 9,841 ops/sec ±2.10% (86 runs sampled)
fast-json-stringify long string x 9,857 ops/sec ±2.14% (85 runs sampled)
fast-json-stringify-uglified long string x 9,862 ops/sec ±1.99% (85 runs sampled)
JSON.stringify short string x 4,071,077 ops/sec ±1.99% (83 runs sampled)
fast-json-stringify short string x 22,269,979 ops/sec ±2.17% (80 runs sampled)
fast-json-stringify-uglified short string x 23,330,578 ops/sec ±1.87% (85 runs sampled)
JSON.stringify obj x 1,670,370 ops/sec ±1.83% (86 runs sampled)
fast-json-stringify obj x 5,627,527 ops/sec ±2.15% (85 runs sampled)
fast-json-stringify-uglified obj x 5,726,660 ops/sec ±2.48% (87 runs sampled)
```

**no-numbers**:

```
JSON.stringify array x 3,460 ops/sec ±1.77% (89 runs sampled)
fast-json-stringify array x 6,622 ops/sec ±2.37% (87 runs sampled)
fast-json-stringify-uglified array x 6,590 ops/sec ±1.64% (89 runs sampled)
JSON.stringify long string x 9,932 ops/sec ±2.15% (84 runs sampled)
fast-json-stringify long string x 9,835 ops/sec ±2.46% (83 runs sampled)
fast-json-stringify-uglified long string x 9,859 ops/sec ±2.40% (85 runs sampled)
JSON.stringify short string x 4,236,321 ops/sec ±2.00% (88 runs sampled)
fast-json-stringify short string x 23,252,000 ops/sec ±2.10% (85 runs sampled)
fast-json-stringify-uglified short string x 22,688,699 ops/sec ±2.34% (83 runs sampled)
JSON.stringify obj x 1,620,479 ops/sec ±2.34% (83 runs sampled)
fast-json-stringify obj x 5,773,492 ops/sec ±1.87% (88 runs sampled)
fast-json-stringify-uglified obj x 5,749,619 ops/sec ±1.88% (86 runs sampled)
```

In order to maintain the throughput, I had to inline `$asNumber` and `$asInteger` in a couple of places, in fact making the generation **faster** when rendering arrays.

Please review carefully.

Related to fastify/fastify#693.

### DIFF
--- a/index.js
+++ b/index.js
@@ -314,12 +314,18 @@ function additionalProperty (schema, externalSchema, fullSchema) {
   } else if (type === 'integer') {
     code += `
         ${addComma}
-        json += $asString(keys[i]) + ':' + $asInteger(obj[keys[i]])
+        var t = Number(obj[keys[i]])
+        if (isLong && isLong(obj[keys[i]]) || !isNaN(t)) {
+          json += $asString(keys[i]) + ':' + $asInteger(obj[keys[i]])
+        }
     `
   } else if (type === 'number') {
     code += `
-        ${addComma}
-        json += $asString(keys[i]) + ':' + $asNumber(obj[keys[i]])
+        var t = Number(obj[keys[i]])
+        if (!isNaN(t)) {
+          ${addComma}
+          json += $asString(keys[i]) + ':' + t
+        }
     `
   } else if (type === 'boolean') {
     code += `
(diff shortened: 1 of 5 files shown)


mcollina:
cc @allevo

StarpTech:
Sry for the late response. Is it correct that no error is thrown when a value could not be used as a number?

mcollina:
It throws only if the field is marked as required. Otherwise it is skipped.

StarpTech:
ok, but in JSON schema a valid value is required as soon a value is set.

I know this change was made due to performance reason but the correct schema validation should be prioritized.

mcollina:
This module is not about validating outgoing data. This module is about spitting out JSON in a fast way according to a given schema. If you care about the input being valid to protect against programmer errros, I think we should provide a dev-only flag to do so using ajv. 

This PR fixes the output, as `"null"`  for numbers was a bug.

StarpTech:
That's a good point I didn't consider it at the review.

mcollina:
@StarpTech that's a valid use case to discuss, maybe we should provide an option when building (then we can expose that option in fastify). Would you like to send a PR?

PR #71, opened 2018-02-01
Removed fast-safe-stringify, use plain JSON.stringify.

mcollina:
As titled. We removed it from fastify long ago anyway.


SimenB:
That means there's no way to handle circular objects :( Thoughts on that? I can just use an old version, that makes me a bit sad, though

mcollina:
@SimenB you should specify a schema, or just use `fast-safe-stringify`  yourself and then send the JSON down.

SimenB:
Right now we have an `extras` field on the object (we use this for a logger) which can be whatever, everything else logged is a schema.

Passing just that part through `fast-safe-stringify` manually should work though. Would I have to send the result from that through `JSON.parse` before passing it to this module to avoid double-escaping?

mcollina:
You can possibly just add a `toJSON` function on that object.

PR #98, opened 2018-05-07
Nested if then else

allevo:
Implement nested if then else as the test explains

https://github.com/fastify/fast-json-stringify/blob/0f3e9726c141de6dea10feba9717b51967973bf9/test/if-then-else.test.js#L31-L67

NB: the developer shouldn't mention the property twice

### DIFF
--- a/index.js
+++ b/index.js
@@ -518,6 +518,7 @@ function addIfThenElse (schema, name, externalSchema, fullSchema) {
   var code = ''
   var r
   var laterCode = ''
+  var innerR
 
   const i = schema.if
   const then = schema.then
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -512,57 +514,82 @@ function buildInnerObject (schema, name, externalSchema, fullSchema) {
...
+  var r
+  var laterCode = ''
+  var innerR
+
+  const i = schema.if
+  const then = schema.then
+  const e = schema.else
+  delete schema.if
+  delete schema.then
+  delete schema.else
```
mcollina: we might want to clone this before deleting.

allevo: why?

allevo: Good point I clone the schema deeply

allevo: done

PR #102, opened 2018-06-01
Fixes 101

delvedor:
As titled.

With this change we have a significant drop in the performances when serializing short strings, on my machine from ~8 million ops/sec to ~7 millions.
I'll investigate if we can improve this.

cc @mixecan

### DIFF
--- a/index.js
+++ b/index.js
@@ -207,6 +207,7 @@ function $asString (str) {
 function $asStringSmall (str) {
   var result = ''
   var last = 0
+  var found = false
   var l = str.length
   var point = 255
   for (var i = 0; i < l && point >= 32; i++) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -217,7 +217,7 @@ function $asStringSmall (str) {
     }
   }
   if (last === 0) {
-    result = str
+    result += str
```
mcollina: can you try assigning last to `-1`  at the top of the function, and checking for `-1` in the if?

delvedor: It will not work, because we are using it inside the the loop: `result += str.slice(last, i) + '\\'`.
Slicing with a negative index as first value breaks some tests.

mcollina:
I've pushed an alternate implementation. Please verify.

delvedor:
It works perfectly, and we are going faster than before :)

mixecan:
Great!
When do you think this will be available in Fastify?

delvedor:
Now ;)

PR #105, opened 2018-07-22
Allow empty schemas as properties.

ruiaraujo:
It allows covering opaque types.

I am using this together with a GraphQL endpoint which has custom scalars. These scalars are opaque from the type system so a JSON schema cannot be derived.

### DIFF
--- a/index.js
+++ b/index.js
@@ -726,7 +726,11 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
         code += `
           else json+= null
         `
-      } else throw new Error(`${schema} unsupported`)
+      } else {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -726,7 +726,11 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
         code += `
           else json+= null
         `
-      } else throw new Error(`${schema} unsupported`)
+      } else {
```
delvedor: Probably just to be safe we can add a check here to verify that you are passing an empty schema.

delvedor:
Landed in `v1.7.0`, thanks!

Issue #118, opened 2018-09-25
Support for toJSON


sarneeh:
I'd like to have a little different model structure internally in my app than returning it externally via API. Unfortunately, using `fastify` and `fast-json-stringify` this can't be done as `fast-json-stringify` does not use the objects `toJSON` function to map the data somehow.

I guess it could be on purpose to improve the performance, but this is a little bit unintuitive to do: 

`reply.send(model.toJSON());`

and what even worse:

```js
const models = [arrayOfModels];
reply.send(models.map(model => model.toJSON()));
```

This doesn't make much sense as there's no difference if I have to do this manually or if it would be done by `fast-json-stringify` - just as I'd expect.

Here's a little reproduction:

```js
const fastJson = require("fast-json-stringify");

const stringify = fastJson({
    type: "object",
    properties: {
        groups: {
            type: "array",
            items: {
                type: "string"
            }
        }
    }
});

class User {
    constructor() {
        this.groups = [{ name: "admin" }, { name: "subscriber" }];
    }

    toJSON() {
        return {
            ...this,
            groups: this.groups.map(group => group.name)
        };
    }
}

const user = new User();

console.log(stringify(user)); // returns: {"groups":["[object Object]","[object Object]"]}
console.log(JSON.stringify(user)); // returns: {"groups":["admin","subscriber"]}
```

Could you somehow relate to this issue? And maybe you have some idea for a solution in `fastify`?

mcollina:
If it is not too costly perf-wise, we should check if an object has a toJSON method, and call that before applying the schema. Would you like to send a PR that does that?

cemremengu:
@sarneeh Yes we even pushed the same second. Sorry it's my bad I was just lurking around and saw this, I should have asked first.  It is your right to have this PR 😄

sarneeh:
@cemremengu No problem 😄 I needed some time to find out what's going on in this project. I hope I didn't mess this up completely 😄

cemremengu:
@sarneeh I will close mine in favor of yours and we can continue work from there

sarneeh:
@mcollina Awesome, thank you for the fast publish 🎉

mcollina:
Glad to help! Keep sending PRs :).

PR #119, opened 2018-09-25
Apply toJSON if exists

cemremengu:
Attempts #118 

Node v8.11.3
```
FJS creation x 5,338 ops/sec ±4.96% (83 runs sampled)
JSON.stringify array x 3,254 ops/sec ±1.73% (94 runs sampled)
fast-json-stringify array x 4,697 ops/sec ±1.30% (95 runs sampled)
fast-json-stringify-uglified array x 4,446 ops/sec ±1.37% (95 runs sampled)
JSON.stringify long string x 10,435 ops/sec ±0.39% (97 runs sampled)
fast-json-stringify long string x 10,417 ops/sec ±0.59% (91 runs sampled)
fast-json-stringify-uglified long string x 10,401 ops/sec ±0.50% (95 runs sampled)
JSON.stringify short string x 4,313,552 ops/sec ±4.12% (86 runs sampled)
fast-json-stringify short string x 9,798,598 ops/sec ±0.94% (94 runs sampled)
fast-json-stringify-uglified short string x 9,851,120 ops/sec ±0.79% (93 runs sampled)
JSON.stringify obj x 1,804,050 ops/sec ±1.18% (96 runs sampled)
fast-json-stringify obj x 4,279,013 ops/sec ±1.63% (92 runs sampled)
fast-json-stringify-uglified obj x 4,198,298 ops/sec ±0.86% (96 runs sampled)
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,7 +14,7 @@ var ajv = new Ajv({
   // them.
   logger: {
     log: console.log,
-    warn: function () {},
+    warn: function () { },
     error: console.error
   }
 })
@@ -557,6 +557,7 @@ function addIfThenElse (schema, name, externalSchema, fullSchema) {
 function buildObject (schema, code, name, externalSchema, fullSchema) {
   code += `
     function ${name} (obj) {
+      if (obj.toJSON) { return obj.toJSON() }
       var json = '{'
       var addComma = false
   `
--- a/test/tojson.test.js
+++ b/test/tojson.test.js
@@ -1,0 +1,37 @@
+'use strict'
+
+const test = require('tap').test
+const build = require('..')
+
+function User () {
+  this.groups = [{ name: 'admin' }, { name: 'subscriber' }]
+  this.toJSON = function () {
+    var result = []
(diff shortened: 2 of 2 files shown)


sarneeh:
This approach is wrong. You should not call `JSON.stringify()` on your `toJSON` method. Quoting MDN: `If the value has a toJSON() method, it's responsible to define what data will be serialized.`. `toJSON` is for mapping your object to some other structure (for external use), not for serialization.

cemremengu:
@sarneeh Yes, you are right I was actually just trying to mimic the output of `fast-stringify` instead of a correct implementation

lmammino:
I have seen the `toJSON` being used also to evaluate lazy loaded properties (most often in the context of ORM systemsl, to make sure that the serialized items would contain all the attributes. Even in this case the `toJSON` would return a plain JS object with the fields that should be serialized and not a serialized string.

cemremengu:
@lmammino Yeah, looks like I messed up. Didnt know that the spec/usage is that strict on how you use it.

Thanks for the info I really appreciate it! :smile:

PR #122, opened 2018-10-09
basic support to test against official JSON-Schema-Test-Suite

StarpTech:
This PR will add basic support to test against the official https://github.com/json-schema-org/JSON-Schema-Test-Suite. It contains only tests for the `required` feature.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: false
 node_js:
 - '6'
 - '8'
-- '9'
 - '10'
 
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: node_js
 sudo: false
 node_js:
-- '4'
 - '6'
 - '8'
 - '9'
--- a/index.js
+++ b/index.js
@@ -118,7 +118,7 @@ function build (schema, options) {
   return (Function.apply(null, dependenciesName).apply(null, dependencies))
 }
 
-var objectKeywords = [
+const objectKeywords = [
   'maxProperties',
   'minProperties',
   'required',
@@ -128,7 +128,7 @@ var objectKeywords = [
   'dependencies'
 ]
 
-var arrayKeywords = [
+const arrayKeywords = [
   'items',
   'additionalItems',
   'maxItems',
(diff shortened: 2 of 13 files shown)


StarpTech:
It should be added incrementally or what do you mean?

mcollina:
Would it be hard to just add it to this PR?

StarpTech:
The `required` test-case or all JSON files?

StarpTech:
I added draft 6,7 of the `required` test-case.

StarpTech:
@mcollina one important note: I had to add explicitly the `type: 'object'` property to all schemas otherwise `fast-json-stringify` will throw `undefined unsupported` 😥 acccording to the spec it should fallback to type object when `properties` is used.

StarpTech:
Do we really need to support Node 4 ?

mcollina:
> @mcollina one important note: I had to add explicitly the type: 'object' property to all schemas otherwise fast-json-stringify will throw undefined unsupported 😥 acccording to the spec it should fallback to type object when properties is used.

We should actually behave properly and assume the same, it shouldn't be hard to patch.

> Do we really need to support Node 4 ?

We can bump the major without much problems if it's just for the support.

StarpTech:
> We should actually behave properly and assume the same, it shouldn't be hard to patch.

I think it is very hard because we have to handle the `any` type-case which will make optimizations very complicated. In my opinion it's fine to set some prerequisites in order to build an optimized stringify function this includes to always add the type in your schemas.

e.g 

schema
```json
{
    "properties": {
        "foo": {},
        "bar": {}
    },
    "required": ["foo"]
}
```
this is valid:
```json
{"foo": 1}
```

https://github.com/json-schema/json-schema/issues/172

StarpTech:
> We can bump the major without much problems if it's just for the support.

Great, I will remove node 4 from the matrix.

StarpTech:
Please check again I infer type based on the validation keywords.

.travis.yml
```diff
@@ -1,7 +1,6 @@
 language: node_js
 sudo: false
 node_js:
-- '4'
 - '6'
 - '8'
 - '9'
```
mcollina: can you drop 9 as well?

index.js
```diff
@@ -114,6 +118,59 @@ function build (schema, options) {
...
+  'pattern'
+]
+
+var numberKeywords = [
+  'multipleOf',
+  'maximum',
+  'exclusiveMaximum',
+  'minimum',
+  'exclusiveMinimum'
+]
```
mcollina: I would just use `const` for these.

PR #127, opened 2018-11-14
Fix: move `long` and `uglify-es` to `dependencies`

stefanmirck:
To avoid errors after `[npm|yarn] install --production`, both `long` and `uglify-es` should be listed under `dependencies` instead of `devDependencies`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,15 +29,15 @@
   "devDependencies": {
     "benchmark": "^2.1.4",
     "is-my-json-valid": "^2.19.0",
-    "long": "^4.0.0",
     "pre-commit": "^1.2.2",
     "proxyquire": "^2.1.0",
     "standard": "^12.0.1",
-    "tap": "^12.0.0",
-    "uglify-es": "^3.3.9"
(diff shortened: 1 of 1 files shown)


stefanmirck:
I tried to deploy a lambda to `now` that uses this package and it can’t resolve `long` and `uglify-es` during build time. Below is a part of the log file:

```
2018-11-14T19:39:29.258Z  'uglify-es' is imported by ../../tmp/5f659d4a/user/node_modules/fast-json-stringify/index.js, but could not be resolved – treating it as an external dependency
2018-11-14T19:39:29.260Z  'long' is imported by ../../tmp/5f659d4a/user/node_modules/fast-json-stringify/index.js, but could not be resolved – treating it as an external dependency
```

Indeed, both packages are `require`d in the actual source of this package, which led me to believe the packages should be listed under `dependencies` rather than `devDependencies`.

mcollina:
You should probably open an issue on now, as those dependencies are optional, and we deal with them not being there programmatically.

stefanmirck:
Will do!

PR #128, opened 2018-11-15
Fix serialization of anyOf with both array and object types

Marsup:
`anyOf` used with both a type `object` and `array` fails to serialize either of the type since the serialization functions are named the same. I added a prefix to the array one but I'm not sure it's the right way to go, if you have a better suggestion I can change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -796,7 +796,7 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
       `
       break
     case 'array':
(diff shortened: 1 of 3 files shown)


Marsup:
Last commit fixes the glob failing to match test dir root files, not sure what happened there.

mcollina:
Awesome!

mcollina:
Can you open an issue of what is currently not supported? If it’s valid json schema, why not supporting it if it does not have a overhead?

Issue #130, opened 2018-12-04
Better support for `null` values


marcbachmann:
I ran into an issue with the serialization of values I query from a postgres database.
The behavior in `fast-json-stringify` is currently not documented and there's no mention anywhere about how non-matching input values behave, so I guess this topic didn't come up yet (at least not on github).

We should be able to provide a good solution for the problem, be it dropping the properties, throw an error or coerce the value to something else and then also document that.

With `ajv` and json-schema, such multi-type values are supported with `type: ["string", "null"]`, optionally in combination with an `anyOf` declaration.
Maybe we could support something similar in this library.
```bash
> node -p "(new (require('ajv'))).validate({type: ['string', 'null']}, 'string')"
true

> node -p "(new (require('ajv'))).validate({type: ['string', 'null']}, null)"
true
```

This topic already came up for strings in https://github.com/fastify/fast-json-stringify/pull/41, but there's no mention of other types. It got fixed by coercing the value to an empty string whereas for the other types the behavior is still not defined.
```bash
> node -p "require('fast-json-stringify')({type: 'string'})(null)"
""
```

Integers:
```bash
> node -p "require('fast-json-stringify')({type: 'integer'})(null)"
0
```

Objects:
```bash
> node -p "require('fast-json-stringify')({type: 'object'})(null)"
undefined:62
      var obj = typeof input.toJSON === 'function'
                             ^
TypeError: Cannot read property 'toJSON' of null
    at $main (eval at build (/repo/node_modules/fast-json-stringify/index.js:118:20), <anonymous>:62:30)
    at [eval]:1:49
    at ContextifyScript.Script.runInThisContext (vm.js:50:33)
    at Object.runInThisContext (vm.js:139:38)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:653:30)
    at evalScript (bootstrap_node.js:479:27)
    at startup (bootstrap_node.js:180:9)
    at bootstrap_node.js:625:3
```

mcollina:
I agree that we should fix this. I’m a bit snowed under atm, would you mind to send a PR with a fix? Both coercion or dropping properties are ok for me.

marcbachmann:
Oh, I just had a look at the benchmark and saw that `"type": ["string", "null"]` and `"type": ["object", "null"]` are already supported. So this issue would be about defining the behavior a strict `"type: ["object"]` should behave.
[...]

PR #131, opened 2018-12-21
Better support for `null` values (#130)

jtassin:
Support of the `nullable` option of ajv (Open API 3 specification) on object.

An object declared nullable can have a null value.

Today, a null object stringified lead to this error: 

```javascript
TypeError: Cannot read property 'toJSON' of null
```
As described [in the issue #130](https://github.com/fastify/fast-json-stringify/issues/130)

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ fast-json-stringify-uglified obj x 5,331,581 ops/sec ±0.73% (91 runs sampled)
  - <a href="#ref">`Reuse - $ref`</a>
  - <a href="#long">`Long integers`</a>
  - <a href="#uglify">`Uglify`</a>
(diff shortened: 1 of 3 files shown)


mcollina:
@jtassin I do not think this is fixing https://github.com/fastify/fast-json-stringify/issues/130. Can you also add a unit test/fix for that case?

mcollina:
Can you also add some docs in the readme about nullable?

jtassin:
Thx @mcollina for your feedback.

I'll add tests on #130 specific case and some docs.

jtassin:
Here comes a few docs and a test case on #130 issue.

mcollina:
Thanks you very much!

PR #134, opened 2019-01-15
Corrects integer handling

### LINKED ISSUE #133: multiple types with null fails
Related to #77
Tested on master this morning.

Please find below the test code:
```js
'use strict';

const fastJsonStringify = require('fast-json-stringify');

const f = fastJsonStringify({
  type: 'object',
  properties: {
    data: {
      type: ['integer', 'null']
    }
  }
});

console.log(f({data:4}))
```

The expected output is `{"data":4}` but I've got `{"data":null}`.

bhamon:
Closes #133

### DIFF
--- a/index.js
+++ b/index.js
@@ -854,6 +854,11 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
               ${index === 0 ? 'if' : 'else if'}(Array.isArray(obj${accessor}))
               ${nestedResult.code}
             `
+          } else if (type === 'integer') {
+            code += `
+              ${index === 0 ? 'if' : 'else if'}(Number.isInteger(obj${accessor}))
+              ${nestedResult.code}
+            `
           } else {
             code += `
               ${index === 0 ? 'if' : 'else if'}(typeof obj${accessor} === "${type}")
--- a/test/typesArray.test.js
+++ b/test/typesArray.test.js
@@ -3,7 +3,7 @@
 const test = require('tap').test
 const build = require('..')
 
-test('possibly nullable primitive alternative', t => {
+test('possibly nullable primitive alternative', (t) => {
   t.plan(1)
 
   const schema = {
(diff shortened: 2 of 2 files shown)


bhamon:
The hidden question was: is fast-json-stringify expected to be JSON-schema compliant ?
If yes, this call should throws.
If no, I can ignore the last check I introduced in the test suite and go on with the PR.

mcollina:
I think we should throw if it’s a required field, or omit it if it’s not.

@delvedor?

bhamon:
By "required" are you talking about the parent object?
Because if I have the following schema it is still invalid to coerce the `data=null` value to `0`:
```
{
  type: 'object',
  properties: {
    data: {
      type: 'integer'
    }
  }
}
```

mcollina:
If `data = null`  is passed, nothing should be produced, and it should throw if that property is required.

mcollina:
What's the status of this @bhamon?

bhamon:
I'm currently overcharged on other subjects. I'll get back to this as soon as I have some free time.

cemremengu:
Ping @bhamon

cemremengu:
@mcollina @delvedor is it possible to merge this and I will add the required test case in separate PR ?

delvedor:
@cemremengu are you able to push directly to this pr?
Otherwise, you can open a new one and maintain the same commits that are in this one.

cemremengu:
May I patch release this?

bhamon:
The added test (https://github.com/fastify/fast-json-stringify/pull/134/commits/3380dc16f9eb9803cc0326865a59c999aa039790#diff-bfdcad0d83d6ed08c8e0c1ad2d02ddf8R31) doesn't check that `{"data": null}` works, 50% coverage.
Also, the PR doesn't solve the illegitimate `{"data": null}` serialization problem.

cemremengu:
@bhamon would you like send another PR or at least provide failing cases so someone else can patch it?

cemremengu:
Ah I guess you mean this comment ?https://github.com/fastify/fast-json-stringify/pull/134#issuecomment-454698980

PR #135, opened 2019-01-17
Issue 134 - Clarify docs on additionalProperties 

romansavchenko:


romansavchenko:
That's fair. Maybe there could be a doc that says that other things throw then. It's the one validation that won't throw so I like the extra clarity.

README.md
```diff
@@ -227,7 +227,8 @@ console.log(stringify(obj)) // '{"matchfoo":"42","otherfoo":"str","matchnum":3,"
 *additionalProperties* must be an object or a boolean, declared in this way: `{ type: 'type' }`.  
 *additionalProperties* will work only for the properties that are not explicitly listed in the *properties* and *patternProperties* objects.
 
-If *additionalProperties* is not present or is set to `false`, every property that is not explicitly listed in the *properties* and *patternProperties* objects,will be ignored, as described in <a href="#missingFields">Missing fields</a>.  
+If *additionalProperties* is not present or is set to `false`, every property that is not explicitly listed in the *properties* and *patternProperties* objects,will be ignored, as described in <a href="#missingFields">Missing fields</a>.
+Note that this is unique compared to other schema rules in that it will not throw an error, it will simply ignore missing fields.
```
mcollina: I would remove the term unique, as this might change/it is hard to validate. I would write something like “Missing fields are ignored to avoid having to rewrite objects before serializing. However, other schema rules would throw in similar situations.”

mcollina:
Good work!

PR #140, opened 2019-02-28
feat: add schema name to error

### LINKED ISSUE #139: Be more verbose on schema validation.
In Fastify when some schema is not valid it's almost impossible to guess which one. So in order to make it clear error should have schema `key` somewhere.

My suggestion is to wrap https://github.com/fastify/fast-json-stringify/blob/master/index.js#L937 in `try..catch`:

```js
Object.keys(externalSchema).forEach((key) => {
  try {
    ajv.addSchema(externalSchema[key], key);
  } catch (err) {
    err.message = `${key}: ${err.message}`;
    // or create custom error
    throw err;
  }
});
```

SkeLLLa:
Closes #139 

Resulting error will look like:
```
Error: "mySchema" schema is invalid: ...
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -934,7 +934,12 @@ function loadUglify () {
 function isValidSchema (schema, externalSchema) {
   if (externalSchema) {
     Object.keys(externalSchema).forEach(key => {
-      ajv.addSchema(externalSchema[key], key)
+      try {
+        ajv.addSchema(externalSchema[key], key)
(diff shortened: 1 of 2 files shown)


mcollina:
Can you please add a unit test?

SkeLLLa:
@mcollina  Should they just check presence of schema name?

mcollina:
yes!

PR #144, opened 2019-03-03
Add more tests

cemremengu:
Add more tests for #134

### DIFF
--- a/test/typesArray.test.js
+++ b/test/typesArray.test.js
@@ -28,6 +28,31 @@ test('possibly nullable primitive alternative', (t) => {
   }
 })
 
+test('possibly nullable primitive alternative with null value', (t) => {
+  t.plan(1)
+
+  const schema = {
+    title: 'simple object with multi-type nullable primitive',
+    type: 'object',
+    properties: {
+      data: {
+        type: ['integer']
+      }
+    }
+  }
+
+  const stringify = build(schema)
+
+  try {
+    const value = stringify({
+      data: null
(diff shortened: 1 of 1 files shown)


cemremengu:
@mcollina The schema below still converts `null` to `0` should we change that?

```js
  const schema = {
    title: 'simple object with multi-type nullable primitive',
    type: 'object',
    properties: {
      data: {
        type: 'integer' // works if ['integer']
      }
    }
  }
```

mcollina:
How does it behave with an explicit undefined? What would you propose?

cemremengu:
For `undefined` it will just return `{}`. 

I propose to not convert `null` to `0` since they are not the same thing especially in database terms. So `'integer'` and `['integer']` should produce the same output. However, then, having `['integer', null]` will only for verbosity instead of providing new functionality.

I am a bit confused actually 😄 In essence I don't want it to have runtime errors for not specifying `null` explicitly but also want to have the field as `null` there instead of `0` or completely omitting it.

mcollina:
Go with your best shot, we’ll change it if we need to.

Il giorno dom 3 mar 2019 alle 21:54 Cemre Mengu <notifications@github.com>
ha scritto:

StarpTech:
@cemremengu don't forget the [JSON-Schema testuite](https://github.com/fastify/fast-json-stringify/tree/master/test/json-schema-test-suite) if you want to improve the tests. It would be great to have a structure.

PR #145, opened 2019-03-04
Make null handling for integer and number consistent

cemremengu:
Since `['integer']` and `['number']` types are nullable, we should make 'integer' and 'number' types nullable by default as well.

Should we throw if a `null` value is given to a `required` property?

### DIFF
--- a/index.js
+++ b/index.js
@@ -195,6 +195,8 @@ function $asNull () {
 }
 
 function $asInteger (i) {
+  if (i === null) return $asNull()
+
   if (isLong && isLong(i)) {
     return i.toString()
   } else {
@@ -203,6 +205,8 @@ function $asInteger (i) {
 }
 
 function $asNumber (i) {
+  if (i === null) return $asNull()
+
   var num = Number(i)
   if (isNaN(num)) {
     return 'null'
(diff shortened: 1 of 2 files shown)


mcollina:
I don't think this is correct. Maybe we did a mistake in the other PR. Does this is validate with that JSON schema?

cemremengu:
> Does this is validate with that JSON schema?
Can you elaborate a little?

The author of the previous PR later mentioned that his work was incomplete :/

mcollina:
Can you please verify that:

1. both the PRs original JSON schema validates the output parsed from fast-json-stringify?
2. `'integer'` does not seems nullable at all by default.

cemremengu:
If I understand correctly (apologies for being slow), assigning `null` to both:

```js
{
    type: 'object',
    properties: {
      int: {
        type: ['integer']
      }
    }
}
```
or 

```js
{
    type: 'object',
    properties: {
      int: {
        type: 'integer'
      }
    }
}
```
are NOT valid according to JSON schema. However, while it should normally throw (again according to JSON schema), it coerces `null` values to `0` which is still not correct. So we should either deviate from JSON schema or make it throw

mcollina:
Let’s make it throw then.

cemremengu:
Note that to be consistent it should throw for

- strings (currently null becomes empty string)
- boolean (currently null becomes false)

as well

mcollina:
@cemremengu then we should coerce to `0`, with consistency with strings as boolean.

mcollina:
(we should probably make them throw, but that'd be a semver-major change).

cemremengu:
While it is somewhat alright to coerce strings and boolean, for numbers coercing `null` to `0` is not a good idea in my opinion in addition to not being JSON schema compliant.

I would like to work on a PR where they all throw but I guess we can only land that in Fastify v3

mcollina:
I think we should solve the issue that this PR was addressing, making things consistent.
Coercing `null` to `0` is the same of coercing `null` to `''`. Let's make v1 consistent and then fix this in the next major.

PR #146, opened 2019-03-04
Make integer/number handling consistent for type arrays

cemremengu:
Make integer handling consistent for type arrays as discussed in #145


mcollina:
Can you add a quick note about null and coercion in the README?

PR #161, opened 2019-04-15
Support for external $ref plain name fragment

yndc:
Allows referencing external schemas using plain name fragment identifier


Eomm:
@yndc could you link me some docs/standard details?

I ask this because for what I know this pattern `#someId` is a reference to the json schema itself, a sort of localhost. So I don't think this logic is compliant to the standard.
[Here](https://tools.ietf.org/html/rfc6901#section-6) the standard.

yndc:
@Eomm Based on [this](https://json-schema.org/latest/json-schema-core.html#rfc.section.8.2.3), I think that there's 2 way the pound sign `#` could be interpreted when dereferencing schemas in $ref.

The first one is like the standard you mentioned (`#` acts like a 'base' of the current/root JSON document), and the second one is location independent identifiers where the rest of the $id after # is used as an arbitrary unique identifier of a schema regardless of its location.

I think to separate between the two, implementations need to check wether the character after `#` in $id is `/` (example: `#/abc` which will use JSON pointer to dereference $ref), or letters (according to [the spec](https://json-schema.org/latest/json-schema-core.html#rfc.section.8.2.3)).

I'll update my PR shortly

yndc:
Alright, updated my PR. The last commits from the maintainer really helps in resolving arbitrary-located schemas in `isValidSchema`. Now plain name fragments could be truly location independent in the external schema.

yndc:
@Eomm Added the test, it's very crucial to ensure that $refs should resolve to the main schema first!

I also added a check to not searching the $id in all of external schemas when the `$ref` explicitly state which external schema to use. 

So if a `$ref` is referencing to `first#target` and `first` doesn't have the schema `#target`, it should error even if `#target` is defined elsewhere since it's explicitly state to look into `first`.

mcollina:
@Eomm is this a semver-patch?

Eomm:
> @Eomm is this a semver-patch?

Yes it is because it fixes the implementation of the [Dereferencing](https://tools.ietf.org/html/draft-handrews-json-schema-01#section-8.3.2) paragraph in the standard

Landing later today 👍

PR #166, opened 2019-05-08
Dropped uglifier support

### LINKED ISSUE #97: Remove uglifier support
This is only relevant for Node 6 now, and I think we should drop this feature in the next major release.

L2jLiga:
As far as I understand the issue #97, uglify-es was added to improve throughput in Node.js 6 which is [EOL](https://github.com/nodejs/Release#end-of-life-releases) nowadays and I guess it can be safety dropped now.
One more point to drop support is that uglify-es is [no longer maintained](https://github.com/mishoo/UglifyJS2/issues/3156#issuecomment-392943058)

Closes #97

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,19 +13,15 @@ FJS creation x 8,951 ops/sec ±0.51% (92 runs sampled)
 
 JSON.stringify array x 5,146 ops/sec ±0.32% (97 runs sampled)
 fast-json-stringify array x 8,402 ops/sec ±0.62% (95 runs sampled)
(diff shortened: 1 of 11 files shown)


mcollina:
Good work! We'll land this when we cut a new major dropping Node 6 support.

mcollina:
I think we should land this, drop Node 6, and cut a new major.

L2jLiga:
I can rebase it on top of master

mcollina:
I've rebased this to a next branch, so we can land when we cut v2.

PR #168, opened 2019-05-28
Sanitize input so it's safer

mcollina:
This PR is the coordinated work on myself and @phra to make fast-json-stringify more robust in case of untrusted user input. This does not make it safe to use (bugs and vulns can always be there), but it covers all the cases we could think of.

### DIFF
--- a/index.js
+++ b/index.js
@@ -310,7 +310,7 @@ function addPatternProperties (schema, externalSchema, fullSchema) {
     }
     var type = pp[regex].type
     code += `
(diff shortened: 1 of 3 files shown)


mcollina:
> I think we should also add a note in the README explaining when and where this utility should be used.
Since the main goal is performance, if someone is building the serializer function every time it will lose all the benefits given by the library and thus there is no reason to use it.

This is already covered in the README, would you mind proposing some text change in a follow-up PR? I'll get this released now.

PR #171, opened 2019-05-28
[Benchmark]: Add more similar library to benchmark

dalisoft:
_Note: The `json-strify` does not validate, just stringifies, `json-strify` is made for users who don't want write schemas and direct replacement to `JSON.stringify` for little extra performance`_

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,5 +56,9 @@
     "ignore": [
       "tap"
     ]
+  },
+  "optionalDependencies": {
+    "compile-json-stringify": "^0.1.2",
+    "json-strify": "0.0.2"
   }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -56,5 +56,9 @@
     "ignore": [
       "tap"
     ]
+  },
+  "optionalDependencies": {
+    "compile-json-stringify": "^0.1.2",
+    "json-strify": "^0.1.7"
```
mcollina: These should be `devDependencies`, not `optionalDependencies`.

dalisoft: Now looks OK?

PR #172, opened 2019-05-28
[Feature PR]: Add `noThrow` option

### LINKED ISSUE #169: [Feature Request]: Allow return "errors" instead of throwing
Hello and thanks for this library.
For my case, this library improves Request/second by 3-7% which is good for me.
But when schema fails, i want get error by string and show to users.

Using try/catch decreases performance by ~40% for my app. Is there any way get it or available options?

PS: I'm not using `fastify`, i made own library :)

dalisoft:
This PR closes #169 

Related issue: #169

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,7 +28,9 @@ function isValidSchema (schema, name) {
       name = ''
     }
     const first = validate.errors[0]
-    const err = new Error(`${name}schema is invalid: data${first.dataPath} ${first.message}`)
+    const err = new Error(
+      `${name}schema is invalid: data${first.dataPath} ${first.message}`
+    )
     err.errors = isValidSchema.errors
     throw err
   }
@@ -87,7 +89,7 @@ function build (schema, options) {
   switch (schema.type) {
     case 'object':
       main = '$main'
-      code = buildObject(schema, code, main, options.schema, schema)
+      code = buildObject(schema, code, main, options.schema, schema, options)
       break
     case 'string':
       main = schema.nullable ? $asStringNullable.name : $asString.name
@@ -106,7 +108,7 @@ function build (schema, options) {
       break
     case 'array':
       main = '$main'
-      code = buildArray(schema, code, main, options.schema, schema)
+      code = buildArray(schema, code, main, options.schema, schema, options)
       break
     default:
       throw new Error(`${schema.type} unsupported`)
@@ -129,7 +131,7 @@ function build (schema, options) {
   }
 
   dependenciesName.push(code)
-  return (Function.apply(null, dependenciesName).apply(null, dependencies))
+  return Function.apply(null, dependenciesName).apply(null, dependencies)
 }
 
 const objectKeywords = [
@@ -151,11 +153,7 @@ const arrayKeywords = [
   'contains'
 ]
 
-const stringKeywords = [
-  'maxLength',
-  'minLength',
-  'pattern'
-]
+const stringKeywords = ['maxLength', 'minLength', 'pattern']
 
 const numberKeywords = [
   'multipleOf',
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -28,7 +28,9 @@ function isValidSchema (schema, name) {
       name = ''
     }
     const first = validate.errors[0]
-    const err = new Error(`${name}schema is invalid: data${first.dataPath} ${first.message}`)
+    const err = new Error(
+      `${name}schema is invalid: data${first.dataPath} ${first.message}`
+    )
```
mcollina: Can you avoid changing the format of the code?

dalisoft: I will try, but this formatted with `standard` code-style. I manually didn't touched this part

dalisoft: I'm formatted with `npx standard --fix`

dalisoft:
For objects and string should work, for string and numbers not required like objects, because strings or numbers can be manually filtered.
Using `try/catch` reduces performance by ~50% (after optimising other parts of code) which is not good.

For `fastify` same performance degaration should be apply as `fastify` gets error? (I not sure about this, but i think yes)

mcollina:
I don't understand the point you are making.

I agree that supporting this pattern is a good thing, but we should returning the data through the same value.  As an example check: https://www.npmjs.com/package/is-my-json-valid#verbose-mode-shows-more-information-about-the-source-of-the-error.

dalisoft:
Yep, thats what i want. Without errors (`throw`), only returning `.errors`

jsumners:
The only viable way I see is as outlined in https://medium.com/knock-engineering/happy-path-error-handling-1f698068a345

dalisoft:
@jsumners Thanks for help, but i'm trying improve my library RPS by count (which is matters), i like `fast-json-stringify` and works very very best, but when data mismatches, performance is reducing by `fast-json-stringify` `throw new Error(...)` which i'm don't want

jsumners:
My suggestion has nothing to do with throwing an error. I am suggesting a complete change in the returned data type:

```js
const result = fastStringify({foo: 'bar'})
if (result.error) {
  console.error(result.error.message)
  process.exit(1)
}
console.log(result.value) // `{"foo":"bar"}`
```

dalisoft:
Wow, sorry, looks I didn’t understand good. Now understand. +1

mcollina:
@dalisoft would you mind to update this PR to match https://github.com/fastify/fast-json-stringify/pull/172#issuecomment-498245616?

dalisoft:
@mcollina I don't think i can, because i don't know how to push `errors` or `return with like this`. Maybe i other OPEN when i can do this? (but not in few weeks)

mcollina:
I'm sorry for the long delay, I didn't see this. Are you still interesting in pushing this forward?

index.js
```diff
@@ -351,9 +338,16 @@ function addPatternProperties (schema, externalSchema, fullSchema) {
           json += $asString(keys[i]) + ':' + $asBoolean(obj[keys[i]])
       `
     } else {
-      code += `
+      if (options && options.noThrow) {
+        code += `
+        ${addComma}
+        return {error: "Cannot coerce " + obj[keys[i]] + " to ${type}"}
```
jsumners, line 344: Why isn't the value of the `error` property an instance of `Error`?

dalisoft: Sometimes i wanted to get only object instead of Error improve performance before

dalisoft:
As i deprecated (maybe currently) project, I don’t think this feature need. 
I used default API and worked fine last time i tried except invalid response mismatching performance drop

index.js
jsumners: I would be okay with this change if the error property is an instance of the Error object.

PR #176, opened 2019-07-29
Check for null value before using `in`.

wprl:
Would it be possible to add a check for null values in this function?  The `in` operator is causing an error when `null` is found in the schema.  For example:

```javascript
 response: {
    200: {
      description: 'Success',
      type: 'string',
      nullable: true
      enum: [null] 
    }
  }
```

With JSON Schema, one could use `type: 'null'`, but I have a requirement to only use OAS 3.0 features, so I can't change the schema to use `type: 'null'`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -188,6 +188,7 @@ function inferTypeByKeyword (schema) {
 }
 
 function hasAnyOf (schema) {
+  if (!schema) { return false }
   if ('anyOf' in schema) { return true }
 
   var objectKeys = Object.keys(schema)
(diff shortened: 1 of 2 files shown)


Eomm:
Could you add more details to your use case and error?

This snippet is not throwing errors:

```
const stringify = fastJson({
  description: 'Success',
  type: 'string',
  nullable: true,
  enum: [null]
})

const out = stringify('hello')
console.log(out)
```

Then, looking docs, the [type string and enum with null could be have conflicts](https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values)

mcollina:
Would you mind adding a unit test?

wprl:
Added the test

wprl:
Thanks!

PR #190, opened 2019-10-24
Support of multiple level refs in anyOf

### LINKED ISSUE #189: anyOf + ref doesn't compile
## 🐛 Bug Report

anyOf + ref doesn't compile. Ajv compiles the schema fine though

## To Reproduce

```js
require('fast-json-stringify')({
  type: 'object',
  properties: {
    'cs': {
      'anyOf': [
        {
          '$ref': '#/definitions/Option'
        },
        {
          'type': 'boolean'
        }
      ]
    }
  }, 'definitions': {
    'Option': {
      'type': 'string'
    }
  }
});

yields 

/fastify-bug-report/node_modules/fast-json-stringify/index.js:945
        throw new Error(`${schema.type} unsupported`)
```

## Expected behavior

Compile the schema - please note that anyOf without refs works fine.

## Your Environment

- *node version*: 10
- *fast-json-stringify version*: >=1.15.6
- *os*: Mac
- I tried also older fast-json-stringy versions (13) and got same problem

hazam:
Closes #189 

#### Checklist

- [X] documentation is changed or added
- [X] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/index.js
+++ b/index.js
@@ -960,8 +960,16 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
     case undefined:
       if ('anyOf' in schema) {
         schema.anyOf.forEach((s, index) => {
+          // inner levels of nesting
           while (s.$ref) {
             s = refFinder(s.$ref, fullSchema, externalSchema)
+          }
+          if ('anyOf' in s) {
+            s.anyOf.forEach((innerS) => {
+              if (innerS.$ref) {
+                throw new Error('Multiple nesting levels of anyOf not supported.')
+              }
+            })
           }
           var nestedResult = nested(laterCode, name, key, s, externalSchema, fullSchema, subKey !== '' ? subKey : 'i' + index)
           code += `
--- a/index.js
+++ b/index.js
@@ -214,6 +214,8 @@ function $asNull () {
 function $asInteger (i) {
   if (isLong && isLong(i)) {
     return i.toString()
+  } else if (typeof i === 'bigint') {
+    return i.toString()
   } else {
     return $asNumber(i)
   }
--- a/index.js
+++ b/index.js
@@ -479,6 +479,8 @@ function idFinder (schema, searchedId) {
 }
 
 function refFinder (ref, schema, externalSchema) {
+  if (externalSchema && externalSchema[ref]) return externalSchema[ref]
+
   // Split file from walk
   ref = ref.split('#')
 
--- a/index.js
+++ b/index.js
@@ -685,15 +685,13 @@ function buildCodeWithAllOfs (schema, code, laterCode, name, externalSchema, ful
 }
 
 function buildInnerObject (schema, name, externalSchema, fullSchema) {
-  var laterCode = ''
-  var code = ''
+  var result = buildCodeWithAllOfs(schema, '', '', name, externalSchema, fullSchema)
   if (schema.patternProperties) {
-    code += addPatternProperties(schema, externalSchema, fullSchema)
+    result.code += addPatternProperties(schema, externalSchema, fullSchema)
   } else if (schema.additionalProperties && !schema.patternProperties) {
-    code += addAdditionalProperties(schema, externalSchema, fullSchema)
+    result.code += addAdditionalProperties(schema, externalSchema, fullSchema)
   }
-
-  return buildCodeWithAllOfs(schema, code, laterCode, name, externalSchema, fullSchema)
+  return result
 }
 
 function addIfThenElse (schema, name, externalSchema, fullSchema) {
--- a/index.js
+++ b/index.js
@@ -911,7 +911,7 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
   subKey = subKey || ''
 
   if (schema.$ref) {
-    schema = refFinder(schema.$ref, fullSchema, fullSchema)
+    schema = refFinder(schema.$ref, fullSchema, externalSchema)
   }
 
   if (schema.type === undefined) {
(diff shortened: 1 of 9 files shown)


test/anyof.test.js
```diff
@@ -248,3 +248,34 @@ test('null value in schema', (t) => {
...
+        type: 'string'
+      }
+    }
+  }
+
+  try {
+    build(schema)
+  } catch (e) {
+    t.fail()
+  }
```
mcollina: Would you mind to also test if this serializes ok?

hazam: I can only fix it for one level of nestedness: if one the anyOfs is a ref, it expands that in the generated code. But if one of them is another anyOf with a ref, then it blows up.

I couldn't figure out how to recursively expand schema

mcollina: I think one is better than nothing, and it might cater to the most common case. Would you mind throwing a proper error in that case something like "nested $ref are not supported yet".

jsumners:
I think the reference resolution needs someone with expertise to really dig in and overhaul it. My team has also hit this issue yesterday via response schemas in Fastify. It's the fourth such issue uncovered since I started experimenting with JSON Schema instead of Joi.

hazam:
@jsumners feel free to take a stab.

the PR handles this:
```js
{
    type: 'object',
    properties: {
      cs: {
        anyOf: [
          {
            $ref: '#/definitions/Option'
          },
          {
            type: 'boolean'
          }
        ]
      }
    },
    definitions: {
      Option: {
        anyOf: [
          {
            type: 'number'
          },
          {
            type: 'boolean'
          }
        ]
      }
    }
  }
```

but not this

```js
 {
    type: 'object',
    properties: {
      cs: {
        anyOf: [
          {
            $ref: '#/definitions/Option'
          },
          {
            type: 'boolean'
          }
        ]
      }
    },
    definitions: {
      Option: {
        anyOf: [
          {
            $ref: '#/definitions/Option2'
          },
          {
            type: 'boolean'
          }
        ]
      },
      Option2: {
        type: 'number'
      }
    }
  }
```

jsumners:
I wish I had time.

index.js
```diff
@@ -927,6 +931,17 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
...
       if ('anyOf' in schema) {
         schema.anyOf.forEach((s, index) => {
+          // inner levels of nesting
+          while (s.$ref) {
+            s = refFinder(s.$ref, fullSchema, externalSchema)
+          }
+          if ('anyOf' in s) {
+            s.anyOf.forEach((innerS) => {
+              if (innerS.$ref) {
+                throw new Error('Multiple nesting levels of anyOf not supported.')
```
mcollina:
```suggestion
                throw new Error('Multiple nesting levels of anyOf and $ref are not supported.')
```

jsumners: I don't think this is an acceptable solution. The documentation states that the module uses JSON Schema -- https://github.com/fastify/fast-json-stringify/tree/a6652fd93a70656786f894f5c9b0d3a5d6eb4ce4#fastjsonstringifyschema . The JSON Schema spec does not limit the nesting level of references.

mcollina: I agree. But I prefer landing a 1-level solution that provides a meaningful error. Should we add "Please send a PR?"

Maybe we could add a comment before it?

jsumners: Okay, how about the error and an open issue asking for help?

hazam: I'm sure you already discussed, but how about finding a way to pre-feed the whole schema to ajv - which then gets used only to process ifElse or anyOf - instead of extracting part of the schema and generating code with these ad-hoc logic?

ajv would have been happy if he had the whole schema from the start

mcollina: I don't think this is about ajv, but rather our internal code generation for the fast serialization.

hazam: Ok today I read the fast-json-stringify internals for the first time so I may be mistaken, but what i detected is that the issue happens at `ajv.validate(schema, obj)` time.

Ajv is passed only a small portion of the schema, and fast-json-stringify takes care of fully 'de-referencing' all the $refs. From this the need to "follow the refs".

mcollina: Exactly! `fast-json-stringify` needs to have the full schema resolved before running. So, this should be fixable.

jsumners: See how this was solved for passing custom AJV instances into Fastify.

hazam: I am just coming from that - custom ajv in fastify. Unfortunately I can't devote more time to the issue, I hope this PR gets accepted to at least provide some basic support.

If not, I'll hack my own npm package or some other work around.

jsumners: That’s a poor attitude to take. Matteo is trying to give feedback and guidance to get the PR accepted and merged.

hazam: Been accepting feedback whole day yesterday - now I don't have any more time. 
I fail to grasp the poor attitude, but ¯\_(ツ)_/¯

hazam: I happen to resurface a bit, should have some time to dedicate to this.

What is actually required for this PR to be accepted? I lost track

mcollina: We need to implement full recursive deferencing of the refs for fast-json-stringify to work properly.

hazam: Actually turns out nested() func was already recursive enough :)
Multiple levels working

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fast-json-stringify",
-  "version": "1.15.6",
+  "version": "1.16.2",
```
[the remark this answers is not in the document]
hazam: I had a conflict since version on master was bumped - checks were not running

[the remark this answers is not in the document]
mcollina: Can you rebase on top of master?

mcollina:
I would recommend to open a fresh PR. There is something really off with this PR that is confusing Github.

package.json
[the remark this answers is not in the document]
hazam: I can't - I literally can't. All rebase bring force me to resolve conflicts of 20+ commits.

Issue #194, opened 2019-11-11
Add support for Moment.JS


ilaif:
## 🚀 Feature Proposal

Given an object to stringify, if it contains properties of type Moment.JS, it should handle it correctly, or support extensibility for the user to do so.

## Motivation

Moment.JS (among other datetime libraries) are very common. As a Fastify user, I'm using Ajv to transform date-time input into Moment.JS, which I move around the code, and when I return a response that contains Moment.JS properties, I would like for the serialization library to format it automatically, just like it handles vanilla Date objects.

## Example

The following example should stringify the `time` property in an ISO format.

```js
const moment = require('moment')

const stringify = fastJson({
  title: 'Time Schema',
  type: 'object',
  properties: {
    time: {
      type: 'string',
      format: 'date-time',
    },
  }
})

const payload = {
  time: moment(),
}

console.log(stringify(payload))
```
* I'm not sure that the schema should say the time is of type 'string' since we pass an object to it (moment.js object).

I understand that inserting non-vanilla logic into the library is probably not the right way, but providing a way to extend the library with external logic (like in Ajv) sounds like a good balance.

jsumners:
1. Doesn't the `moment` object already stringify correctly?
2. Why not? `const payload = { time: moment().toDate() }`

ilaif:
1. The library reads it as not object and outputs "{}".
2. It means I have to maintain this in the code in every reply. I can just do moment().toISOString() if I go that way, but I want it handled automatically.

mcollina:
I think supporting it would be a good feature to add, essentially we would like to call .toIsoString() if date-time format is specified. Would you like to send a PR?

ilaif:
Glad to hear! Will do.

ilaif:
Opened a PR: https://github.com/fastify/fast-json-stringify/pull/195

PR #195, opened 2019-11-11
added support for date-time format

ilaif:
Closes https://github.com/fastify/fast-json-stringify/issues/194

#### Checklist

- [X] documentation is changed or added
- [X] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

#### Additional notes
- As this is the first time I'm touching this beautiful piece of code, please criticize the way I implemented the feature and that I didn't miss any important cases.
- I only implemented the `date-time` format, but `date` and `time` also exist. I think this is a rarer use-case that might require more complex code to achieve safely. 

Thanks!


ilaif:
Done.

Eomm:
@ilaif do you think it is possible to support the format `date` and `time` also?

ilaif:
Yeah, sure, will have a look.

mcollina:
What's the status @ilaif?
@Eomm do you think this is landable?

jsumners:
@delvedor it is part of the spec -- https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.3

Eomm:
@ilaif would you mind to carry this PR to the finish line? :muscle:

package.json
```diff
@@ -35,11 +35,13 @@
     "is-my-json-valid": "^2.20.0",
     "json-strify": "^0.1.7",
     "long": "^4.0.0",
+    "moment": "^2.24.0",
     "pre-commit": "^1.2.2",
     "proxyquire": "^2.1.3",
     "semver": "^7.1.0",
     "standard": "^14.0.0",
     "tap": "^12.6.5",
+    "tap-mocha-reporter": "^3.0.9",
```
mcollina, line 44: Why was this added?

ilaif: My bad, missed it while merging from upstream.

ilaif:
@mcollina @Eomm Hey guys,

I've added support for `date` and `time` formats as defined by `is-my-json-valid` that determines the following formats:
- `date`: `YYYY-MM-DD`
- `time`: `HH:mm:ss` (without UTC offset)

Another important addition is adding moment.js as a runtime dependency of the project so I can use it to check if an object is an instance of moment, and to format to `date` and `time` formats more easily. Let me know if this sounds reasonable to you.

index.js
```diff
@@ -248,6 +282,36 @@ function $asBooleanNullable (bool) {
   return bool === null ? null : $asBoolean(bool)
 }
 
+function $asDatetime (date) {
+  if (date instanceof Date) {
+    return '"' + date.toISOString() + '"'
+  } else if (date instanceof moment) {
```
mcollina, line 288: Unfortunately this check will not work because a user can have more than one version of moment in their dep tree.

index.js
```diff
@@ -248,6 +282,36 @@ function $asBooleanNullable (bool) {
...
+    return '"' + date.toISOString() + '"'
+  } else {
+    return $asString(date)
+  }
+}
+
+function $asDate (date) {
+  if (date instanceof Date) {
+    return '"' + moment(date).format('YYYY-MM-DD') + '"'
+  } else if (date instanceof moment) {
```
mcollina, line 298: Unfortunately this check will not work because a user can have more than one version of moment in their dep tree.

index.js
```diff
@@ -248,6 +282,36 @@ function $asBooleanNullable (bool) {
...
+    return '"' + date.format('YYYY-MM-DD') + '"'
+  } else {
+    return $asString(date)
+  }
+}
+
+function $asTime (date) {
+  if (date instanceof Date) {
+    return '"' + moment(date).format('HH:mm:ss') + '"'
+  } else if (date instanceof moment) {
```
mcollina, line 308: Unfortunately this check will not work because a user can have more than one version of moment in their dep tree.

mcollina:
I'm also -1 in adding `moment` as a runtime dependency.

ilaif:
Fair enough, will write a custom formatting function and use duck typing.

Eomm:
@mcollina I don't have permissions to release this module, could you publish a new version?

mcollina:
Done in 1.19.0

PR #197, opened 2019-12-14
Added BigInt support

delvedor:
As titled! 🚀 

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -214,6 +214,8 @@ function $asNull () {
 function $asInteger (i) {
   if (isLong && isLong(i)) {
     return i.toString()
+  } else if (typeof i === 'bigint') {
+    return i.toString()
   } else {
     return $asNumber(i)
   }
--- a/test/integer.test.js
+++ b/test/integer.test.js
@@ -1,6 +1,8 @@
 'use strict'
 
-const test = require('tap').test
+const t = require('tap')
+const test = t.test
+const semver = require('semver')
 const validator = require('is-my-json-valid')
 const proxyquire = require('proxyquire')
 const build = proxyquire('..', { long: null })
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -214,6 +214,8 @@ function $asNull () {
 function $asInteger (i) {
   if (isLong && isLong(i)) {
     return i.toString()
+  } else if (typeof i === 'bigint') {
```
jsumners: This library currently tests all the way back to Node 6 -- https://github.com/fastify/fast-json-stringify/blob/4b6df2e11c4f5a8b2c408741506f48c3f589865a/azure-pipelines.yml#L5-L13

What will this line do on < 10.13?

delvedor: Nothing, it's just a string comparison, which means that it will always be false.
Anyhow, we will drop Node v6 and v8 once we release Fastufy v3 :)

jsumners: I know. But it still has to work for the versions this will actually target.

test/integer.test.js
```diff
@@ -83,3 +85,10 @@ test('render an object with an additionalProperty of type integer as JSON', (t)
   t.equal(output, '{"num":1615}')
   t.ok(validate(JSON.parse(output)), 'valid schema')
 })
+
+if (semver.gt(process.versions.node, '10.3.0')) {
+  require('./bigint')(t.test, build)
+} else {
+  t.pass('Skip because Node version < 10.4')
```
jsumners: Shouldn’t there be at least one test here to show the behavior on old versions?

index.js
delvedor: It does, as you can see all the tests are passing :)
You can use the BigInt syntax only from Node 10.4 and above.
In any other case the typeof will return something else, so the if will never be true.

test/integer.test.js
delvedor: I fear that I’m missing something. Can you propose a test?
As I said above, this is a good old typeof, which will work in any Node version.
In version ma older than 10.4 you can’t use the BigInt syntax.

PR #202, opened 2020-01-17
 Improve the debugging experience in case of a bad reference

delvedor:
Hello folks!
Typos are a significant pain point in programming, especially with JSON schemas.
This pr introduces a better debugging experience to help the user figure out what's going wrong.

To improve the experience even further, I've used [string-similarity](https://github.com/aceakash/string-similarity), a library that uses the Dice's Coefficient to figure out if two strings are similar. I think this approach can be reused in Fastify as well.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,7 @@ var Ajv = require('ajv')
 var merge = require('deepmerge')
 var util = require('util')
 var validate = require('./schema-validator')
+var stringSimilarity = null
 
 var uglify = null
 var isLong
@@ -490,6 +491,10 @@ function refFinder (ref, schema, externalSchema) {
   if (ref[0]) {
     schema = externalSchema[ref[0]]
 
+    if (schema === undefined) {
+      findBadKey(externalSchema, [ref[0]])
+    }
+
     if (schema.$ref) {
       return refFinder(schema.$ref, schema, externalSchema)
     }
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -517,8 +522,30 @@ function refFinder (ref, schema, externalSchema) {
...
     }
   }
-  const result = (new Function('schema', code))(schema)
+  var result
+  try {
+    result = (new Function('schema', code))(schema)
+  } catch (err) {}
+
+  if (result === undefined) {
+    findBadKey(schema, walk.slice(1))
```
Eomm: Why this sentence is not in the previous `catch`?

delvedor: Because there is a case where the function above does not throw and just returns `undefined`.

Eg:
```js
schema['bad']['good'] // throws
schema['good']['bad'] // undefined and throws later
```

index.js
```diff
@@ -517,8 +522,30 @@ function refFinder (ref, schema, externalSchema) {
...
+  }
   return result.$ref ? refFinder(result.$ref, schema, externalSchema) : result
+
+  function findBadKey (obj, keys) {
+    if (keys.length === 0) return null
+    const key = sanitizeKey(keys.shift())
+    if (obj[key] === undefined) {
+      stringSimilarity = stringSimilarity || require('string-similarity')
+      const { bestMatch } = stringSimilarity.findBestMatch(key, Object.keys(obj))
+      if (bestMatch.rating >= 0.5) {
```
StarpTech: `0.5` is it random? I would just return the best match. The error message describes already both cases (couldn't find, possible match)

delvedor: Yes, it's "random", I think that returning a bad best match is worse than returning nothing :)

PR #216, opened 2020-04-04
add debug mode

### LINKED ISSUE #180: debug mode
ref #175 

In order to let users understand better this module and debug faster their code, I would like to think a `debug mode` that act like this:

- [ ] print out the raw generated code (that help me to understand the problem)
- [ ] when the built `stringify` function is called print (or return?) additional validations information

WDYT?

Eomm:
Closes #180 

I'm not sure about `global.ajv` setup to let it works in some cases (checkout the last test in the new file)

Any ideas about how to improve this "debug mode"? 👍

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,6 +578,31 @@ allowing user input to directly supply a schema.
 It can't be guaranteed that allowing user input for the schema couldn't feasibly expose an attack
 vector.
 
+<a name="debug"></a>
+### Debug Mode
+
+The debug mode can be activated during your development to understand what is going on when things do not
+work as you expect.
+
+```js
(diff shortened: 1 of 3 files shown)


Eomm:
> I think you could add another helper that restores the function by wrapping it in a function that accepts ajv as a parameter and wraps the fjs string in a closure. It's the same approach we use in the original code.

Thks for the suggestion, I have added the utility `.restore()`: I thought that it could be dangerous to add a `typeof schema === 'string'` check in the `build` function, only to avoid add this new `restore` key.

Do you agree?

mcollina:
I agree, this is ok!

PR #219, opened 2020-04-05
serializing null value crash fix

### LINKED ISSUE #218: date field serialization crashes when null
## 🐛 Bug Report

I have a timestamp field in my schema named 'updatedAt' and defined as follows:

```js
{
  updatedAt: { type: ['string', null], format: 'date-time'}
}
```

When a record is created for the first time, the ORM return updatedAt with null, because only createdAt is set on record creation. 

When it goes through serialization, I get the following error:

```
TypeError: Cannot read property 'toISOString' of null
     at $asDatetime (eval at build (/usr/src/app/node_modules/fast-json-stringify/index.js:142:20), <anonymous>:60:26)
     at $arr$mainusermembershipsi (eval at build (/usr/src/app/node_modules/fast-json-stringify/index.js:142:20), <anonymous>:554:17)
     at $arr$mainusermemberships (eval at build (/usr/src/app/node_modules/fast-json-stringify/index.js:142:20), <anonymous>:452:17)
     at $mainuser (eval at build (/usr/src/app/node_modules/fast-json-stringify/index.js:142:20), <anonymous>:403:17)
     at Object.$main (eval at build (/usr/src/app/node_modules/fast-json-stringify/index.js:142:20), <anonymous>:134:17)
     at serialize (/usr/src/app/node_modules/fastify/lib/validation.js:132:41)
     at preserializeHookEnd (/usr/src/app/node_modules/fastify/lib/reply.js:298:15)
    at next (/usr/src/app/node_modules/fastify/lib/hooks.js:101:7)
     at handleResolve (/usr/src/app/node_modules/fastify/lib/hooks.js:112:5)
     at process._tickCallback (internal/process/next_tick.js:68:7)
```

## To Reproduce

The generated eval code strictly checks for underfined, so null is considered a valid value! Here's is what the generated code looks like:
```js
if (obj['updatedAt'] !== undefined) {
  if (addComma) {
    json += ','
  }
  addComma = true
  json += '"updatedAt":'
  json += $asDatetime(obj['updatedAt'])
}
```
And since the function below doesn't check for `null`,  the second condition `date.toISOString ` triggers the error `Cannot read property 'toISOString'` because `date` itself is null!
 ```js
function $asDatetime (date) {
  if (date instanceof Date) {
    return '"' + date.toISOString() + '"'
  } else if (typeof date.toISOString === 'function') {  <----- Error thrown here
    return '"' + date.toISOString() + '"'
  } else {
    return $asString(date)
  }
}
```

My suggestion is to make 2 changes as proposed below:
```js
function $asDatetime (date) {
  if (date instanceof Date) {
    return '"' + date.toISOString() + '"'
  } else if (date && typeof date.toISOString' === 'function') {
    return '"' + date.toISOString() + '"'
  } else {
    return $asStringNullable(date)
  }
}
```

## Your Environment

- *node version*: 10
- *fastify version*: 2.13.0
- *os*: Mac and alpine (dockerized)

mum-never-proud:
resolves #218 

#### Checklist

- [ ] ~~documentation is changed or added~~ (**not needed**)

### DIFF
--- a/index.js
+++ b/index.js
@@ -279,7 +279,7 @@ function $asBooleanNullable (bool) {
 function $asDatetime (date) {
   if (date instanceof Date) {
     return '"' + date.toISOString() + '"'
-  } else if (typeof date.toISOString === 'function') {
+  } else if (date && typeof date.toISOString === 'function') {
     return '"' + date.toISOString() + '"'
   } else {
     return $asString(date)
@@ -1077,9 +1077,10 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
         sortedTypes.forEach((type, index) => {
           var tempSchema = Object.assign({}, schema, { type })
           var nestedResult = nested(laterCode, name, key, tempSchema, externalSchema, fullSchema, subKey)
+
           if (type === 'string') {
             code += `
-              ${index === 0 ? 'if' : 'else if'}(typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || typeof obj${accessor}.toISOString === "function" || obj${accessor} instanceof RegExp)
+              ${index === 0 ? 'if' : 'else if'}(typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || (obj${accessor} && typeof obj${accessor}.toISOString === "function") || obj${accessor} instanceof RegExp)
                 ${nestedResult.code}
             `
           } else if (type === 'null') {
--- a/index.js
+++ b/index.js
@@ -277,9 +277,13 @@ function $asBooleanNullable (bool) {
 }
 
 function $asDatetime (date) {
+  if (!date) {
+    return '""'
+  }
+
   if (date instanceof Date) {
     return '"' + date.toISOString() + '"'
-  } else if (date && typeof date.toISOString === 'function') {
+  } else if (typeof date.toISOString === 'function') {
     return '"' + date.toISOString() + '"'
   } else {
     return $asString(date)
@@ -300,6 +304,10 @@ function $asDate (date) {
 }
 
 function $asTime (date) {
+  if (!date) {
+    return '""'
+  }
+
   if (date instanceof Date) {
     var hour = new Intl.DateTimeFormat('en', { hour: 'numeric', hour12: false }).format(date)
     var minute = new Intl.DateTimeFormat('en', { minute: 'numeric' }).format(date)
@@ -1080,7 +1088,7 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
 
           if (type === 'string') {
             code += `
-              ${index === 0 ? 'if' : 'else if'}(typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || (obj${accessor} && typeof obj${accessor}.toISOString === "function") || obj${accessor} instanceof RegExp)
+              ${index === 0 ? 'if' : 'else if'}(obj${accessor} === null || typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || typeof obj${accessor}.toISOString === "function" || obj${accessor} instanceof RegExp)
                 ${nestedResult.code}
             `
           } else if (type === 'null') {
--- a/index.js
+++ b/index.js
@@ -277,13 +277,9 @@ function $asBooleanNullable (bool) {
 }
 
 function $asDatetime (date) {
-  if (!date) {
-    return '""'
-  }
-
   if (date instanceof Date) {
     return '"' + date.toISOString() + '"'
-  } else if (typeof date.toISOString === 'function') {
+  } else if (date && typeof date.toISOString === 'function') {
     return '"' + date.toISOString() + '"'
   } else {
     return $asString(date)
@@ -296,7 +292,7 @@ function $asDate (date) {
     var month = new Intl.DateTimeFormat('en', { month: '2-digit' }).format(date)
     var day = new Intl.DateTimeFormat('en', { day: '2-digit' }).format(date)
     return '"' + year + '-' + month + '-' + day + '"'
-  } else if (typeof date.format === 'function') {
+  } else if (date && typeof date.format === 'function') {
     return '"' + date.format('YYYY-MM-DD') + '"'
   } else {
     return $asString(date)
@@ -304,16 +300,12 @@ function $asDate (date) {
 }
 
 function $asTime (date) {
-  if (!date) {
-    return '""'
-  }
-
   if (date instanceof Date) {
     var hour = new Intl.DateTimeFormat('en', { hour: 'numeric', hour12: false }).format(date)
     var minute = new Intl.DateTimeFormat('en', { minute: 'numeric' }).format(date)
     var second = new Intl.DateTimeFormat('en', { second: 'numeric' }).format(date)
     return '"' + $pad2Zeros(hour) + ':' + $pad2Zeros(minute) + ':' + $pad2Zeros(second) + '"'
-  } else if (typeof date.format === 'function') {
+  } else if (date && typeof date.format === 'function') {
     return '"' + date.format('HH:mm:ss') + '"'
   } else {
     return $asString(date)
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -344,26 +344,3 @@ test('render a single quote as JSON', (t) => {
   t.equal(output, JSON.stringify(toStringify))
   t.ok(validate(JSON.parse(output)), 'valid schema')
 })
-
-test('should not break when the value of field is null', (t) => {
-  t.plan(2)
-
-  const schema = {
-    title: 'an object in a string',
-    type: 'object',
-    properties: {
-      updatedAt: {
-        type: ['string'],
-        format: 'date-time'
-      }
-    }
-  }
-
-  const toStringify = { updatedAt: null }
-  const validate = validator(schema)
-  const stringify = build(schema)
-  const output = stringify(toStringify)
-
-  t.equal(output, JSON.stringify(toStringify))
-  t.notOk(validate(JSON.parse(output)), 'valid schema')
-})
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -344,3 +344,26 @@ test('render a single quote as JSON', (t) => {
   t.equal(output, JSON.stringify(toStringify))
   t.ok(validate(JSON.parse(output)), 'valid schema')
 })
+
+test('should not break when the value of field is null', (t) => {
+  t.plan(2)
+
+  const schema = {
+    title: 'an object in a string',
+    type: 'object',
+    properties: {
+      updatedAt: {
+        type: ['string'],
+        format: 'date-time'
+      }
+    }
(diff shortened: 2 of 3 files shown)


index.js
```diff
@@ -279,7 +279,7 @@ function $asBooleanNullable (bool) {
 function $asDatetime (date) {
   if (date instanceof Date) {
     return '"' + date.toISOString() + '"'
-  } else if (typeof date.toISOString === 'function') {
+  } else if (date && typeof date.toISOString === 'function') {
```
Eomm, line 282: could you add this also in $asTime?

test/basic.test.js
```diff
@@ -344,3 +344,26 @@ test('render a single quote as JSON', (t) => {
...
+      }
+    }
+  }
+
+  const toStringify = { updatedAt: null }
+  const validate = validator(schema)
+  const stringify = build(schema)
+  const output = stringify(toStringify)
+
+  t.equal(output, JSON.stringify(toStringify))
```
Eomm, line 367:
```suggestion
  t.equal(output, '{"updatedAt":""}')
```

When there is a type `string` not nullable, we must return an empty string.

test/basic.test.js
```diff
@@ -344,3 +344,26 @@ test('render a single quote as JSON', (t) => {
...
+  t.notOk(validate(JSON.parse(output)), 'valid schema')
```
Eomm, line 368:
```suggestion
  t.notOk(validate(JSON.parse(output)), 'an empty string is not a date-time format')
```

test/basic.test.js
```diff
@@ -344,3 +344,26 @@ test('render a single quote as JSON', (t) => {
   t.equal(output, JSON.stringify(toStringify))
   t.ok(validate(JSON.parse(output)), 'valid schema')
 })
+
+test('should not break when the value of field is null', (t) => {
```
Eomm, line 348: we should add more tests:

The test without the type's array
```
{
    type: 'object',
    properties: {
      updatedAt: {
        type: 'string'
      }
    }
  }
```

The test with nullable
```
{
    type: 'object',
    properties: {
      updatedAt: {
        type: ['string', 'null']
      }
    }
  }
```

simoami:
@mum-never-proud what about substituting `$asString(date)` with `$asStringNullable(date)` since in that conditional branch date can still be null.

simoami:
@mum-never-proud 
I suggest that we drop the format attribute from the basic tests and add the following date format specific tests to `date.test.js`:

**Test 1**
```
test('should render a null value', (t) => {
  t.plan(2)

  const schema = {
    title: 'an object in a string',
    type: 'object',
    properties: {
      date: {
        type: ['string', 'null'],
        format: 'date-time'
      }
    }
  }
  const toStringify = { date: null }

  const validate = validator(schema)
  const output = stringify(toStringify)
  t.equal(output, JSON.stringify(toStringify))
})
```

**Test 2**
```
test('should fail validation on unexpected null values', (t) => {
  t.plan(2)

  const schema = {
    title: 'an object in a string',
    type: 'object',
    properties: {
      date: {
        type: 'string',
        format: 'date-time'
      }
    }
  }
  const toStringify = { date: null }

  const validate = validator(schema)
  const output = stringify(toStringify)
  console.log(output)
  t.equal(output, JSON.stringify(toStringify))
  t.notOk(validate(JSON.parse(output)), 'valid schema')
})
```

There are some interesting things about the second test:
- As is, it will fail because the output will be `{"date": ""}` instead of `{"date":null}` so the equality check fails. This proves my point about replacing `$asString(date)` with `$asStringNullable(date)` in my previous comment so that null is not stringified. We should replicate these 2 tests for other date/time formats and potentially also use the same patches there.
- What's interesting here is that if you specify the `type` as `['string']` instead of `'string'` the output comes out as expected `{"date":null}`. Might be good to track why it has a different behavior.

test/basic.test.js
```diff
@@ -344,3 +344,26 @@ test('render a single quote as JSON', (t) => {
...
+  t.plan(2)
+
+  const schema = {
+    title: 'an object in a string',
+    type: 'object',
+    properties: {
+      updatedAt: {
+        type: ['string'],
+        format: 'date-time'
```
simoami, line 357: Perhaps we don't specify any date formats in this file and let `date.test.js` takes care of those cases for better organization.

test/basic.test.js
simoami: @Eomm Are you sure that's the right behavior?
expected: string, given: null, output: "" or null ?
expected: date, given: null, output: <blind date> or null ? :)
expected: bool, given: null, output: false or null

index.js
```diff
@@ -1077,9 +1077,10 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
         sortedTypes.forEach((type, index) => {
           var tempSchema = Object.assign({}, schema, { type })
           var nestedResult = nested(laterCode, name, key, tempSchema, externalSchema, fullSchema, subKey)
+
           if (type === 'string') {
             code += `
-              ${index === 0 ? 'if' : 'else if'}(typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || typeof obj${accessor}.toISOString === "function" || obj${accessor} instanceof RegExp)
+              ${index === 0 ? 'if' : 'else if'}(typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || (obj${accessor} && typeof obj${accessor}.toISOString === "function") || obj${accessor} instanceof RegExp)
```
Eomm, line 1083: This cover the type:[string] in array issue

```suggestion
              ${index === 0 ? 'if' : 'else if'}(obj${accessor} === null || typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || typeof obj${accessor}.toISOString === "function" || obj${accessor} instanceof RegExp)
```

Eomm:
> This proves my point about replacing $asString(date) with $asStringNullable(date)

No, if you set `type: string` you want a string and `null` is not a string, so the type is coerced to an empty string. This is right.
You should use `nullable` or `type[string, null]` to get `null` as output.

> Might be good to track why it has a different behavior.

The fix didn't cover it, I have suggested what should solve

test/basic.test.js
Eomm: null is a type in the json schema, so it will never be returned if you don't want it.

date, is a format, so the right behaviour is to return the default string when it is null: an empty string

index.js
mum-never-proud: sure would like to know your thoughts on this

```javascript
function $asDatetime (date) {
  if (!date) {
    return '""'
  }

  if (date instanceof Date) {
    return '"' + date.toISOString() + '"'
  } else if (typeof date.toISOString === 'function') {
    return '"' + date.toISOString() + '"'
  } else {
    return $asString(date)
  }
}
```

for both `$asDateTime` & `$asTime` makes sense?

test/basic.test.js
mum-never-proud: yep for a schema like this, we can return an empty string

```javascript
updatedAt: {
  type: 'string',
  format: 'date-time'
}
```

but for 

```javascript
updatedAt: {
  type: ['string'],
  format: 'date-time'
}
```

it will return `null` ref https://github.com/fastify/fast-json-stringify/blob/1241c5d2091436200210df9404f64a82a3af1a8e/index.js#L1113-L1114

test/basic.test.js
mum-never-proud: sorry ignore this I didn't read the comment below

mum-never-proud:
I have one more concern, ain't sure if it's valid or not

I don't think the JSON format is valid with the way we are returning `null` value, the JSON parse will obviously break with the current implementation

I guess it's intentional still I think we should be returning `"null"` instead of `null`? we can let the user handle, many times a negligible error response from backend breaks the entire page

test/date.test.js
```diff
@@ -194,3 +194,70 @@ test('render a nested object in a string when type is date-format as ISOString',
...
+        format: 'date-time'
+      }
+    }
+  }
+
+  const toStringify = { updatedAt: null }
+  const stringify = build(schema)
+  const output = stringify(toStringify)
+
+  t.equal(output, '{"updatedAt":null}')
```
Eomm, line 239: Why did you remove `t.ok(validate(JSON.parse(output)), 'valid schema')` here?

test/basic.test.js
Eomm: Yes, this was wrong for sure!

index.js
Eomm: My only concern is that we are not using the `$asString` function in this way, and we are managing a string in the date function format, for this I would put this check in the `else if` as it was before

test/date.test.js
mum-never-proud: sorry my bad

index.js
mum-never-proud: we also need it in `$asDate`?

```javascript
function $asDate (date) {
  if (date instanceof Date) {
    var year = new Intl.DateTimeFormat('en', { year: 'numeric' }).format(date)
    var month = new Intl.DateTimeFormat('en', { month: '2-digit' }).format(date)
    var day = new Intl.DateTimeFormat('en', { day: '2-digit' }).format(date)
    return '"' + year + '-' + month + '-' + day + '"'
  } else if (date && typeof date.format === 'function') {
    return '"' + date.format('YYYY-MM-DD') + '"'
  } else {
    return $asString(date)
  }
}
```

Eomm: yes, it is! ✨

mum-never-proud:
my naming convention might suck, let me know for any changes :)

test/date.test.js
```diff
@@ -194,3 +194,211 @@ test('render a nested object in a string when type is date-format as ISOString',
...
+        }
+      }
+
+      const {
+        output,
+        validate
+      } = serialize(createSchema(prop), input)
+
+      t.equal(output, '{"updatedAt":null}')
+      t.ok(validate(JSON.parse(output)), 'an empty string is not a date-time format')
```
Eomm, line 363:
```suggestion
      t.ok(validate(JSON.parse(output)), 'valid schema')
```

test/date.test.js
```diff
@@ -194,3 +194,211 @@ test('render a nested object in a string when type is date-format as ISOString',
...
+        }
+      }
+
+      const {
+        output,
+        validate
+      } = serialize(createSchema(prop), input)
+
+      t.equal(output, '{"updatedAt":null}')
+      t.ok(validate(JSON.parse(output)), 'an empty string is not a date format')
```
Eomm, line 382:
```suggestion
      t.ok(validate(JSON.parse(output)), 'valid schema')
```

test/date.test.js
```diff
@@ -194,3 +194,211 @@ test('render a nested object in a string when type is date-format as ISOString',
...
+        }
+      }
+
+      const {
+        output,
+        validate
+      } = serialize(createSchema(prop), input)
+
+      t.equal(output, '{"updatedAt":null}')
+      t.ok(validate(JSON.parse(output)), 'an empty string is not a time format')
```
Eomm, line 401:
```suggestion
      t.ok(validate(JSON.parse(output)), 'valid schema')
```

PR #222, opened 2020-05-07
Renders Infinity as null

### LINKED ISSUE #221: 'Infinite' numbers are returned in JSON output
#  🐛 Bug Report

If the serialized data contains a number field with an `Infinite` value it's being serialized as `{"value": Infinite}` which produces an invalid JSON.

Is this an expected behavior?

## To Reproduce

```js
const fast = require("fast-json-stringify")
const stfy  = fast({
  type: 'object',
  required: ['a'],
  properties: {
    a: { type: 'number' }
  }
})
const obj = { a: Infinity }
console.log(stfy(obj)) // => "{\"a\":-Infinity}"
console.log(JSON.stringify(obj)) // => "{\"a\":null}"
```
Runkit: https://runkit.com/maximilian-krauss/5eb27a8a3651fb001a5b6521

## Expected behavior

Output should be `null`.

## Your Environment

- *node version*: 12
- *fastify version*: >=2.0.0
- *os*: Mac

maximilian-krauss:
#### Checklist

Closes #221

### DIFF
--- a/test/infinity.test.js
+++ b/test/infinity.test.js
@@ -1,0 +1,47 @@
+'use strict'
+
+const t = require('tap')
+const test = t.test
+const proxyquire = require('proxyquire')
+const build = proxyquire('..', { long: null })
+
+test('render Infinity as null (integer)', (t) => {
+  t.plan(1)
+
+  const schema = {
+    title: 'integer',
+    type: 'integer'
+  }
+
+  const stringify = build(schema)
+  const output = stringify(Infinity)
+
+  t.equal(output, 'null')
+})
+test('render Infinity as null (number)', (t) => {
+  t.plan(1)
+
+  const schema = {
+    title: 'number',
+    type: 'number'
+  }
+
+  const stringify = build(schema)
+  const output = stringify(Infinity)
+
+  t.equal(output, 'null')
+})
+
+test('render negative Infinity as null', (t) => {
+  t.plan(1)
+
+  const schema = {
+    title: 'integer',
+    type: 'integer'
+  }
+
+  const stringify = build(schema)
+  const output = stringify(-Infinity)
+
+  t.equal(output, 'null')
+})
--- a/index.js
+++ b/index.js
@@ -252,7 +252,7 @@ function $asIntegerNullable (i) {
 
 function $asNumber (i) {
   var num = Number(i)
-  if (isNaN(num)) {
+  if (isNaN(num) || !isFinite(num)) {
     return 'null'
   } else {
     return '' + num


test/infinity.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+  const schema = {
+    title: 'integer',
+    type: 'integer'
+  }
+
+  const stringify = build(schema)
+  const output = stringify(Infinity)
+
+  t.equal(output, 'null')
```
Eomm, line 19: I think this should be `0`

For this schema:

```
{
  title: 'integer',
  type: 'integer'
}
```

the user wants only integer as output

For this one instead, user accepts `null` value also
```
{
  title: 'integer',
  nullable: true,
  type: 'integer'
}
```

maximilian-krauss: Sure, will do :)

jsumners: @Eomm I'm not sure it should be `0`. That has a very different meaning than `Infinity`. But I'm also not sure what to do here. The spec merely says:

https://tools.ietf.org/html/rfc8259#section-6
> Numeric values that cannot be represented in the grammar below (such
   as Infinity and NaN) are not permitted.

maximilian-krauss:
Reverted PR back to draft, I added more tests and saw that nested objects are not rendered as expected. Will reopen when I'm done

test/infinity.test.js
Eomm: You are right @jsumners 

I would just don't output the field (instead of zero) and output `null` when it is nullable.
I say so because when a JSON doesn't fit a schema, `fast-json-stringify` simply doesn't process it (like `foo: {type: 'string'}` and but then foo is an array)

maximilian-krauss: @Eomm  ok cool, then I will change it that way.
One question I have though, what's then the right output if I pass `{ type: 'integer' }` as schema and `Infinity` as value? `` or `undefined`?

Eomm: Edge cases 🎉 

The target is to serialize only a number (odd, but it is ok):

```js
schema = { type: 'integer' }
serialize = fjs(schema)
serialize(Infinity)
```

I would avoid undefined because this would break clients that do `JSON.parse(undefined)`
For this reason, I would say `''`

Do you agree @jsumners ?

jsumners: It should just be omitted or generate an error. The JSON Schema spec defers to the JSON spec — https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-02#section-6.1

Therefore, `{ foo: Infinity }` is invalid.

Eomm: We do so when want to serialize a json, but when we try to serialize a number only?
It is doable for this module and I quite agree to throw a "CANNOT SERIALIZE" error.

It would be the first error thrown TBH

jsumners: It is clearly an error according to spec.

mcollina:
Are you still interested in pursuing this?

maximilian-krauss:
Hey @mcollina, oh I totally forgot about this one. Yes, I would give it another spin since we now know that it should be an error case :)

PR #226, opened 2020-06-09
feat: add required check for object key not in properties

### LINKED ISSUE #225: Required property not set in properties
## 🐛 Bug Report

A required property not set in properties is not checked also if set as required. This whether it is set with additionalProperties to true and to false.

## To Reproduce

```js
test('object with required field not in properties schema', (t) => {
  const schema = {
    title: 'object with required field',
    type: 'object',
    properties: {
      num: {
        type: 'integer'
      }
    },
    additionalProperties: true,
    required: ['str']
  }

  try {
    stringify({})
    t.fail()
  } catch (e) {
    t.is(e.message, 'str is required!')
    t.pass()
  }
}
```

## Expected behavior

I suppose this test should pass (ref of the required property here https://tools.ietf.org/html/draft-zyp-json-schema-03#section-5.7).

Otherwise, since it is an edge case usually due to a badly generated schema, this behavior could also be documented.

## Your Environment

- *node version*: 12
- *os*: Mac

davidebianchi:
This PR closes #225

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -631,6 +631,8 @@ function buildCode (schema, code, laterCode, name, externalSchema, fullSchema) {
     schema = refFinder(schema.$ref, fullSchema, externalSchema)
   }
 
+  var required = schema.required
+
   Object.keys(schema.properties || {}).forEach((key, i, a) => {
     if (schema.properties[key].$ref) {
       // if the schema object is deep in the tree, we must resolve the ref in the parent scope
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -731,9 +735,33 @@ function buildCode (schema, code, laterCode, name, externalSchema, fullSchema) {
...
     }
   })
+
+  if (required && required.length > 0) {
+    code += 'var required = ['
+    for (var i = 0; i < required.length; i++) {
+      if (i > 0) {
+        code += ','
+      }
+      code += `'${required[i]}'`
```
mcollina, line 745: This should be properly escaped.

test/allof.test.js
```diff
@@ -178,10 +178,10 @@ test('object with nested allOfs', (t) => {
       })
       t.is(value, '{"id1":1,"id2":2,"id3":3}')
     } catch (e) {
-      t.fail()
+      t.fail(e.message)
     }
   } catch (e) {
-    t.fail()
+    t.fail(e.message)
```
mcollina, line 184: are those changes necessary?

davidebianchi:
Actual benchmark does not have required key in schema (nor required schema not present in properties), so there is no change to the generated code actually tested.

test/allof.test.js
davidebianchi: No, I removed

index.js
davidebianchi: I escaped using `$asString` function, is this the correct way to escape?

Issue #230, opened 2020-06-12
Mixed element in array


Eomm:
##  🐛 Bug Report

Only the first item in `anyOf` is evaluated: 

```
{
  $schema: 'http://json-schema.org/draft-07/schema#',
  type: 'array',
  items: {
    anyOf: [
      {
        type: 'object',
        properties: {
          savedId: { type: 'string' }
        }
      },
      {
        type: 'object',
        properties: {
          error: { type: 'string' }
        }
      }
    ]
  }
}
```

## To Reproduce

```js
const fjs = require('fast-json-stringify')

const S = require('fluent-schema')

const schema = S.array().items(
  S.anyOf([
    S.object().prop('savedId', S.string()),
    S.object().prop('error', S.string())
  ])
)

console.log(JSON.stringify(schema.valueOf(), null, 2))
const s = fjs(schema.valueOf())

console.log(s([{ savedId: 'great' }, { error: 'ops' }]))

// shows only:
// [{"savedId":"great"},{}]
```

## Expected behavior

The output should be:

```js
[{"savedId":"great"},{"error": "ops"}]
```

## Your Environment

- *node version*: 12
- *version*: 2.0.2

Maybe related with #179

artemruts:
I'll take a stab at this if no one minds 🙇

jsumners:
Go for it.

PR #235, opened 2020-06-23
Adds test case for anyOf with multiple array items

artemruts:
#### Checklist

I don't think #230 is technically an issue, I think it's just how AJV's `properties` validation keyword works.

### Explanation

Assume we have this schema:

```js
{
  title: 'type array that may have any of declared items',
  type: 'array',
  items: {
    anyOf: [
      {
        type: 'object',
        properties: {
          savedId: {
            type: 'string'
          }
        }
      },
      {
        type: 'object',
        properties: {
          error: {
            type: 'string'
          }
        }
      }
    ]
  }
}
```

and test array:
```js
[{ savedId: 'great' }, { error: 'oops' }]
```

If we `build` the schema it'll correctly precompile it with the following `validate` calls to AJV:

debug

but AJV's docs on `properties` say:

> properties keyword does not require that the properties mentioned in it are present in the object
> -- https://ajv.js.org/keywords.html#properties

Because of that:
- first `ajv.validate` call matches any object
- only the first precompiled [buildObject](https://github.com/fastify/fast-json-stringify/blob/master/index.js#L854) function is ever called
- call to incorrect `BuildObject` function results in `{}`
- we get `[{"savedId":"great"},{}]`

### Solution ?

```js
// ...
anyOf: [
{
  type: 'object',
  properties: {
    savedId: {
      type: 'string'
    }
  },
  required: ['savedId']
},
// ...
```

Adding `required` keyword to `anyOf` objects certainly works but is this the only solution?

### DIFF
--- a/index.js
+++ b/index.js
@@ -1024,9 +1024,9 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
   }
 
   if (schema.type === undefined) {
-    var inferedType = inferTypeByKeyword(schema)
-    if (inferedType) {
-      schema.type = inferedType
+    var inferredType = inferTypeByKeyword(schema)
+    if (inferredType) {
+      schema.type = inferredType
     }
   }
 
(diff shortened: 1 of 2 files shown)


artemruts:
I'll open a PR with docs then, thank you!

PR #237, opened 2020-07-07
fix cutting off arrays with more than 100 items in anyOf and oneOf

johakr:
Node.js [util.inspect](https://nodejs.org/api/util.html#util_util_inspect_object_options) defaults to `maxArrayLength` of 100. Hence, arrays with more than 100 items inside `anyOf` or `oneOf` (e. g. large enums) were cutoff, i.e. build function would throw an exception.

Setting `maxArrayLength` to `null` fixes this issue.

#### Checklist


test/anyof.test.js
```diff
@@ -435,3 +435,198 @@ test('anyOf looks for all of the array items', (t) => {
...
+            'XAU',
+            'XCD',
+            'XDR',
+            'XOF',
+            'XPF',
+            'YER',
+            'ZAR',
+            'ZMK',
+            'ZMW',
+            'ZWL'
```
mcollina, line 617: Can you please generate such a long array with code? Thanks.

test/oneof.test.js
```diff
@@ -397,3 +397,198 @@ test('oneOf and $ref: multiple levels should throw at build.', (t) => {
...
+            'XAU',
+            'XCD',
+            'XDR',
+            'XOF',
+            'XPF',
+            'YER',
+            'ZAR',
+            'ZMK',
+            'ZMW',
+            'ZWL'
```
mcollina, line 579: Same

johakr:
Arrays are now generated with code. :)

Issue #238, opened 2020-06-10
$ref to /definitions in same JSON Schema not working for serialization


ahgentil:
## 🐛 Bug Report

`$ref`s to `/definitions` in same JSON Schema work for validation but not for serialization - the 'absolute path' needs to be used instead.

I was not sure if this was a `fastify` issue or a `fast-json-stringify` issue, so I decided to post it here.

## To Reproduce

```js
fastify.register((instance, opts, done) => {
  instance.addSchema({
    $id: 'http://example.com/plugin1/',
    definitions: {
      item1Id: {
        type: 'string',
        enum: ['1', '2', '3']
      },
      item1IdParam: {
        type: 'object',
        required: ['id'],
        properties: {
          id: { $ref: '#/definitions/item1Id'}
        },
        additionalProperties: false
      },
      item1: {
        type: 'object',
        properties: {
          id: { $ref: '#/definitions/item1Id' }, // does not work with schema's response set in the route
          // id: { $ref: 'http://example.com/plugin1/#/definitions/item1Id' }, // works
          name: { type: 'string' }
        },
        additionalProperties: false
      }
    }
  })  

  instance.get('/item1/:id',
    { schema: {
      params: { $ref: 'http://example.com/plugin1/#/definitions/item1IdParam' },
      response: { 200: { $ref: 'http://example.com/plugin1/#/definitions/item1' } }
    }},
    (req, reply) => { reply.send({}) }
  )

  done()
})
```

## Expected behavior

$ref to /definitions in same JSON Schema to always work

## Your Environment

- *node version*: 12
- *fastify version*: 3rc
- *os*: Mac

ahgentil:
Here's a test that does not pass:

```javascript
test('should resolve relative $ref in item1', (t) => {
  t.plan(1)

  const externalSchema = {
    'http://example.com/plugin1/': {
      $id: 'http://example.com/plugin1/',
      definitions: {
        item1Id: { type: 'string' },
        item1: {
          type: 'object',
          properties: {
            // does not pass the test
            id: { $ref: '/#/definitions/item1Id' },
            // passes the test
            // id: { $ref: 'http://example.com/plugin1/#/definitions/item1Id' },
            name: { type: 'string' }
          },
          additionalProperties: false
        }
      }
    }
  }

  const schema = { $ref: 'http://example.com/plugin1/#/definitions/item1' }

  try {
    build(schema, { schema: externalSchema })
    t.pass()
  } catch (e) {
[...]

PR #239, opened 2020-07-16
Fix relative json paths in external schema

paolochiodi:
This PR is intended to fix https://github.com/fastify/fast-json-stringify/issues/238.

The problem in the current implementation is that relative JSON pointers in `$ref`s like `#/definitions/def` always search in the main schema, even if the `$ref` is found in an external schema.

With this change the code will keep track of the current parent (or "context", if you will) so that subsequent `$ref`s will search in the appropriate schema.

There was an original attempt to fix this in https://github.com/fastify/fast-json-stringify/pull/185/files, but that was limited to some special conditions (i.e. only in object properties).
Also the assumptions in https://github.com/fastify/fast-json-stringify/pull/185/files#diff-168726dbe96b3ce427e7fedce31bb0bcR537 about existence of `schema.definitions` is wrong as using `definitions` is customary, but not mandatory (see https://json-schema.org/understanding-json-schema/structuring.html#reuse).

With this PR I hope to provide a comprehensive solution that correctly handle relative pointers in `$ref` for all scenarios.

#### Checklist

- [ ] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1153,6 +1153,13 @@ function nested (laterCode, name, key, location, subKey, isArray) {
         anyOfLocations.forEach((location, index) => {
           var nestedResult = nested(laterCode, name, key, location, subKey !== '' ? subKey : 'i' + index, isArray)
 
+          // Since we are only passing the relevant schema to ajv.validate, it needs to be full dereferenced
+          // otherwise any $ref pointing to an external schema would result in an error.
+          // Full dereference of the schema happens as side effect of two functions:
+          // 1. `dereferenceOfRefs` loops through the `schema.anyOf`` array and replaces any top level reference
+          // with the actual schema
+          // 2. `nested`, through `buildCode`, replaces any reference in object properties with the actual schema
+          // (see https://github.com/fastify/fast-json-stringify/blob/6da3b3e8ac24b1ca5578223adedb4083b7adf8db/index.js#L631)
           code += `
             ${index === 0 ? 'if' : 'else if'}(ajv.validate(${require('util').inspect(location.schema, { depth: null, maxArrayLength: null })}, obj${accessor}))
               ${nestedResult.code}
@@ -1167,6 +1174,8 @@ function nested (laterCode, name, key, location, subKey, isArray) {
         var oneOfLocations = dereferenceOfRefs(location, 'oneOf')
         oneOfLocations.forEach((location, index) => {
           var nestedResult = nested(laterCode, name, key, location, subKey !== '' ? subKey : 'i' + index, isArray)
+
+          // see comment on anyOf about derefencing the schema before calling ajv.validate
           code += `
             ${index === 0 ? 'if' : 'else if'}(ajv.validate(${require('util').inspect(location.schema, { depth: null, maxArrayLength: null })}, obj${accessor}))
               ${nestedResult.code}
--- a/index.js
+++ b/index.js
@@ -642,11 +642,15 @@ function refFinder (ref, location) {
     findBadKey(schema, walk.slice(1))
   }
 
-  return result.$ref ? refFinder(result.$ref, {
-    schema: schema,
-    root: root,
-    externalSchema: externalSchema
-  }) : {
+  if (result.$ref) {
+    return refFinder(result.$ref, {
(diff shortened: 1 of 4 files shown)


paolochiodi:
In the current state the PR has all tests green, but I only added a test for the issue originally posted.
I want to add tests for other scenarios (i.e. `ref$` in array items), cleanup the code and run the benchmarks.

I do have a questions: what js feature do we want to use?
I see most of the code uses `var`, but some places also have `const`. Are we aiming for the broadest possible support?

mcollina:
> I see most of the code uses var, but some places also have const. Are we aiming for the broadest possible support?



`let` was slower than `var`. `const` are ok.

Overall you can use most JS features when _generating_ the code, but then stick to fast things in the _generated_ code.

index.js
```diff
@@ -1061,25 +1134,27 @@ function nested (laterCode, name, key, schema, externalSchema, fullSchema, subKe
...
       `
       break
     case undefined:
       if ('anyOf' in schema) {
-        dereferenceOfRefs(schema, externalSchema, fullSchema, 'anyOf')
-        schema.anyOf.forEach((s, index) => {
-          var nestedResult = nested(laterCode, name, key, s, externalSchema, fullSchema, subKey !== '' ? subKey : 'i' + index, isArray)
+        // beware: dereferenceOfRefs has side effects and changes schema.anyOf
+        var anyOfLocations = dereferenceOfRefs(location, 'anyOf')
+        anyOfLocations.forEach((location, index) => {
```
paolochiodi, line 1153: This was by far the most complicated part to understand and fix.

The reason is that it heavily relies on side effects: 
- `dereferenceOfRefs` changes `schema.anyOf`
- calls to `buildCode` also change the schema (replacing any property with a reference to the actual schema, see here: https://github.com/fastify/fast-json-stringify/blob/master/index.js#L631)

This is necessary because ajv will need to have the "dereferenced" schema  at run time to work properly.

My problem was that changing just the schema would loose the full location/context and tests with multiple external schema in anyOf wouldn't work (this would fail: https://github.com/fastify/fast-json-stringify/pull/239/files#diff-2dbacb6426a8e9d68e93312ea682b74dR400).

I think the best solution would have been to dereference the entire schema at the start and avoid dereferencing it piece by piece - but that would require a complete rewrite.

Here I opted for an hack: `dereferenceOfRefs` not only changes `schema.anyOf`, but also returns an array of locations which contain the context about where the dereferenced schema was found.

I also had a different implementation where I was not calling dereferenceOfRefs but was replacing the schema in place:

```js
// var locations = dereferenceOfRefs(location, 'anyOf')
schema.anyOf.forEach((s, index) => {
  var sLocation = mergeLocation(location, { schema: s })
  if (s.$ref) {
    sLocation = refFinder(s.$ref, sLocation)
    s = sLocation.schema
  }

  schema.anyOf[index] = s

  var nestedResult = nested(laterCode, name, key, sLocation, subKey !== '' ? subKey : 'i' + index, isArray)

  code += `
    ${index === 0 ? 'if' : 'else if'}(ajv.validate(${require('util').inspect(s, { depth: null, maxArrayLength: null })}, obj${accessor}))
      ${nestedResult.code}
  `
  laterCode = nestedResult.laterCode
})
```

In the end I've discarded it because it seemed even more unreadable the the one proposed in this PR.

All of this applies equally to `oneOf`

mcollina: Can you:

1. add some of that explanation in the comment before `deferenceOfRefs`?
2. create a new issue explaining this problem and the full solution.

paolochiodi: Added expanded comment right before the call of `ajv.validate`: https://github.com/fastify/fast-json-stringify/pull/239/files#diff-168726dbe96b3ce427e7fedce31bb0bcR1156-R1162

paolochiodi: And I've also created the issue explaining the problem: https://github.com/fastify/fast-json-stringify/issues/241

As a bonus I also added this possible bug: https://github.com/fastify/fast-json-stringify/issues/242

Issue #193, opened 2019-11-06
Support multiple types in one schema


KostyaZgara:
## 🚀 Feature Proposal

Hi,  I'm using `fastify` and I know that `fastify` use `fast-json-stringify` underhood in serialization. Will be better if `fast-json-stringify` will support multiple types. For example

```js
{
  type: ['object', 'string'],
  properties: {
    foo: {
      type: 'string'
    }
}
```

So property `foo` can be string or `object`

## Motivation

I'm using `fastify` with MongoDB where all ids are `ObjectId` and `fast-json-stringify` can't properly determine that `ObjectId` can be serialized to a simple string. Also, a field that contains `ObjectId` can be expanded and be a plain object with nested properties. So one field can be either `ObjectId` that should be properly serialized to `string` or plain object with nested properties that should be serialized as `object`.
Also, I have tried to use `anyOf` keyword but I did get the same result.

## Example
I can provide a code example with current behavior:

```js
const { ObjectId } = require('bson');
const fastJson = require('fast-json-stringify');

const stringify = fastJson({
  type: 'object',
  properties: {
    user: {
      type: ['object', 'string'],
      properties: {
        id: {
          type: ['object', 'string'],
        },
        name: {
          type: 'string',
        },
      },
    },
  },
});

const stringifyAnyOf = fastJson({
  type: 'object',
  properties: {
    user: {
      anyOf: [
        { type: 'string' },
        {
          type: 'object',
          properties: {
            id: {
              type: ['object', 'string'],
            },
            name: {
              type: 'string',
            },
          },
        }
      ],
    },
  },
});

const id = new ObjectId();

const dataWithObjectId = {
  user: id,
}

const dataWithExpandedUser = {
  user: {
    id,
    name: 'John',
  },
};

console.log(`Current id: ${id}`)
// serialize with type: ['object', 'string']
console.log(stringify(dataWithObjectId));
console.log(stringify(dataWithExpandedUser));
// serialize with anyOf
console.log(stringifyAnyOf(dataWithObjectId));
console.log(stringifyAnyOf(dataWithExpandedUser));
```

### Likn to repl
https://repl.it/@KostyaZgara/fast-json-stringify 

Current behavior:
```json
Current id: 5dc2a3ba50a3a8d8f1b73f30
{"user":{}}
{"user":{"id":{},"name":"John"}}
{"user":{}}
{"user":{"id":{},"name":"John"}}
```

Expected behavior:
```json
[...]

PR #263, opened 2020-10-05
feat: simultaneously test example

Eomm:
Ref #193 

Details https://github.com/fastify/fast-json-stringify/issues/193#issuecomment-703804249

### DIFF
--- a/index.js
+++ b/index.js
@@ -1199,7 +1199,7 @@ function nested (laterCode, name, key, location, subKey, isArray) {
 
           if (type === 'string') {
             code += `
-              ${index === 0 ? 'if' : 'else if'}(obj${accessor} === null || typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || typeof obj${accessor}.toISOString === "function" || obj${accessor} instanceof RegExp)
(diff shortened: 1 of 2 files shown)


mcollina:
CI is failing, am I missing anything?

Eomm:
This PR want to understand the solution for the issue.

Right now you commented on the issue:

> If type: ['object', 'string'], an object with toString() will be rendered as an object. If type: ['string', 'object'], we check typeof obj === 'string' || typeof obj === 'object' && Object.prototype.hasOwnProperty.call(obj, 'toString').

I'm not sure it will work because our case is: "there is an object and I want to render it like a string sometimes and like an object sometimes". The test added show this example

index.js
```diff
@@ -1199,7 +1199,7 @@ function nested (laterCode, name, key, location, subKey, isArray) {
 
           if (type === 'string') {
             code += `
-              ${index === 0 ? 'if' : 'else if'}(obj${accessor} === null || typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || typeof obj${accessor}.toISOString === "function" || obj${accessor} instanceof RegExp)
+              ${index === 0 ? 'if' : 'else if'}(obj${accessor} === null || typeof obj${accessor} === "${type}" || obj${accessor} instanceof Date || typeof obj${accessor}.toISOString === "function" || obj${accessor} instanceof RegExp || typeof obj${accessor} === "object")
```
mcollina, line 1202: I think you are missing a `hasOwnProperty` call here.

mcollina (approved):
lgtm

RafaelGSS (approved):
LGTM.

Issue #249, opened 2020-08-06
null array value throws TypeError: Cannot read property 'length' of null


bburke4:
## 🐛 Bug Report

When a null value is expected to be an array it throws `TypeError: Cannot read property 'length' of null`. 

This looks very similar to: https://github.com/fastify/fast-json-stringify/issues/218

I love fast-json-stringify and fastify. Thanks a lot!

## To Reproduce

```js
const fastJson = require('fast-json-stringify');

fastJson({
  type: 'object',
  properties: {
    arr: {
      type: 'array',
      items: {
        type: 'number'
      }
    }
  }
})({ arr: null })
```
```
Uncaught TypeError: Cannot read property 'length' of null
    at $arr$mainarr (eval at build (/Users/b.burke/dev/play/node_modules/fast-json-stringify/index.js:158:20), <anonymous>:144:17)
    at $main (eval at build (/Users/b.burke/dev/play/node_modules/fast-json-stringify/index.js:158:20), <anonymous>:132:17)
```

Putting the 'array' type in an array yields different results...
```js
fastJson({
  type: 'object',
  properties: {
    arr: {
      type: ['array'],
      items: {
        type: 'number'
      }
    }
  }
})({ arr: null });

'{"arr":null}'
```

using `undefined` rather than `null` yields different results...
```js
fastJson({
  type: 'object',
  properties: {
    arr: {
      type: 'array',
      items: {
        type: 'number'
      }
    }
  }
})({ arr: undefined });

'{"arr":null}'
```

## Expected behavior

I would expect it to either print 'null' or coerce it to an empty array. I tried setting `default: []` but it didn't make a difference. setting `type: ['array', 'null']` does make it print null instead.

## Your Environment

- *node version*: v12.16.1
- *fastify version*: 2.2.3
- *os*: Mac

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

alemagio:
Hi, is anyone working on this?

mcollina:
Maybe yourself! :P

PR #265, opened 2020-10-12
fix: passing null object to type array defaults to empty array (#249)

alemagio:
#### Checklist

      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.

### DIFF
--- a/test/typesArray.test.js
+++ b/test/typesArray.test.js
@@ -417,3 +417,22 @@ test('object that is simultaneously a string and a json switched', (t) => {
   const valueObj = stringify({ simultaneously: { foo: likeObjectId } })
   t.is(valueObj, '{"simultaneously":{"foo":"hello"}}')
 })
+
+test('should default to [] when type is array and object is null', (t) => {
+  t.plan(1)
+  const schema = {
+    type: 'object',
+    properties: {
+      arr: {
+        type: 'array',
+        items: {
+          type: 'number'
+        }
+      }
+    }
+  }
+
+  const stringify = build(schema)
+  const valueStr = stringify({ arr: null })
+  t.is(valueStr, '{"arr":[]}')
+})
--- a/test/typesArray.test.js
+++ b/test/typesArray.test.js
@@ -418,7 +418,7 @@ test('object that is simultaneously a string and a json switched', (t) => {
   t.is(valueObj, '{"simultaneously":{"foo":"hello"}}')
 })
 
-test('should default to [] when type is array and object is null', (t) => {
+test('should throw an error when type is array and object is null', (t) => {
   t.plan(1)
   const schema = {
     type: 'object',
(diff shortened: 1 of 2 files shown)


zekth (approved):
lgtm

RafaelGSS (commented):
Not sure if this is unexpected behavior. The following snippet sent an empty array by default
```js
const schema = {
  type: 'object',
  properties: {
    arr: {
      type: 'array',
      items: {
        type: 'number'
      },
      default: [],
    }
  }
}

const stringify = fastJson(schema)
const valueStr = stringify({}) // {"arr":[]}
```

Although throwing an error doesn't look good.

alemagio:
@RafaelGSS Added a unit test with your case.
I'm not having any error, am I missing somenthing?

RafaelGSS:
No, I said the current implementation(master) seems to the expected behavior since `null` is sent. I'll wait for some contributors to land this.

Btw the last test case is not necessary, is already being tested [here](https://github.com/fastify/fast-json-stringify/blob/master/test/defaults.test.js#L214)

alemagio:
> No, I said the current implementation(master) seems to have expected behavior since `null` is sent. I'll wait for some contributors to give their opinion.
> 
> Btw the last test case is not necessary, is already being tested [here](https://github.com/fastify/fast-json-stringify/blob/master/test/defaults.test.js#L214)

Oh ok.
I'll remove last test in the meantime.

test/typesArray.test.js
```diff
@@ -417,3 +417,22 @@ test('object that is simultaneously a string and a json switched', (t) => {
...
+    type: 'object',
+    properties: {
+      arr: {
+        type: 'array',
+        items: {
+          type: 'number'
+        }
+      }
+    }
+  }
```
mcollina, line 433: Are you sure passing null for the array would validate this schema? Can you check with ajv?

My understanding is that null is not a valid array.

mcollina (changes requested):

test/typesArray.test.js
zekth:
```javascript
const schema = {
  type: 'object',
  properties: {
    foo: {
      $id: '#/properties/foo',
      type: 'array',
      default: [],
      items: {
        $id: '#/properties/foo/items'
      }
    }
  }
}
const obj = {
  foo: null
}
var Ajv = require('ajv')
var ajv = new Ajv()
var validate = ajv.compile(schema)
var valid = validate(obj)
if (!valid) console.log(validate.errors)
// output
// [
//   {
//     keyword: 'type',
//     dataPath: '.foo',
//     schemaPath: '#/properties/foo/type',
//     params: { type: 'array' },
//     message: 'should be array'
//   }
// ]

```
Indeed

alemagio:
I followed the issue, too strictly actually.
Tried with ajv and null is not valid when type is array.
I think a good way to handle this is to throw an error with a more informative message.
What do you think?

mcollina:
> I think a good way to handle this is to throw an error with a more informative message.

Yes, go for it!

alemagio:
Updated the code.
Now in case something expected to be an array is something else a clearer error is thrown.

alemagio:
Would you mind adding `hacktoberfest-accepted` label to this PR so I can complete my Hacktoberfest? :smile:

alemagio:
Thanks 😁

PR #273, opened 2020-11-18
chore: add dependabot automerge

salmanm:
- Upgraded dependabot config
- Added dependabot automerge action to the workflow


mcollina (approved):
lgtm

PR #278, opened 2021-01-06
add float truncation

### LINKED ISSUE #277: Integer are not truncated
## 🐛 Bug Report

This script:

```js
const fjs = require("fast-json-stringify")
const serial = fjs({
  type: 'object',
  properties: { foo: { type: 'integer' } }
})
console.log(serial({ foo: 1.2 }))
```

Produces: `{"foo":1.2}`

## Expected behavior

A truncated integer: `{"foo":1}`

I think a rounding function could be provided as input aswell like:

```
const serial = fjs({
  type: 'object',
  properties: { foo: { type: 'integer' } }
}, { rounding: Math.floor } )
```

Do you have any thought on this?

## Your Environment

- *fast-json-stringify*: 2.3.0

Eomm:
Fix #277 

Note that to implement a `rounding` input option the most straightforward way is to generate the output function calling a global named function.

```js
  let customAsInteger = $asInteger.toString()
  if (options.rounding) {
    customAsInteger = customAsInteger.replace('parseInt', options.rounding)
  }
```

In this case the input would be the ugly (note the Math.round as 'string`):

```
const debugCompiled = fastJson(schema, { debugMode: true, rounding: 'Math.round' })
```

Or:

```
function round(in){ return Math.round(in) }
const debugCompiled = fastJson(schema, { debugMode: true, rounding: round })
```

Where

```diff
-  customAsInteger = customAsInteger.replace('parseInt', options.rounding)
+ customAsInteger = customAsInteger.replace('parseInt', options.rounding.name)
```

Which one do you think would be better? Or do you see nicest solutions?

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,7 +83,8 @@ function build (schema, options) {
     `
   } else {
     code += `
-      var $asInteger = $asNumber
+      var isLong = false
+      ${$asInteger.toString()}
     `
   }
 
@@ -239,8 +240,11 @@ function $asInteger (i) {
     return i.toString()
   } else if (typeof i === 'bigint') {
     return i.toString()
-  } else {
+  } else if (Number.isInteger(i)) {
     return $asNumber(i)
+  } else {
+    // if the output is NaN the type is coerced to int 0
+    return $asNumber(parseInt(i) || 0)
   }
 }
 
--- a/index.js
+++ b/index.js
@@ -54,6 +54,15 @@ function build (schema, options) {
     }
   }
 
+  var intParseFunctionName = 'trunc'
+  if (options.rounding) {
+    if (['floor', 'ceil', 'round'].includes(options.rounding)) {
+      intParseFunctionName = options.rounding
+    } else {
+      throw new Error(`Unsupported integer rounding method ${options.rounding}`)
+    }
+  }
+
   /* eslint no-new-func: "off" */
   var code = `
     'use strict'
@@ -76,6 +85,8 @@ function build (schema, options) {
     ${$asBooleanNullable.toString()}
 
     var isLong = ${isLong ? isLong.toString() : false}
+
+    function parseInteger(int) { return Math.${intParseFunctionName}(int) }
     `
 
   var location = {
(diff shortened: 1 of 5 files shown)


mcollina:
I would support only strings: `'round'`, `'floor'`, `'ceil'`.

index.js
```diff
@@ -239,8 +230,11 @@ function $asInteger (i) {
     return i.toString()
   } else if (typeof i === 'bigint') {
     return i.toString()
-  } else {
+  } else if (Number.isInteger(i)) {
     return $asNumber(i)
+  } else {
+    // if the output is NaN the type is coerced to int 0
+    return $asNumber(parseInt(i) || 0)
```
jsumners, line 237:
```suggestion
    return $asNumber(parseInt(i) || 0, 10)
```

jsumners:
There isn't any concept of "rounding" defined in the spec https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.2

mcollina:
so you think the default behavior should be to throw if it's not an integer?

jsumners:
I believe that is the consensus we reached the last time we discussed how to handle non-spec compliant values (i.e. `undefined`).

mcollina:
I would actually prefer to not do anything in this case if we want to throw.

jsumners:
Looking over this again, it seems this PR is making use of an undocumented second parameter to the `FJS` function: `const stringifierFunction = FJS(jsonSchema, {...options})`. So I am okay with this PR since it isn't adding new syntax to the JSON Schema spec. _However_, this PR should also update the documentation for this mysterious `options` object or there should be a fast follow-up PR that does so.

Eomm:
Yes, this pr doesn't force the user to add non-spec fields to the input schema.

This pr force the type coertion of those properties defined by user as "integer"

The second parameter is documented to add external schemas and debug the stringify function, but I agree on a follow up pr to document better this option that is spread across the readme and I will take care of it 👍🏼

Eomm:
Are there any concerns for this PR?

jsumners (approved):

mcollina (approved):
lgtm

PR #284, opened 2021-01-14
feat: make long dependency configurable

gajus:
#### Checklist

#### Benchmark

```
FJS creation x 62,633 ops/sec ±0.99% (91 runs sampled)
CJS creation x 166,758 ops/sec ±1.18% (89 runs sampled)
JSON.stringify array x 5,083 ops/sec ±1.22% (92 runs sampled)
fast-json-stringify array x 7,234 ops/sec ±1.00% (90 runs sampled)
compile-json-stringify array x 6,601 ops/sec ±1.45% (88 runs sampled)
JSON.stringify long string x 13,391 ops/sec ±1.01% (88 runs sampled)
fast-json-stringify long string x 13,245 ops/sec ±1.06% (89 runs sampled)
compile-json-stringify long string x 13,358 ops/sec ±1.08% (89 runs sampled)
JSON.stringify short string x 10,635,798 ops/sec ±1.36% (88 runs sampled)
fast-json-stringify short string x 31,372,088 ops/sec ±1.23% (89 runs sampled)
compile-json-stringify short string x 28,003,490 ops/sec ±1.00% (89 runs sampled)
JSON.stringify obj x 2,347,149 ops/sec ±1.56% (86 runs sampled)
fast-json-stringify obj x 6,576,019 ops/sec ±1.01% (88 runs sampled)
compile-json-stringify obj x 15,935,888 ops/sec ±0.93% (90 runs sampled)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,12 @@ const stringify = fastJson(schema, { schema: externalSchema })
 #### Long integers
 By default the library will handle automatically [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) from Node.js v10.3 and above.
 If you can't use BigInts in your environment, long integers (64-bit) are also supported using the [long](https://github.com/dcodeIO/long.js) module.
+
+In order to use long integers you have to:
+
+1. Install `long` dependency
+1. Pass `{useLong: true}` option when building schema
+
 Example:
 ```javascript
 // => using native BigInt
@@ -524,7 +530,7 @@ const stringify = fastJson({
       type: 'integer'
     }
   }
-})
+}, {useLong: true})
 
 const obj = {
   id: Long.fromString('18446744073709551615', true)
--- a/index.js
+++ b/index.js
@@ -8,13 +8,6 @@ const merge = require('deepmerge')
 const validate = require('./schema-validator')
 let stringSimilarity = null
 
-let isLong
-try {
-  isLong = require('long').isLong
-} catch (e) {
-  isLong = null
-}
-
 const addComma = `
   if (addComma) {
     json += ','
@@ -44,8 +37,15 @@ function mergeLocation (source, dest) {
   }
 }
 
+let isLong
+
 function build (schema, options) {
   options = options || {}
+
+  if (options.useLong) {
+    isLong = require('long').isLong
+  }
+
   isValidSchema(schema)
   if (options.schema) {
     // eslint-disable-next-line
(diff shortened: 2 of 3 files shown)


RafaelGSS (approved):
LGTM.

jsumners (approved):

L2jLiga (approved):
LGTM!

One small question - is it semver-major?

index.js
```diff
@@ -44,8 +37,15 @@ function mergeLocation (source, dest) {
   }
 }
 
+let isLong
+
 function build (schema, options) {
   options = options || {}
+
+  if (options.useLong) {
```
mcollina, line 45: This should be enabled by default unfortunately :(, otherwise it'd be semver-major.

mcollina (changes requested):

index.js
gajus: I suggest making it a major change.

mcollina: I'm ok with it.. unless you plan to use this in Fastify v3.

L2jLiga: major update seems to be the one possible imo, even if  we enable it by default it will be breaking changes for users without `long` package installed

mcollina: I mean that this should be enabled by default if long is available, with an option to disable it entirely.

gajus: We can add:

```js
let useLong = options.useLong === undefined ? Boolean(require.resolve('long')) : options.useLong;

if (options.useLong) {
  isLong = require('long').isLong
}

```

if the goal is to make it backwards compatible.

gajus: Never mind. I didn't realize that `require.resolve` throws an error if it cannot resolve. That gets us to where we started.

L2jLiga: `require.resolve` will throw as well, see https://nodejs.org/api/modules.html#modules_require_resolve_request_options

mcollina: Adding a try/catch inside build seems ok to me. Would that fix your problem?

gajus: The annoying part are the warnings that appear in the build log when using the `try..catch` error.

Overall, seems low-value reason to keep compatibility with. The fact that library changes behavior simply based on whether a dependency exists (somewhere in the reachable dependency tree) seems very wonky.

I advocate for major version bump.

jsumners: You're saying this creates an error log?:

```js
try {
  useLong = Boolean(require.resolve('long'))
} catch {
  useLong = false
}
```

README.md
```diff
@@ -524,7 +530,7 @@ const stringify = fastJson({
       type: 'integer'
     }
   }
-})
+}, {useLong: true})
```
Eomm, line 533:
```suggestion
}, { useLong: true })
```

mcollina:
We are starting to prepare for Fastify v4. If you would like this change to be made (or change anything regarding how longs are handled) could you send a PR targeting the `next` branch?

PR #291, opened 2021-03-07
fix: avoid side effect per oneOf and anyOf

### LINKED ISSUE #289: oneOf with $ref mutate the schema
## 🐛 Bug Report

The `oneOf` keyword is replaced with the schema

Causes https://github.com/fastify/fastify/issues/2901

## To Reproduce

```js
const fjs = require('fast-json-stringify')

const reusedSchema = {
  $id: 'example2',
  type: 'object',
  properties: {
    name: {
      oneOf: [{ $ref: 'example1' }]
    }
  }
}

fjs(reusedSchema, {
  schema: {
    example1: {
      $id: 'example1',
      type: 'object',
      properties: {
        name: { type: 'string' }
      }
    }
  }
})

console.log(JSON.stringify(reusedSchema, null, 2))

```

## Expected behavior

`reusedSchema` should not change

## Your Environment

- *node version*: 12
- *fastify version*: >=3.10.0
- fast-json-stringify: 2.4.2

Eomm:
Fixes #289 

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -1191,7 +1191,7 @@ function nested (laterCode, name, key, location, subKey, isArray) {
 
           // see comment on anyOf about derefencing the schema before calling ajv.validate
           code += `
-            ${index === 0 ? 'if' : 'else if'}(ajv.validate(${require('util').inspect(location.schema, { depth: null, maxArrayLength: null })}, obj${accessor}))
+            ${index === 0 ? 'if' : 'else if'}(ajv.validate(${JSON.stringify(location.schema)}, obj${accessor}))
               ${nestedResult.code}
           `
           laterCode = nestedResult.laterCode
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1183,7 +1191,7 @@ function nested (laterCode, name, key, location, subKey, isArray) {
 
           // see comment on anyOf about derefencing the schema before calling ajv.validate
           code += `
-            ${index === 0 ? 'if' : 'else if'}(ajv.validate(${require('util').inspect(location.schema, { depth: null, maxArrayLength: null })}, obj${accessor}))
+            ${index === 0 ? 'if' : 'else if'}(ajv.validate(${JSON.stringify(location.schema)}, obj${accessor}))
```
Eomm, line 1194: I had to change this line because it was producing this output that trigger a syntax error:

```
if(ajv.validate({
  oneOf: [ { type: 'number' }, { type: 'boolean' } ],
  [Symbol(fast-json-stringify.cloned)]: true
}, obj["cs"]))
```

I don't know why this sentence was using `inspect`

I tried `showHidden: false` since I was expecting that symbols would be ignored but nothing:

```

const obj = {
  [Symbol('hello')]: 'world'
}

const a = require('util').inspect(obj, {
  showHidden: false,
  depth: null,
  maxArrayLength: null
})
console.log(a) // { [Symbol(hello)]: 'world' }

```

mcollina (approved):
lgtm

index.js
mcollina: It was just a mistake, we should use JSON.stringify().

mcollina (approved):
lgtm

PR #302, opened 2021-03-13
fix side effect on items $ref

Eomm:
Fixes https://github.com/fastify/fastify/issues/2914
Follow up PR https://github.com/fastify/fastify/pull/2925


mcollina (approved):
lgtm

PR #304, opened 2021-03-22
Moving "long" to dependency from dev dependency

Rafi993:
#### Checklist

#### Description
package 'long' is present in dev dependency. But is used in https://github.com/fastify/fast-json-stringify/blob/master/index.js#L15 this causes error when package using `fast-json-stringify` is installed as dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "benchmark": "^2.1.4",
     "compile-json-stringify": "^0.1.2",
     "is-my-json-valid": "^2.20.0",
-    "long": "^4.0.0",
     "moment": "^2.24.0",
     "pre-commit": "^1.2.2",
     "proxyquire": "^2.1.3",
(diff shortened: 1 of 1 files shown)


L2jLiga:
Could you describe your environment and build/run configuration? `require` statement wrapped in `try/cache` block and should work without any problems

Rafi993:
> Could you describe your environment and build/run configuration? `require` statement wrapped in `try/cache` block and should work without any problems

I was using 'roarr' package that was using this project as dependency (it is in browser).

L2jLiga:
> I was using 'roarr' package that was using this project as dependency (it is in browser).

Then probably it's good option to start from issue in `roarr` ?
Also additional information and/or steps to reproduce would be awesome to have

Rafi993:
> > I was using 'roarr' package that was using this project as dependency (it is in browser).
> 
> Then probably it's good option to start from issue in `roarr` ?
> Also additional information and/or steps to reproduce would be awesome to have

I agree let me start issue there and close this. Thanks for your time :)

PR #307, opened 2021-03-25
Update to ajv version 7.2.3

MSE99:
#### Checklist

#### Primary changes

1. Removed ajv-pack from  devDependencies, it's deprecated and ajv 7 comes with a built-in module
generator.

2. Rewrote `build-schema-validator.js` to use ajv 7 standalone validator compiler, and recompiled json-schema-draft-07
validator.

3. Modified all ajv imports and changed them from `require('ajv')` to `require('ajv').default` , ajv 7 uses typescript and exports
an ES2015 class unlike ajv 6 which exported a plain old constructor function as the exports object .

4. Added a new function `$validateWithAjv` into the stringifier scope, ajv 7 will throw an error if you call `ajv.validate` twice with a schema that has an `$id` property, ajv compiles the schema into a validator function and caches it on the first call, the second call will throw an error because the stringifier always passes the full schema definition (with the `$id` property) to every `ajv.validate` call instead of passing the `$id` property, I'm not sure why ajv 6 didn't throw when this happened but this behavior broke `side-effect.test.js` [second test](https://github.com/fastify/fast-json-stringify/blob/master/test/side-effect.test.js#L39) , the `$validateWithAjv` basically keeps a cache of `$id` values from schemas and will either call `ajv.validate` with the entire schema (if it's the first call) or with the `$id` value on subsequent calls, i admit that this is probably not the best solution and i was hoping other people would give me suggestions on how to better do this.

5. Added `"target": "ES2015"` and `"moduleResolution": "node"` to `tsconfig.json` inside the types test directory, upgrading to ajv 7 broke the test types because it uses data structures from ES2015.

6. Changed error [message](https://github.com/fastify/fast-json-stringify/blob/master/test/allof.test.js#L111) assertion.

7. Compiled stringifier with ajv option `{ strictTypes: false }` in `if-then-else.test.js` test line 313.

8. Compiled stringifier with ajv option `{allowUnionTypes: true}` in two tests inside `typesArray.test.js` test suite.


jsumners (commented):
I think this would be a major and need to coincide with a new major in Fastify.

index.js
```diff
@@ -87,6 +87,28 @@ function build (schema, options) {
     ${$asBoolean.toString()}
     ${$asBooleanNullable.toString()}
 
+    const $validateWithAjv = (function() {
```
mcollina, line 90: Can you add a comment explaining this? What you put in the PR is totally fine.

mcollina (approved):
lgtm

mcollina:
I'll update the `next` branch so this could be landed there.

PR #316, opened 2021-04-09
Fastify issue 2990

jsumners:
This resolves https://github.com/fastify/fastify/issues/2990

I am _guessing_ that an AJV instance isn't passed to every validator as a memory optimization technique. Frankly, I think that's a bit extreme. I think it would be far simpler to rip out the `dependsOnAjv` check and simply:

```js
const dependencies = [new Ajv(options.ajv]
const dependenciesName = ['ajv']

dependenciesName.push(code)
```

But I didn't do that. Instead I added a new conditional check in the `dependsOnAjv` function. While I was at it, I split the regular expression up into its constituent parts. I had a really difficult time seeing the `|` operators in the expression, and it's clear this is just going to continue to grow.

### DIFF
--- a/index.js
+++ b/index.js
@@ -225,7 +225,19 @@ function inferTypeByKeyword (schema) {
 
 function dependsOnAjv (schema) {
   const str = JSON.stringify(schema)
-  return (/"if":{.*"then":{|"(anyOf|oneOf)":\[|"const":/.test(str))
+  switch (true) {
+    case /"if":{.*"then":{/.test(str):
+    case /"(anyOf|oneOf)":\[/.test(str):
+    case /"const"/.test(str):
+    case /"\$ref"/.test(str):
+    {
+      return true
+    }
+
+    default: {
+      return false
+    }
+  }
 }
 
 const stringSerializerMap = {
--- a/test/requiresAjv.test.js
+++ b/test/requiresAjv.test.js
@@ -1,0 +1,48 @@
+'use strict'
+
+const t = require('tap')
+const build = require('..')
+
+t.test('nested ref requires ajv', async t => {
+  const schemaA = {
+    $id: 'urn:schema:a',
+    definitions: {
+      foo: { anyOf: [{ type: 'string' }, { type: 'null' }] }
+    }
+  }
+
+  const schemaB = {
+    $id: 'urn:schema:b',
+    type: 'object',
+    properties: {
+      results: {
+        type: 'object',
+        properties: {
+          items: {
+            type: 'object',
+            properties: {
+              bar: {
+                type: 'array',
+                items: { $ref: 'urn:schema:a#/definitions/foo' }
+              }
+            }
+          }
+        }
+      }
+    }
+  }
+
+  const stringify = build(schemaB, {
+    schema: {
+      [schemaA.$id]: schemaA
+    }
+  })
+  const result = stringify({
+    results: {
+      items: {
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

L2jLiga (approved):
LGTM

jsumners:
According to `npm owner ls`, only @mcollina and @delvedor can publish a new version of this package.

Eomm (commented):
> I am guessing that an AJV instance isn't passed to every validator as a memory optimization technique

It is true

The error trigger here was that the `anyOf` keyword was in the external schema and no one of the keywords to validate (`anyOf, oneOf, const and if`) was in the main schema.

I think this solution works best, another possible approach was to check all the external schemas before, but it would mean to run a lot more `JSON.stringify`)

jsumners:
An empty container is just a function. I really don't think this optimization is providing any real measurable benefit.

jsumners:
Here's a short experiment https://github.com/jsumners/fjs-ajv-experiment

The experiment creates a Fastify server with 2,000 routes. Each route has a very basic response schema that does not require AJV for serialization. The first server uses FJS as designed. The second server skips the optimization and always adds AJV.

My conclusion is that adding AJV every time does not significantly impact memory usage. But I look forward to your evaluation of the methodology and results, @Eomm and @mcollina.

Eomm:
> Here's a short experiment https://github.com/jsumners/fjs-ajv-experiment
> 
> The experiment creates a Fastify server with 2,000 routes. Each route has a very basic response schema that does not require AJV for serialization. The first server uses FJS as designed. The second server skips the optimization and always adds AJV.
> 
> My conclusion is that adding AJV every time does not significantly impact memory usage. But I look forward to your evaluation of the methodology and results, @Eomm and @mcollina.

I had these constant results:

```
Exercising server-a...
Exercising server-b...
server-a stats: {
  rss: 87650304,
  heapTotal: 61538304,
  heapUsed: 34964928,
  external: 1569273,
  arrayBuffers: 26814
}
server-b stats: {
  rss: 89767936,
  heapTotal: 63619072,
  heapUsed: 30390512,
  external: 1569273,
  arrayBuffers: 26814
}

heap delta: 2080768
----------------------------

Exercising server-a...
Exercising server-b...
server-a stats: {
  rss: 87453696,
  heapTotal: 62062592,
  heapUsed: 35038096,
  external: 1569273,
  arrayBuffers: 26814
}
server-b stats: {
  rss: 89350144,
  heapTotal: 63881216,
  heapUsed: 30434408,
  external: 1569273,
  arrayBuffers: 26814
}

heap delta: 1818624

-----------------

Exercising server-a...
Exercising server-b...
server-a stats: {
  rss: 87699456,
  heapTotal: 62062592,
  heapUsed: 34953792,
  external: 1569273,
  arrayBuffers: 26814
}
server-b stats: {
  rss: 89735168,
  heapTotal: 63881216,
  heapUsed: 30488952,
  external: 1569273,
  arrayBuffers: 26814
}

heap delta: 1818624
```

Dell xps13 2016
8GB ram
CPU intel i5

I think a little memory consumption could worth the speed up process since we will avoid this line:

https://github.com/fastify/fast-json-stringify/blob/be7fa47e30432eabc4e14a01c1a08cdedff2db6a/index.js#L227

for every response schema in fastify

mcollina:
I would say.. let's go with it.

Issue #321, opened 2021-04-18
Additional Items is not working


ajotaos:
## 🐛 Bug Report

The additional items modifier is causing the module to fail for tuples. AJV works with the schema but fjs fails with the same schema and data.

## To Reproduce

```js
const fast = require('fast-json-stringify');
const serialize = fast({
      type: 'object',
      properties: {
        code: { type: 'string', const: 'BAD_REQUEST' },
        path: {
          type: 'array',
          items: [
            {
              type: 'string',
              enum: ['headers', 'params', 'querystring', 'body']
            }
          ],
          additionalItems: true
        },
        reason: { type: 'string', minLength: 1 }
      },
      required: ['code', 'path', 'reason'],
      additionalProperties: false
    });

serialize({
  "code": "BAD_REQUEST",
    "path": [
      "body",
      "query"
    ],
    "reason": "should be graphql source"
  });
```

## Expected behavior

Expected the serialization to occur, but the `path` schema fails if `additionalItems` is `true` or `{ type: 'string'}`.

## Your Environment

- *node version*: v15.14.0
- *fastify version*:3.14.2
- *os*: Mac

mcollina:
I have a feeling we never implemented this. 
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

KiraPC:
@mcollina Can this issue be closed by https://github.com/fastify/fast-json-stringify/pull/322?

@ajotaos Could you check now?

PR #322, opened 2021-04-26
added support to additionalItems property in array object

KiraPC:
It is the first time I saw the code of this lib, I am amazed by the idea and its implementation. 

So many compliments!

---
#### Checklist

---

The PR is intended to fix the issue https://github.com/fastify/fast-json-stringify/issues/321

Brief explanation: when the 'items' property of an array object is described as array itself, only the schemas defined are admitted. However, with the property 'additionalItems' defined in the array definition, it should accept more, without validating the type.

### DIFF
--- a/test/array.test.js
+++ b/test/array.test.js
@@ -210,6 +210,69 @@ buildTest({
   foo: [1, 'string', {}, null]
 })
 
+test('array items is a list of schema and additionalItems is true, just the described item is validated', (t) => {
+  t.plan(1)
+
+  const schema = {
+    type: 'object',
+    properties: {
+      foo: {
+        type: 'array',
+        items: [
+          {
+            type: 'string'
+          }
+        ],
+        additionalItems: true
+      }
+    }
+  }
+
+  const stringify = build(schema)
+  const result = stringify({
+    foo: [
+      'foo',
+      'bar',
+      1
+    ]
+  })
+
+  t.equal(result, '{"foo":["foo","bar",1]}')
+})
+
+test('array items is a list of schema and additionalItems is false', (t) => {
+  t.plan(1)
+
+  const schema = {
+    type: 'object',
+    properties: {
+      foo: {
+        type: 'array',
+        items: [
+          {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

test/array.test.js
```diff
@@ -210,6 +210,69 @@ buildTest({
...
+  try {
+    stringify({
+      foo: [
+        'foo',
+        'bar'
+      ]
+    })
+    t.fail()
+  } catch (error) {
+    t.ok(/does not match schema definition./.test(error.message))
```
zekth, line 272: This matches the definition. Should fail if it was `{foo:['foo','bar',1]}`

zekth (changes requested):
I feel the test assertion is wrong or am i missing something?

test/array.test.js
climba03003: Reference: https://json-schema.org/understanding-json-schema/reference/array.html
The `additionalItems` keyword controls whether it’s valid to have additional items in the array beyond what is defined in `items`.

KiraPC: @zekth The case pass because additionItems is 'false', and so the schema expect that the array foo should has just one string item.

climba03003: Missed this line. Thanks for clarify.
> Tuple validation is useful when the array is a collection of items where each has a different schema and the ordinal index of each item is meaningful.

KiraPC: @zekth Do you still need some changes?

zekth: Thanks for the clarification.

zekth (approved):
LGTM.
Thanks for clarifying the test case

test/array.test.js
KiraPC: It's a must. I want to contribute, but I don't want you lose much time to understand what I did.

I'll try to be more clear in PR description next time :)

RafaelGSS (approved):
LGTM.

Issue #274, opened 2020-11-26
enum values are not handled well by the serializer


iMoses:
## 🐛 Bug Report

When using enum in a json schema the serialization results turn string values in an array/object of chars.

## To Reproduce

https://codesandbox.io/s/fastify-2707-forked-47hkd

```js
const fastJson = require("fast-json-stringify");
const stringify = fastJson({
  title: "Example Schema",
  type: "object",
  properties: {
    order: {
      enum: ["asc", "desc"]
    }
  }
});
console.log(
  stringify({
    order: "asc"
  })
); // output: {"order":{"0":"a","1":"s","2":"c"}}
```

## Expected behavior

Should return the value of the enum without any modifications.

```js
console.log(
  stringify({
    order: "asc"
  })
); // should output: {"order":"asc"}
```

## Your Environment

- *node version*: 14
- *fastify version*: >=3.8.0
- *os*: Linux

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

iMoses:
> Would you like to send a Pull Request to address this issue? Remember to add unit tests.

I haven't investigated it enough to understand why this is happening. 
I'll try to clear up time for it, but any input on where to look can be helpful here as i don't have too much free time on my hands :)

iMoses:
@mcollina I found the issue [here](https://github.com/fastify/fast-json-stringify/blob/master/index.js#L169) but i'm not sure how this should be handled.

Seems that `enum` is defined as an `objectKeywords` meaning it will be recognize as an object and that's why the string is breaking down. If moved to the `stringKeywords` array the problem is solved, but then it will only work in case the `enum` is a string. Although I'm not sure it should, because it isn't part of json schema, adding a `type: 'string'` does help as the inferred type check never happens this way.

This also opens another bug/feature that I'm not sure how you want to handle, which is that enum values are not validated at all. Seems to me that if the serializer fails when passing it a string instead of an array then it should also fail when passing it `100` for an enum that takes only `10`.

I think a better way is needed here in order to handle enum, as it is possible to use multiple types and it requires a special kind of validation. WDYT?

mcollina:
See https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values. It seems we should have some logic to detect the value at runtime and adapt accordingly, maybe using some deep equality logic.

Eomm:
This is related to #254 

The string is treated like and object and this can be improved 👍

[...]

PR #254, opened 2020-09-06
fix enum without type

Eomm:
Fixes #253 

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #283, opened 2021-01-08
Support for empty schemas #255

sinclairzx81:
This PR adds support for empty `{}` `any` schemas to resolve issue #255. This update checks for the existence of a `type` property on the schema, and if not found, defaults to a `JSON.stringify()` for serialization. 

Have added a few tests to `any.test.js` and updated `index.d.ts` accordingly. Submitting this PR for review. 

### benchmark
```
cpu  AMD Ryzen 7 3700X 8-Core Processor
ram  32GB 
node v12.18.2

FJS creation x 56,104 ops/sec ±0.26% (92 runs sampled)
CJS creation x 162,430 ops/sec ±0.27% (98 runs sampled)

JSON.stringify array x 3,320 ops/sec ±0.33% (91 runs sampled)
fast-json-stringify array x 8,465 ops/sec ±0.37% (90 runs sampled)

compile-json-stringify array x 7,756 ops/sec ±0.75% (94 runs sampled)
JSON.stringify long string x 10,028 ops/sec ±0.17% (95 runs sampled)

fast-json-stringify long string x 10,098 ops/sec ±0.13% (96 runs sampled)
compile-json-stringify long string x 10,108 ops/sec ±0.10% (99 runs sampled)

JSON.stringify short string x 8,700,507 ops/sec ±0.46% (91 runs sampled)
fast-json-stringify short string x 47,164,570 ops/sec ±0.21% (96 runs sampled)

compile-json-stringify short string x 40,127,602 ops/sec ±0.38% (97 runs sampled)
JSON.stringify obj x 1,827,988 ops/sec ±0.12% (96 runs sampled)

fast-json-stringify obj x 17,418,215 ops/sec ±0.39% (95 runs sampled)
compile-json-stringify obj x 24,095,360 ops/sec ±0.23% (95 runs sampled)
```

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #324, opened 2021-04-28
handle enum property when type is not specified

KiraPC:
#### Checklist

@mcollina @Eomm 

As mentioned in https://github.com/fastify/fast-json-stringify/issues/274, I just revert the changes of https://github.com/fastify/fast-json-stringify/pull/254 and everything works, the test cases introduced in 254 and the case mentioned in the issue (I also added it in the enum.test.js).

Anyway, I don't know if I missed something, or better, why enum was introduced as object keywords.

### DIFF
--- a/index.js
+++ b/index.js
@@ -167,8 +167,7 @@ const objectKeywords = [
   'properties',
   'patternProperties',
   'additionalProperties',
-  'dependencies',
-  'enum'
+  'dependencies'
 ]
 
 const arrayKeywords = [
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Eomm (approved):
The tests are red on the old 2.2.3 version when the issue show itself
here my fast snippet: https://runkit.com/embed/kbejnxagtdie

some other changes fixed that use case too it seems, it would be nice to understand which one

KiraPC:
> The tests are read on the old 2.2.3 version when the issue show itself
> here my fast snippet: https://runkit.com/embed/kbejnxagtdie
> 
> some other changes fixed that use case too it seems, it would be nice to understand which one

Got it.

I'll try to investigate starting from 2.2.3 which other fix solve this issue, if need.

KiraPC:
@Eomm 

The fix to this particular issue was silent introduced by https://github.com/fastify/fast-json-stringify/pull/283.

That was in conflict with the changes made in https://github.com/fastify/fast-json-stringify/pull/254.

So, merging mine, we should have everything works :)

Issue #325, opened 2021-05-03
oneOf with Date results to null


vesse:
## 🐛 Bug Report

`oneOf` with `"format": "date-time"` string results to `null` when given a native `Date` object

## To Reproduce

```js
test('object with field of type string with format or null', (t) => {
  t.plan(2)

  const toStringify = new Date()

  const withoutOneOfSchema = {
    type: 'object',
    properties: {
      prop: {
        type: 'string',
        format: 'date-time'
      }
    }
  }

  const withoutOneOfStringify = build(withoutOneOfSchema)

  t.equal(withoutOneOfStringify({
    prop: toStringify
  }), `{"prop":"${toStringify.toISOString()}"}`)

  const withOneOfSchema = {
    type: 'object',
    properties: {
      prop: {
        oneOf: [{
          type: 'string',
          format: 'date-time'
        }, {
          type: 'null'
        }]
      }
    }
  }

  const withOneOfStringify = build(withOneOfSchema)

  // This fails, is really {"prop":null}

  t.equal(withOneOfStringify({
    prop: toStringify
  }), `{"prop":"${toStringify.toISOString()}"}`)
})
```

## Expected behavior

Test above includes expected behavior too

## Your Environment

- *node version*: 14
- *fastify version*: 3.14.2
- *os*: Mac
- Test above run with `master` branch of this repository

zekth:
Thanks for reporting this, same behavior is observed with `anyOf` . Could you send a PR for this?

vesse:
Wow thanks, that was fast.

ivan-tymoshenko:
Can we reopen this issue. #326 fix will work only with top level date/time fields. And works not efficiently. It serialize datetime value twice.

```js
test('anyOf object with nested field date-time of type string with format or null', (t) => {
  t.plan(1)
  const withOneOfSchema = {
    type: 'object',
    properties: {
      prop: {
        anyOf: [{
          type: 'object',
          properties: {
            nestedProp: {
              type: 'string',
              format: 'date-time'
            }
          }
        }]
      }
    }
  }

  const withOneOfStringify = build(withOneOfSchema)

  const data = {
    prop: { nestedProp: new Date() }
  }

  t.equal(withOneOfStringify(data), JSON.stringify(data))
})
```

PR #326, opened 2021-05-03
fix: ajv validation oneOf anyOf

zekth:
Here is a PR to fix https://github.com/fastify/fast-json-stringify/issues/325

Not sure this is the best implementation so far.


mcollina:
looks good!

mcollina (approved):
lgtm

Issue #330, opened 2021-05-18
Regression "oneOf with Date results to null"


mashpie:
## 💥 Regression Report

#325 "oneOf with Date results to null" seams to persist in 2.7.2 and up

## Last working version

Worked up to version: 2.7.1

Stopped working in version: 2.7.2

## To Reproduce

```js
'use strict'

const fastJson = require('fast-json-stringify')
const S = require('fluent-json-schema')

/**
 * @todo works up to fast-json-stringify@2.7.1
 * but broken in fast-json-stringify@2.7.2
 */ 
const isoDateOrEmptyString = S.anyOf([
  S.string().format('date'),
  S.string().maxLength(0).default('')
])

const exampleSchema = S.object()
  .additionalProperties(false)
  .id('exampleSchema')
  .prop('dateIso', isoDateOrEmptyString)
  .prop('dateEmpty', isoDateOrEmptyString)
  .prop('dateNull', isoDateOrEmptyString)
  .valueOf()

const stringify = fastJson(exampleSchema)

console.log(stringify({
  dateIso: '2020-01-01',
  dateEmpty: '',
  dateNull: null,
}))

/**
 * expected
 * @example {"dateIso":"2020-01-01","dateEmpty":"","dateNull":null}
 */

/**
 * reported
 * @example {"dateIso":null,"dateEmpty":"","dateNull":null}
 */
```

## Expected behavior

the snippet should report:

```sh
{"dateIso":"2020-01-01","dateEmpty":"","dateNull":null}
```

## Your Environment

- *node version*: 14
- *fastify version*: 3.15.1 but reproducable without any fastify
- *os*: Mac, Linux

### Background

I encountered that regression today while upgrading a bunch of services deployed as fastify (v3.7.x - v3.15.1) apps. The JSON returned `null` for all valid iso formatted dates we had while still handling empty strings correctly. Fortunately we were able to monkey patch dependencies by:

```json
  "resolutions": {
    "fast-json-stringify": "2.7.1"
  }
``` 

The same behavior is reproducible without fastify and just plain json-schema. 	Accordingly, the snippet above fails with a `package.json` like so:

```json
{
  "name": "tmp-fast-json-stringify",
  "version": "0.0.0",
  "main": "index.js",
  "license": "MIT",
  "dependencies": {
    "fast-json-stringify": "^2.7.3",
    "fluent-json-schema": "^3.0.0"
  }
}
```

but works with

```json
{
[...]

PR #331, opened 2021-05-18
fix: #330 extra quotes on string test

zekth:
fix: #330 

Extra quotes were added when using the test serializer leading to the null fallback

### DIFF
--- a/test/anyof.test.js
+++ b/test/anyof.test.js
@@ -494,3 +494,26 @@ test('anyOf object with field date of type string with format or null', (t) => {
     prop: toStringify
   }), '{"prop":"2011-01-01"}')
 })
+
+test('anyOf object with invalid field date of type string with format or null', (t) => {
+  t.plan(1)
+  const toStringify = 'foo bar'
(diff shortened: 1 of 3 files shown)


jsumners (approved):
Looks good to me.

test/anyof.test.js
```diff
@@ -471,3 +471,26 @@ test('anyOf object with field of type string with format or null', (t) => {
     prop: toStringify
   }), `{"prop":"${toStringify.toISOString()}"}`)
 })
+
+test('anyOf object with field date of type string with format or null', (t) => {
+  t.plan(1)
+  const toStringify = '2011-01-01'
```
Eomm, line 477: I would add a test with a `foo bar` string and add a disclaimer in the readme that a date format string, will not be validated (because we assume that the user has already parsed the Date object)

Eomm (approved):

mcollina (approved):
lgtm

L2jLiga (approved):
LGTM

PR #332, opened 2021-05-21
fix round internal integer

Eomm:
fix #329 

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #334, opened 2021-05-25
Attempt to fix 333

### LINKED ISSUE #333: Integer properties are always stringified as 0
### Prerequisites

### Fastify version

3.x

### Plugin version

2.7.5

### Node.js version

14.x

### Operating system

macOS

11.3

### Description

Version 2.7.5 introduced a bug where an integer property always ends up being stringified even if the schema doesn't provide a default value for it.

### Steps to Reproduce

```
const stringify = fastJson({
  title: 'Example Schema',
  type: 'object',
  properties: {
    age: {
      type: 'integer'
    }
  }
})

console.log(stringify({}))
```

### Expected Behavior

I'd expect an empty object as string: `{}` but I get `{ age: 0 }` instead.

Version 2.7.4 and prior ones work as expected.

cesarfd:
#### Checklist

Fixes #333 (I think?)

Hi, I tried to fix the problem in #333 by removing the default value introduced by @Eomm  [here](https://github.com/fastify/fast-json-stringify/pull/278/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R250).

I've added a few more tests and all _seems_ to be ok but I'm not sure this is the right approach as it's the first time I've taken a look at this library.

#332 calls `asInteger` instead of `Number` but `asInteger` has a fallback case that will inevitably convert a null value to 0. Removing that line fixes the issue and at the same time the tests in #332 (as a well as a few more added by me) also pass but again, I'm afraid some situations could be affected by this.

### DIFF
--- a/test/date.test.js
+++ b/test/date.test.js
@@ -93,7 +93,7 @@ test('render a date in a string when format is time as kk:mm:ss', (t) => {
   validate(JSON.parse(output))
   t.equal(validate.errors, null)
 
-  t.equal(output, `"${moment(toStringify).format('kk:mm:ss')}"`)
+  t.equal(output, `"${moment(toStringify).format('HH:mm:ss')}"`)
   t.ok(validate(JSON.parse(output)), 'valid schema')
 })
 
--- a/test/date.test.js
+++ b/test/date.test.js
@@ -76,7 +76,7 @@ test('verify padding for rendered date in a string when format is date', (t) => 
   t.ok(validate(JSON.parse(output)), 'valid schema')
 })
 
-test('render a date in a string when format is time as HH:mm:ss', (t) => {
+test('render a date in a string when format is time as kk:mm:ss', (t) => {
   t.plan(3)
 
   const schema = {
@@ -93,7 +93,7 @@ test('render a date in a string when format is time as HH:mm:ss', (t) => {
   validate(JSON.parse(output))
   t.equal(validate.errors, null)
 
-  t.equal(output, `"${moment(toStringify).format('HH:mm:ss')}"`)
+  t.equal(output, `"${moment(toStringify).format('kk:mm:ss')}"`)
   t.ok(validate(JSON.parse(output)), 'valid schema')
 })
 
--- a/test/integer.test.js
+++ b/test/integer.test.js
@@ -6,6 +6,7 @@ const semver = require('semver')
 const validator = require('is-my-json-valid')
 const proxyquire = require('proxyquire')
 const build = proxyquire('..', { long: null })
+const ROUNDING_TYPES = ['ceil', 'floor', 'round']
 
 test('render an integer as JSON', (t) => {
   t.plan(2)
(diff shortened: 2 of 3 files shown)


test/integer.test.js
```diff
@@ -149,3 +153,15 @@ test('should round interger object parameter ', t => {
   t.equal(output, '{"magic":5}')
   t.ok(validate(JSON.parse(output)), 'valid schema')
 })
+
+test('should not stringify a property if it does not exist', t => {
+  t.plan(2)
+
+  const schema = { type: 'object', properties: { magic: { type: 'integer' } } }
+  const validate = validator(schema)
+  const stringify = build(schema)
```
Eomm, line 162: Could you add a test with the `rounding` option?

mcollina (approved):
lgtm

cesarfd:
One of the date.test.js assertions failed (at my local time) because it expected 12:xx.yy but it got a 24:xx:yy string. I changed the moment format to `kk` instead of `HH` and it seems fine.

test/date.test.js
```diff
@@ -76,7 +76,7 @@ test('verify padding for rendered date in a string when format is date', (t) =>
   t.ok(validate(JSON.parse(output)), 'valid schema')
 })
 
-test('render a date in a string when format is time as HH:mm:ss', (t) => {
+test('render a date in a string when format is time as kk:mm:ss', (t) => {
```
Eomm, line 79: Could you revert this change?

You can commit with `git commit -nm 'comment'` to skip the test on commit (and run the tests in the GH CI)

test/integer.test.js
```diff
@@ -40,6 +40,10 @@ test('render a float as an integer', (t) => {
   const cases = [
     { input: Math.PI, output: '3' },
     { input: 5.0, output: '5' },
+    { input: null, output: '0' },
```
Eomm, line 43: Not sure this was expected because the nullable parameter, I think the output should be an empty string

test/integer.test.js
```diff
@@ -149,3 +153,51 @@ test('should round interger object parameter ', t => {
...
+  t.equal(output, '{}')
+  t.ok(validate(JSON.parse(output)), 'valid schema')
+})
+
+test('should not stringify a property if it does not exist (rounding: ceil)', t => {
+  t.plan(2)
+
+  const schema = { type: 'object', properties: { magic: { type: 'integer' } } }
+  const validate = validator(schema)
+  const stringify = build(schema, { rounding: 'ceil' })
```
Eomm, line 174: Just as optimization, you could create a dynamic test that accept it as parameter to avoid the code repetition.

Such as

```

[ceil, floor].forEach(rounding => {
  test('should not stringify a property if it does not exist (rounding: )'+rounding, t => {
    // ...
  }
})

**Review by Eomm — COMMENTED** [review]
Could you add your use case as test?
this would be loud and clear that the issue has been fixed

```const stringify = fastJson({
  title: 'Example Schema',
  type: 'object',
  properties: {
    age: {
      type: 'integer'
    }
  }
})

console.log(stringify({}))
```

mcollina (approved):
lgtm

Issue #338, opened 2021-06-23
Webpack renames $asBoolean and friends, causing generated code to fail with "$asBoolean is not defined"


diwic:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure it has not already been reported

### Fastify version

3.14.1

### Node.js version

14.17

### Operating system

Windows

### Operating system version (i.e. 20.04, 11.3, 10)

10

### Description

I'm trying to package a fastify application with webpack (without externalizing fastify or fast-json-stringify, we need the final code to be all included in a single .js file).
This worked well until I changed from development to production mode in webpack. Now webpack renames `$asBoolean` to a shorter name, so that the generated code no longer compiles.

### Steps to Reproduce

Create these files:

index.js:
```
const fastJson = require('fast-json-stringify')
const stringify = fastJson({
  title: 'Example Schema',
  type: 'object',
  properties: {
    valid: { type: 'boolean' }
  }
}, { debugMode: true });

console.log(stringify.toString());
```

webpack.config.js:
```
module.exports = {
	mode: "production",
	entry: "./index.js",
	target: "node",
	output: {
		filename: "packaged.js"
	}
};
```

package.json:
```
{
  "name": "jsonstringifytest",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "fast-json-stringify": "^2.7.6"
  },
  "devDependencies": {
    "webpack": "^5.40.0",
    "webpack-cli": "^4.7.2"
  }
}
```

Then run:

```
npm i
npx webpack
node dist/packaged.js
```

### Expected Behavior

When running without debugmode, expected working serialization. Instead the packaged program crashes with `$asBoolean is not defined`.

For the debugmode example above, output contains:
```
function k(e){return e?"true":"false"}
/* ... */
json += $asBoolean(obj["valid"])
```

Notice how the function has been renamed, but the call to it has not. The result is that the generated code no longer compiles.

I expected output to contain either:

```
function $asBoolean (bool) {
  return bool && 'true' || 'false' // eslint-disable-line
}
/* ... */
json += $asBoolean(obj["valid"])
```

or possibly

```
function k(e){return e?"true":"false"}
/* ... */
json += k(obj["valid"])
```

mcollina:
[...]

PR #339, opened 2021-06-24
Fix minimizer renaming functions

diwic:
When running the code through a minimizer (e g through Webpack/Terser),
the minimizer renames functions such as "$asBoolean". However, when the
function is called, it's hard coded as the string "$asBoolean", so the
generated code fails with "$asBoolean is not defined".

Make sure that the original $asBoolean function is included, and not
a renamed version of it, by making $asBoolean and friends string constants.

Fixes: https://github.com/fastify/fast-json-stringify/issues/338

#### Checklist

~~No new tests or benchmarks.~~ The one added test causes the time to run the tests to increase from 2 to 7 seconds.

Benchmarks seem to go up or down a few percent, probably within the margin of error.

N/A

### DIFF
--- a/index.js
+++ b/index.js
@@ -72,21 +72,7 @@ function build (schema, options) {
   `
 
   code += `
-    ${$pad2Zeros.toString()}
-    ${$asAny.toString()}
-    ${$asString.toString()}
-    ${$asStringNullable.toString()}
-    ${$asStringSmall.toString()}
-    ${$asDatetime.toString()}
-    ${$asDate.toString()}
-    ${$asTime.toString()}
-    ${$asNumber.toString()}
-    ${$asNumberNullable.toString()}
-    ${$asInteger.toString()}
-    ${$asIntegerNullable.toString()}
-    ${$asNull.toString()}
-    ${$asBoolean.toString()}
-    ${$asBooleanNullable.toString()}
+    ${asFunctions}
 
     var isLong = ${isLong ? isLong.toString() : false}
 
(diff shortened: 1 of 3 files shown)


mcollina:
Could you add a test so that we do not regress?

diwic:
@mcollina I'm a little hesitant to do so given that it does not really fit into the testing infrastructure. It would also require an additional devDependency on webpack and increase the time to run the tests, although I'm not certain how much.
Do you want me to try regardless?

Otherwise a comment explaining why things look the way they do, could be a reasonable compromise between testing time and regression risk?

mcollina:
@jsumners @delvedor wdyt?

I'll prefer to have a testing strategy for this.

index.js
```diff
@@ -928,7 +916,7 @@ function buildObject (location, code, name) {
   if (schema.nullable) {
     code += `
       if(input === null) {
-        return '${$asNull()}';
+        return 'null';
```
jsumners, line 919: Why this change?

index.js
```diff
@@ -967,7 +955,7 @@ function buildArray (location, code, name, key = null) {
   if (schema.nullable) {
     code += `
       if(obj === null) {
-        return '${$asNull()}';
+        return 'null';
```
jsumners, line 958: Why this change?

index.js
```diff
@@ -372,7 +359,7 @@ function $asStringSmall (str) {
       surrogateFound = true
     }
     if (point === 34 || point === 92) {
-      result += str.slice(last, i) + '\\'
+      result += str.slice(last, i) + '\\\\'
```
jsumners, line 362: Why this change?

jsumners:
> @jsumners @delvedor wdyt?
> 
> I'll prefer to have a testing strategy for this.

If there isn't an accompanying test, I don't know what this is fixing.

index.js
diwic: The $asNull function is now inside a string constant, so it is not available to be called in this context.

index.js
diwic: Same here - the $asNull function is now inside a string constant, so it is not available to be called in this context.

index.js
diwic: $asStringSmall is now inside an es6 template string. Inside those strings, slashes need to be escaped.

diwic:
@mcollina  @jsumners  All right, I've added a test case now (in a separate commit).

index.js
jsumners: Is it not the case that an alternative function can be provided for `$asNull` and that such an alternative could return a value other than `'null'`?

diwic: Not to my knowledge? If so I don't see how such a function would be provided.

diwic: I could change it to `return $asNull()` I guess, but then there are other places in the code which just return `'null'` anyway, so it would still make much difference in terms of consistency.

jsumners:
I'm not super familiar with how all of this works. So I'll defer to @Eomm.

index.js
Eomm: there is still one `$asNull` call. I would remove it and remove the function aswell

mcollina (approved):
lgtm

diwic:
Thanks everybody for the quick reviews! Much appreciated.

PR #351, opened 2021-08-19
fix: resolve fragment $refs in nested objects (#350)

### LINKED ISSUE #350: Error when resolving nested schema refs by id ({"$ref": "#foo"})
### Prerequisites

### Fastify version

3.20.2

### Plugin version

_No response_

### Node.js version

v14.16.0

### Operating system

Linux

Ubuntu 20.04.2

### Description

`refFinder` fails to find a `$ref` by `$id` ([docs](https://www.fastify.io/docs/latest/Validation-and-Serialization/#json-schema-support)), when the schema is nested, e.g.:

```json
{
  "title": "Example Schema",
  "type": "object",
  "properties": {
    "nested": {
      "type": "object",
      "properties": {
        "firstName": {
          "$ref": "#foo"
        },
        "lastName": {
          "$ref": "#foo"
        }
      }
    }
  },
  "definitions": {
    "foo": {
      "$id": "#foo",
      "type": "string"
    }
  }
}
```

There are actually two errors ([source](https://github.com/fastify/fast-json-stringify/blob/master/index.js#L615-L627)):

- if there is no `externalSchema`, you'll get `TypeError: Cannot convert undefined or null to object` when executing `for (var key of Object.keys(externalSchema)) {`
- if `externalSchema` is specified, `refFinder` will fail to dereference `$ref` and a property will end up resolved to `undefined` which will trigger `TypeError: Cannot read property 'type' of undefined` on the later stages in `buildCode`

Looks like `refFinder` should look up for definitions starting from `root` instead of `schema` in the same way as it done for [processing](https://github.com/fastify/fast-json-stringify/blob/master/index.js#L635-L644) of `#/definitions/foo` refs. 

### Steps to Reproduce

```js
const fastJson = require('fast-json-stringify');
const stringify = fastJson({
  title: 'Example Schema',
  type: 'object',
  properties: {
    firstName: { $ref: '#foo' },
    lastName: { $ref: '#foo' },
    nested: {
      type: 'object',
      properties: {
        firstName: { $ref: '#foo' },
        lastName: { $ref: '#foo' },
      },
    },
  },
  definitions: {
    foo: {
      $id: '#foo',
      type: 'string',
    },
  },
});

console.log(
  stringify({
    firstName: 'Matteo',
    lastName: 'Collina',
    nested: {
      firstName: 'Matteo',
      lastName: 'Collina',
    },
  }),
);
```

### Expected Behavior

The nested `$ref` should be resolved. Here is an example of schema validation: https://www.jsonschemavalidator.net/s/sMCUkqfP

SleepWalker:
#### Checklist

Closes #350

### DIFF
--- a/index.js
+++ b/index.js
@@ -588,7 +588,7 @@ function refFinder (ref, location) {
   // Split file from walk
   ref = ref.split('#')
 
-  // If external file
+  // If external or subschema
   if (ref[0]) {
     schema = externalSchema[ref[0]]
     root = externalSchema[ref[0]]
@@ -614,8 +614,8 @@ function refFinder (ref, location) {
     const walk = ref[1].split('/')
     if (walk.length === 1) {
       const targetId = `#${ref[1]}`
-      let dereferenced = idFinder(schema, targetId)
-      if (dereferenced === undefined && !ref[0]) {
+      let dereferenced = idFinder(root, targetId)
+      if (dereferenced === undefined && !ref[0] && externalSchema) {
         // eslint-disable-next-line
         for (var key of Object.keys(externalSchema)) {
           dereferenced = idFinder(externalSchema[key], targetId)
@@ -624,6 +624,10 @@ function refFinder (ref, location) {
             break
           }
         }
+      }
+
+      if (dereferenced === undefined) {
+        throw new Error(`Cannot find reference ${targetId}`)
       }
 
       return {
--- a/test/ref.test.js
+++ b/test/ref.test.js
@@ -398,6 +398,63 @@ test('ref internal - plain name fragment', (t) => {
   t.equal(output, '{"obj":{"str":"test"}}')
 })
 
+test('ref internal - plain name fragment in nested objects', (t) => {
+  t.plan(2)
+
+  const schema = {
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -626,6 +626,10 @@ function refFinder (ref, location) {
         }
       }
 
+      if (dereferenced === undefined) {
```
Eomm, line 629: We should test this case

index.js
```diff
@@ -614,8 +614,8 @@ function refFinder (ref, location) {
     const walk = ref[1].split('/')
     if (walk.length === 1) {
       const targetId = `#${ref[1]}`
-      let dereferenced = idFinder(schema, targetId)
-      if (dereferenced === undefined && !ref[0]) {
+      let dereferenced = idFinder(root, targetId)
```
Eomm, line 617: I understand this change, but I wonder if there is a more performant way to accomplish it such as:
it there is not the id in the `schema`, loop-up in the parent

Right now this change could impact performance and add new behaviours (same id in same schema)

Eomm (changes requested):

PR #357, opened 2021-09-14
perf: speed up date formatting by 555x

kevcodez:
The current version of date formatting is constructing a bunch of new objects using Intl and is comparetively slow. The simply `toISOString` and `splice` show a local speed up of up to 300x in terms of formatting and should yield the same results.

The date tests use momentjs formatting as comparison, which seems to respect time zones, thus, I also respected the time zones.

Before change
> fast-json-stringify date format x 2,900 ops/sec ±1.19% (82 runs sampled)

After change
> fast-json-stringify date format x 1,608,371 ops/sec ±1.30% (89 runs sampled)

**x 555 faster**

#### Checklist

### DIFF
--- a/bench.js
+++ b/bench.js
@@ -50,22 +50,38 @@ const arraySchemaCJS = {
   items: schemaCJS
 }
 
+const dateFormatSchema = {
+  description: 'Date of birth',
+  type: 'string',
+  format: 'date'
+}
+
+const dateFormatSchemaCJS = {
+  description: 'Date of birth',
+  type: 'string',
+  format: 'date'
+}
+
 const obj = {
   firstName: 'Matteo',
   lastName: 'Collina',
   age: 32
 }
 
+const date = new Date()
+
 const multiArray = []
 
 const CJS = require('compile-json-stringify')
 const CJSStringify = CJS(schemaCJS)
 const CJSStringifyArray = CJS(arraySchemaCJS)
+const CJSStringifyDate = CJS(dateFormatSchemaCJS)
 const CJSStringifyString = CJS({ type: 'string' })
 
 const FJS = require('.')
 const stringify = FJS(schema)
 const stringifyArray = FJS(arraySchema)
+const stringifyDate = FJS(dateFormatSchema)
 const stringifyString = FJS({ type: 'string' })
 let str = ''
 
@@ -136,6 +152,18 @@ suite.add('fast-json-stringify obj', function () {
 
 suite.add('compile-json-stringify obj', function () {
   CJSStringify(obj)
+})
+
+suite.add('JSON stringify date', function () {
+  JSON.stringify(date)
+})
+
+suite.add('fast-json-stringify date format', function () {
+  stringifyDate(date)
+})
+
+suite.add('compile-json-stringify date format', function () {
+  CJSStringifyDate(date)
 })
 
 suite.on('cycle', cycle)
--- a/index.js
+++ b/index.js
@@ -285,10 +285,7 @@ function $asDatetime (date, skipQuotes) {
 function $asDate (date, skipQuotes) {
   const quotes = skipQuotes === true ? '' : '"'
   if (date instanceof Date) {
-    const year = new Intl.DateTimeFormat('en', { year: 'numeric' }).format(date)
-    const month = new Intl.DateTimeFormat('en', { month: '2-digit' }).format(date)
-    const day = new Intl.DateTimeFormat('en', { day: '2-digit' }).format(date)
-    return quotes + year + '-' + month + '-' + day + quotes
+    return quotes + new Date(date.getTime() - (date.getTimezoneOffset() * 60000 )).toISOString().slice(0, 10) + quotes
   } else if (date && typeof date.format === 'function') {
     return quotes + date.format('YYYY-MM-DD') + quotes
   } else {


zekth:
Indeed good catch:
https://github.com/fastify/fast-json-stringify/blob/e0d93367260a18b6bf8e62352eeee5f01543b3ce/index.js#L313-L316

Could you include benchmarks?

kevcodez:
@zekth Added a benchmark for date formatting

mcollina (approved):
lgtm

bench.js
```diff
@@ -138,6 +154,14 @@ suite.add('compile-json-stringify obj', function () {
   CJSStringify(obj)
 })
 
+suite.add('compile-json-stringify date format', function () {
```
zekth, line 157: Please keep consistency in the order of benchmarks and add `JSON.stringify`. Order should be
- JSON.stringify
- FJS
- CJS

zekth (changes requested):

bench.js
kevcodez: Added JSON.stringify test and changed order

index.js
```diff
@@ -285,10 +285,7 @@ function $asDatetime (date, skipQuotes) {
 function $asDate (date, skipQuotes) {
   const quotes = skipQuotes === true ? '' : '"'
   if (date instanceof Date) {
-    const year = new Intl.DateTimeFormat('en', { year: 'numeric' }).format(date)
-    const month = new Intl.DateTimeFormat('en', { month: '2-digit' }).format(date)
-    const day = new Intl.DateTimeFormat('en', { day: '2-digit' }).format(date)
-    return quotes + year + '-' + month + '-' + day + quotes
+    return quotes + new Date(date.getTime() - (date.getTimezoneOffset() * 60000 )).toISOString().slice(0, 10) + quotes
```
Eomm, line 288: Would you mind to add a test for this?

```
console.log(s(new Date('August 19, 1975 00:15:30 GMT+01:00'))) // "1975-08-19"
```

I tried a bit and it is aligned with the old code output

kevcodez: Added a test for that specific case

Eomm: Missing git push?

kevcodez: 😅

test/date.test.js
```diff
@@ -58,6 +58,24 @@ test('render a date in a string when format is date as YYYY-MM-DD', (t) => {
...
+    type: 'string',
+    format: 'date'
+  }
+  const toStringify = new Date('August 19, 1975 00:15:30 GMT+01:00')
+
+  const validate = validator(schema)
+  const stringify = build(schema)
+  const output = stringify(toStringify)
+
+  t.equal(output, '"1975-08-19"')
```
Eomm, line 75: CI is failing

```
2021-09-14T16:00:05.9090566Z             --- expected
2021-09-14T16:00:05.9090925Z             +++ actual
2021-09-14T16:00:05.9091357Z             @@ -1,1 +1,1 @@
2021-09-14T16:00:05.9091956Z             -"1975-08-19"
2021-09-14T16:00:05.9092379Z             +"1975-08-18"
```

My TZ is GMT+2, it should be related

kevcodez: The result from the formatting is equal to what `moment(toStringify).format('YYYY-MM-DD')` would output.

Instantiating a new Date automatically converts it to UTC.

`getTimezoneOffset` gets the offset compared to the machine running the code, thus yielding different results.

The current implementation takes that into account to always output the UTC date and not put out any system-specific dates (otherwise the other tests fail locally) - this is the same behavior as moment.js.

Tbh, I'd keep this test out and keep the implementation as is.

WDYT?

zekth: > CI is failing
> 
> ```
> 2021-09-14T16:00:05.9090566Z             --- expected
> 2021-09-14T16:00:05.9090925Z             +++ actual
> 2021-09-14T16:00:05.9091357Z             @@ -1,1 +1,1 @@
> 2021-09-14T16:00:05.9091956Z             -"1975-08-19"
> 2021-09-14T16:00:05.9092379Z             +"1975-08-18"
> ```
> 
> My TZ is GMT+2, it should be related

That's what was wondering. We need to enforce UTC ain't we?

kevcodez: @zekth We are enforcing UTC in the implementation. From your comment

> console.log(s(new Date('August 19, 1975 00:15:30 GMT+01:00'))) // "1975-08-19"

I thought, you want to the result of the formatting from that date to be 1975-08-19, which wouldn't be UTC and would be different to the current `Intl` implementation.

Do you want to remove that new test or do you want the assertion to be 1975-08-18?

zekth: Assertion shouldn't be changed. Atm i have no proper solution that comes to my mind.

kevcodez: Again, this would break the current implementation. It does not work with JS dates like that. That assertion would also break with the current `Intl` implementation. The test would fail without my changes given that assertion. I don't think we are trying to introduce a breaking change here.

mcollina:
I'm a bit lost on the next steps here.

kevcodez:
@mcollina Not sure how this conversation got lost.

@Eomm requested an assertion that is not possible to implement or make it system-independent (independent of local timezone). The requested assertion would **also break without my changes**. So in my opinion, the requested assertion is not correct and cannot be solved in this PR.

When instantiating a new date, the date is automatically converted to UTC and there is no way to access that timezone information afterwards. So it is not possible to make this assertion work like that.

zekth (approved):
@kevcodez sorry i misinterpreted the test case added. Indeed it's LGTM to me.

good catch on this

climba03003:
I have look into the [document](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat#using_datetimeformat)

`Intl.DateTimeFormat` will respect the system timezone offset. Is it not the same case in NodeJS?

kevcodez:
@climba03003 The new implementation respects the system timezone offset using `getTimezoneOffset`. However, when instantiating a new Date, Node automatically converts it to UTC and you cannot get the initial timezone. You can convert the UTC date to your system timezone (which my code does, just like Intl), but you cannot access the initial timezone from the instantiated date.

climba03003:
> 
> 
> @climba03003 The new implementation respects the system timezone offset using `getTimezoneOffset`. However, when instantiating a new Date, Node automatically converts it to UTC and you cannot get the initial timezone. You can convert the UTC date to your system timezone (which my change does, just like Intl), but you cannot access the initial timezone from the instantiated date.

@kevcodez Not sure why it is not working in your case. But I have try to reproduce it. It can get the offset.
image

climba03003 (approved):
LGTM. I didn't notice the handling is updated. Currently, it should respect the system timezone when it is `Date` object.

kevcodez:
@climba03003 Intl and `getTimezoneOffset` both use your system time. Just like the proposed changes in this PR. This behavior is unchanged.

When running this on a system with UTC time, i.e. the GitHub action, this will have `0` as timezone offset and will output the day `18`. So, we **cannot** have a test like this

```js
const date = new Date('August 19, 1975 00:15:30 GMT+01:00')
const output = stringify(date)
  t.equal(output, '1975-08-19')
```

This will only work on systems that are in a GMT+ or above timezone, but will fail on any system with UTC or before timezone.

The current tests compare the output with what `momentjs.format` would do, which does respect the system timezone. Meaning, both Intl and the proposed new implementation respect the system timezone.

mcollina (approved):
lgtm

Eomm (approved):

kevcodez:
@mcollina it would be super dope if you could release the changes when you find some time 🙏

mcollina:
I'll get to it once we finish restoring the array perf.

Issue #365, opened 2021-11-01
Recursive references in object types result in maximum call stack size exceeded


simonsinclair:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.22.1

### Node.js version

14.18.1

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

11.5.1

### Description

My Fastify server returns JSON that looks like the following Food object:

```json
{
  "name": "Hare, meat only, stewed, weighed with bone",
  "category": {
    "name": "Hare",
    "parent": {
      "name": "Game",
      "parent": {
        "name": "Meat and meat products",
        "parent": null
      }
    }
  }
}
```

Each Food has a category, which can have a parent category, which can also have a parent category, i.e. `food.category.parent.parent`.

I have tried to model this using the JSON Schema outlined in the reproduction steps below, however it results in my server crashing at startup with the following error:

> Failed building the serialization schema for GET: /foods/:id, due to error Maximum call stack size exceeded

#### 1. JSON Schema
```json
{
  "$schema": "http://json-schema.org/draft-07/schema",
  "$id": "https://www.example.com/schemas/food.json",
  "type": "object",
  "properties": {
    "name": {
      "type": "string"
    },
    "category": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "parent": {
          "$ref": "#/$defs/parent_category"
        }
      }
    }
  },
  "$defs": {
    "parent_category": {
      "type": ["object", "null"],
      "properties": {
        "name": {
          "type": "string"
        },
        "parent": {
          "$ref": "#/$defs/parent_category"
        }
      }
    }
  }
}
```

#### 2. Minimal server
```ts
import 'make-promises-safe';
import { exit } from 'process';
import fastify from 'fastify';
import Food from './schemas/food.json';

const server = fastify({ logger: true });

server.get(
  '/foods/hare',
  {
    schema: {
      response: {
        200: Food,
      },
    },
  },
  async () => ({
    name: 'Hare, meat only, stewed, weighed with bone',
    category: {
      name: 'Hare',
      parent: {
        name: 'Game',
        parent: {
          name: 'Meat and meat products',
          parent: null,
        },
      },
    },
  }),
);

const listen = async () => {
  try {
    await server.listen(8000, '0.0.0.0');
  } catch (error) {
    server.log.error(error);
    exit(1);
  }
[...]

PR #366, opened 2021-11-09
Add support for recursive references in object types

simonsinclair:
Resolves https://github.com/fastify/fast-json-stringify/issues/365.

#### Notes
- This probably needs more tests.

#### Checklist

### DIFF
--- a/test/recursion.test.js
+++ b/test/recursion.test.js
@@ -135,3 +135,46 @@ test('use proper serialize function', t => {
   })
   t.equal(value, '{"people":{"name":"Elizabeth","children":[{"name":"Charles","children":[{"name":"William","children":[{"name":"George"},{"name":"Charlotte"}]},{"name":"Harry"}]}]},"directory":{"name":"directory 1","subDirectories":[{"name":"directory 1.1","subDirectories":[]},{"name":"directory 1.2","subDirectories":[{"name":"directory 1.2.1","subDirectories":[]},{"name":"directory 1.2.2","subDirectories":[]}]}]}}')
 })
+
+test('can stringify recursive references in object types (issue #365)', t => {
+  t.plan(1)
+
+  const schema = {
+    type: 'object',
+    properties: {
+      category: {
+        type: 'object',
+        properties: {
+          parent: {
+            $ref: '#/$defs/parent_category'
+          }
+        }
+      }
+    },
+    $defs: {
+      parent_category: {
+        type: 'object',
+        properties: {
+          parent: {
+            $ref: '#/$defs/parent_category'
+          }
+        }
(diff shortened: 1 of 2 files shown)


test/recursion.test.js
```diff
@@ -135,3 +135,46 @@ test('use proper serialize function', t => {
...
+            $ref: '#/$defs/parent_category'
+          }
+        }
+      }
+    }
+  }
+
+  const stringify = build(schema)
+  const data = {
+    category: {
```
climba03003, line 168: Can you add an unit test for
```
const parant = {
  parant: parant
}
const data = {
  category: {
    parant: parant
  }
}
```

I thiink most of the recursive reference support library will face this unless object reference looping.

test/recursion.test.js
```diff
@@ -135,3 +135,46 @@ test('use proper serialize function', t => {
...
+      category: {
+        type: 'object',
+        properties: {
+          parent: {
+            $ref: '#/$defs/parent_category'
+          }
+        }
+      }
+    },
+    $defs: {
```
Eomm, line 154: please use JSON schema draft 07 syntax

mcollina (approved):
lgtm

test/recursion.test.js
simonsinclair: > Can you add an unit test for
> 
> ```
> const parant = {
>   parant: parant
> }
> const data = {
>   category: {
>     parant: parant
>   }
> }
> ```

Is this a valid use-case? Attempting to reference "`parant`" on L2 before it's defined will result in an error.

L2jLiga (approved):
LGTM

mcollina:
is this ready to land?

L2jLiga:
> is this ready to land?

From my end I see that tests in CI passed and we're step closer to give recursion support, in future we can think about how to have only one map, but for now it's fine for me

simonsinclair:
I had a test in mind for multiple types, which is currently failing, i.e.:

```js
parentCategory: {
  type: ['object', 'null'], // 👈
  properties: {
    parent: {
      $ref: '#/definitions/parentCategory'
    }
  }
}
```

I'm looking into this now, but happy build on this in another PR if you prefer.

L2jLiga:
I have no strong opinion on that, before we had crash, now we can work with this data, even if we have some bugs or mistakes they shouldn't affect functionality that was working and affect only recursion support

test/recursion.test.js
climba03003: My mistake on the example, it should works this time.
```
const parant = {}
const data = {
  category: {  parant }
}
parant.parant = parant
```

PR #371, opened 2021-11-26
Func added to avoid 500 call stack size error 

Vibhaa5:
There was a  error that it exceeded the call stack size .. added a func to avoid that error.

 suggestions are welcome.


climba03003:
Do you see the first line of `schema-validator.js`? `// This file is autogenerated by build-schema-validator.js, do not edit`

I do not know what is the file `clear` used for. Probably an invalid PR in my mind.

mcollina:
If you'd like to solve this.. maybe start with a unit test?

PR #383, opened 2022-01-18
chore: unnecessary stringify calls

Eomm:
Not sure about the `randomUUID()` performance.
I would use something more basic as `Math.random()` or `crypto.randomBytes`.

Need to try it

### DIFF
--- a/index.js
+++ b/index.js
@@ -849,7 +849,7 @@ function addIfThenElse (location, name) {
   ajvInstance.addSchema(i, schemaKey)
 
   code += `
-    valid = ajv.validate(${JSON.stringify(schemaKey)}, obj)
+    valid = ajv.validate("${schemaKey}", obj)
     if (valid) {
   `
   if (merged.if && merged.then) {
@@ -1203,7 +1203,7 @@ function nested (laterCode, name, key, location, subKey, isArray) {
           ajvInstance.addSchema(location.schema, schemaKey)
 
           code += `
-            ${index === 0 ? 'if' : 'else if'}(ajv.validate(${JSON.stringify(schemaKey)}, ${testValue}))
+            ${index === 0 ? 'if' : 'else if'}(ajv.validate("${schemaKey}", ${testValue}))
               ${nestedResult.code}
           `
           laterCode = nestedResult.laterCode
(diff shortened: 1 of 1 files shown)


zekth (approved):
LGTM

`Math.random` can have collision AFAIK

RafaelGSS (commented):
Could you share the benchmark?

Eomm:
Running the script on Node.js 16

```
randomInt x 121,083,649 ops/sec ±0.35% (595 runs sampled)
randomInt + now x 9,471,249 ops/sec ±0.07% (596 runs sampled)
uuid x 24,935,836 ops/sec ±0.74% (589 runs sampled)
math.random x 113,739,051 ops/sec ±1.56% (566 runs sampled)
math.random + now x 6,165,940 ops/sec ±0.44% (591 runs sampled)
```

> Math.random can have collision AFAIK

True, but I'm not so lucky and I never hit this case.
For this, I tried the `+ now()` concatenation but it is the worse performance.

```js
const Benchmark = require('benchmark')
Benchmark.options.minSamples = 500

const crypto = require('crypto')

Benchmark.Suite()
  .add('randomInt', function () { crypto.randomInt(10e6) })
  .add('randomInt + now', function () { return `${Date.now()}${crypto.randomInt(10e6)}` })
  .add('uuid', function () { crypto.randomUUID() })
  .add('math.random', function () { return Math.random() * 10e6 })
  .add('math.random + now', function () { return `${Date.now()}${Math.random() * 10e6}` })
  .on('cycle', function (event) { console.log(String(event.target)) })
  .on('complete', function () {})
  .run()
```

RafaelGSS:
> Running the script on Node.js 16
> 
> ```
> randomInt x 121,083,649 ops/sec ±0.35% (595 runs sampled)
> randomInt + now x 9,471,249 ops/sec ±0.07% (596 runs sampled)
> uuid x 24,935,836 ops/sec ±0.74% (589 runs sampled)
> math.random x 113,739,051 ops/sec ±1.56% (566 runs sampled)
> math.random + now x 6,165,940 ops/sec ±0.44% (591 runs sampled)
> ```

This benchmark in comparison to the master has degraded the performance? I mean, I could not compare with the result in the README or my local benchmarks.

mcollina:
IMHO randomUUID() is fast enough to not cause problems in practice.

Issue #406, opened 2022-04-19
`$ref` doesn't understand the `nullable` option


samchon:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.1.0

### Node.js version

16.6.0

### Operating system

Windows

### Operating system version (i.e. 20.04, 11.3, 10)

Windows 11

### Description

To represent such hierarchical data, I've utilized the `$ref` and encountered the null exception.

If's possible to avoiding such null exception by defining duplicated external definition with nullable `false` and `true`, however, it seems too inefficient. I hope that the `$ref` to support the `nullable` option exactly. In the same reason, I hope the `anyOf` or `someOf` defined object to support the `nullable` option too.

```typescript
import faster from "fast-json-stringify";

// THE RECURSIVE STRUCTURE
interface IDepartment
{
    name: string;
    parent: IDepartment | null;
}

// CREATE STRINGIFY FUCTION
const stringify = faster({ $ref: "external/#o6" },
{
    external: {
        o6: {
            type: "object",
            properties: {
                name: {
                    type: "string",
                    nullable: false
                },
                parent: {
                    $ref: "external#/o6",
                    nullable: true
                }
            }
        }
    }
});

// TypeError: Cannot read property 'name' of null
stringify({
    name: "bottom",
    parent: {
        name: "middle",
        parent: {
            name: "top",
            parent: null
        }
    }
});
``````

### Steps to Reproduce

It's possible to reproduce the bug with such similar code.

```typescript
import faster from "fast-json-stringify";

// VERY SIMPLE STRUCTURE
interface IProduct
{
    name: string;
}

// CONSTRUCT STRINGIFY FUCTION
const stringify = faster
(
    {
        $ref: "external#/o4",
        nullable: true
    },
    {
        external: {
            o4: {
                type: "object",
                properties: {
                    name: {
                        type: "string",
                        nullable: false
                    }
                }
            }
        }
    }
);

// TypeError: Cannot read property 'name' of null
stringify(null);
```

### Expected Behavior

Hope upper codes to work without exception.

climba03003:
Your example is not correct for using external schema.
There is no option called `external`.

If you need something that is `null`, you should not pass it in.
```js
const stringify = faster(
  { $ref: "external#/o6" },
  {
    schema: {
      external: {
        o6: {
          type: "object",
          properties: {
            name: {
              type: "string",
              nullable: false,
[...]

PR #408, opened 2022-04-21
chore: add nullable for $ref

xtx1130:
ref: https://github.com/fastify/fast-json-stringify/issues/406

ps: I'm not sure if these test cases is enough for this pr.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -600,12 +600,14 @@ function refFinder (ref, location) {
   const externalSchema = location.externalSchema
   let root = location.root
   let schema = location.schema
+  const nullable = location.nullable || schema.nullable || false
 
   if (externalSchema && externalSchema[ref]) {
     return {
       schema: externalSchema[ref],
       root: externalSchema[ref],
-      externalSchema: externalSchema
+      externalSchema: externalSchema,
+      nullable: nullable
     }
   }
 
@@ -625,7 +627,8 @@ function refFinder (ref, location) {
       return refFinder(schema.$ref, {
         schema: schema,
         root: root,
-        externalSchema: externalSchema
+        externalSchema: externalSchema,
+        nullable: nullable
       })
     }
   }
@@ -653,7 +656,8 @@ function refFinder (ref, location) {
       return {
         schema: dereferenced,
         root: root,
-        externalSchema: externalSchema
+        externalSchema: externalSchema,
+        nullable: nullable
       }
     } else {
       // eslint-disable-next-line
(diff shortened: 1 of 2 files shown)


climba03003 (changes requested):
It is not the real fix to the issue. The fix should be the property of resolved `$ref` can be overrided. It should not limited to nullable only.

xtx1130:
Why need to support all the properties? I think this is just a json pointer string and `nullable` just represent `$ref`(JSON pointer) can be `null`. for [rfc6901](https://datatracker.ietf.org/doc/html/rfc6901):
> Note that JSON pointers can contain the NUL (Unicode U+0000)
   character.

I'm not sure if I understand this correctly

climba03003:
> Why need to support all the properties? I think this is just a json pointer string and `nullable` just represent `$ref`(JSON pointer) can be `null`. for [rfc6901](https://datatracker.ietf.org/doc/html/rfc6901):
> 
> > Note that JSON pointers can contain the NUL (Unicode U+0000)
> > character.
> 
> I'm not sure if I understand this correctly

Because it is how the spec tells.

From [`JSON Schema 4-7`](https://json-schema.org/understanding-json-schema/structuring.html#ref), 
> When an object contains a $ref property, the object is considered a reference, not a schema. Therefore, **any other properties** you put in that object will not be treated as JSON Schema keywords and will be **ignored by the validator**.

From [`JSON Schema 2019-09`](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.7.7.1.1)
> Applications MAY make decisions on which of multiple annotation values to use based on the schema location that contributed the value. This is intended to allow flexible usage.

The question is, `support` all keywords to be overrided or `do not support`. For easy usage, I would support all.

xtx1130:
thanks for you detailed explanation. And I find the [JSON Reference document](https://datatracker.ietf.org/doc/html/draft-pbryan-zyp-json-ref-03), in Syntax chapter:
> The "$ref" string value contains a URI [[RFC3986](https://datatracker.ietf.org/doc/html/rfc3986)], which identifies
   the location of the JSON value being referenced.  It is an error
   condition if the string value does not conform to URI syntax rules.
   Any members other than "$ref" in a JSON Reference object SHALL be
   ignored.

So, I think it should not be fixed at all.

mcollina:
I don't think a fix is needed.

climba03003:
I would close this PR and the issue together.
If anyone would like the support of `JSON Schema 2019-09` flexible usage mentioned in https://github.com/fastify/fast-json-stringify/pull/408#issuecomment-1106078468
It deserves a new issue for the discussion first.

PR #414, opened 2022-05-13
Draft: fix(toJSON): support json primitives with toJSON

bmenant:
_This failed to bring extensive support for `toJSON()` method on par with native `JSON.stringify()`._

------

This fixes #412 but I’m not quite sure if it’s the right thing to do. Notably, I have very little clue about performance penalties those changes could have introduced for more common use cases.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [NA] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

#### Benchmarks

Hardware: Intel<sup>®</sup> Core™ i7-8665U @ 1.9GHz × 8T/4C, 16GB, SSD.
Node.js: v16.14.0

##### fast-json-stringify@3.2.0

```
FJS creation x 5,967 ops/sec ±1.48% (88 runs sampled)
CJS creation x 128,198 ops/sec ±0.85% (90 runs sampled)
AJV Serialize creation x 52,421,613 ops/sec ±2.90% (86 runs sampled)
JSON.stringify array x 2,900 ops/sec ±0.42% (93 runs sampled)
fast-json-stringify array default x 5,803 ops/sec ±1.28% (87 runs sampled)
fast-json-stringify array json-stringify x 5,699 ops/sec ±1.47% (88 runs sampled)
compile-json-stringify array x 6,314 ops/sec ±1.63% (90 runs sampled)
AJV Serialize array x 6,796 ops/sec ±0.47% (94 runs sampled)
JSON.stringify large array x 153 ops/sec ±0.44% (85 runs sampled)
fast-json-stringify large array default x 85.13 ops/sec ±2.65% (62 runs sampled)
fast-json-stringify large array json-stringify x 149 ops/sec ±1.27% (83 runs sampled)
compile-json-stringify large array x 283 ops/sec ±2.13% (82 runs sampled)
AJV Serialize large array x 80.90 ops/sec ±0.67% (69 runs sampled)
JSON.stringify long string x 9,499 ops/sec ±0.66% (91 runs sampled)
fast-json-stringify long string x 9,469 ops/sec ±0.79% (91 runs sampled)
compile-json-stringify long string x 9,426 ops/sec ±0.88% (89 runs sampled)
[...]

PR #424, opened 2022-05-25
fix: check all types for toJSON function

ivan-tymoshenko:
#### Checklist

toJSON function can return any type as a result. Not it's working only if object has toJSON method that returns another object.

There is a problem with the Date type. It has toJSON function by default, and it's not clear how to handle this case.

### DIFF
--- a/index.js
+++ b/index.js
@@ -58,19 +58,22 @@ let ajvInstance = null
 
 class Serializer {
   constructor (options = {}) {
-    switch (options.rounding) {
-      case 'floor':
-        this.parseInteger = Math.floor
-        break
-      case 'ceil':
-        this.parseInteger = Math.ceil
-        break
-      case 'round':
-        this.parseInteger = Math.round
-        break
-      default:
-        this.parseInteger = Math.trunc
-        break
+    if (options.rounding !== undefined) {
+      switch (options.rounding) {
+        case 'floor':
+          this.parseInteger = Math.floor
+          break
+        case 'ceil':
+          this.parseInteger = Math.ceil
+          break
+        case 'round':
+          this.parseInteger = Math.round
+          break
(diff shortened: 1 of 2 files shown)


mcollina:
Does this causes any perf regression?

ivan-tymoshenko:
Снимок экрана 2022-05-25 в 19 16 05

ivan-tymoshenko:
It's an interesting example https://github.com/fastify/fast-json-stringify/blob/03f8714fc2412c3e46cb7bbb947e316cad0681bc/test/date.test.js#L129.

`moment` is an object. It's not a `Date` type . It has toJSON method. Should we call it?

mcollina:
The test is correct.

mcollina:
Is this still in the works? What's meeting?

ivan-tymoshenko:
If someone would like to help with this issue, I would appreciate it. Working on another problem right now.

bmenant:
@ivan-tymoshenko Did you see https://github.com/fastify/fast-json-stringify/pull/414 ?
I tried to tackle this as well, started with pretty much the same changes you did so far, and went even further way, before eventually giving up:

> Turns out this is far tougher than expected. Notably, the logic laid out toward the nested() function requires too many conditions and exceptions, preventing easy and safe (and fast?) behavior addition such as this one.

IMHO, a thorough refactoring is needed in and around the `nested()` function before. But, to be honest, it’s not easy task to grasp the whole logic of the code builder. It has been written to output optimized code from the ground up for a reason. I personally lack the skills and the time to tackle this.

After all, AFAIK none of the _fast_ json stringify lib I tried out there supports such toJSON feature.

Anyway, I’d be glad to discuss the topic and give a hand if you deem so useful.

PR #438, opened 2022-05-28
fix: generated function names collision

ivan-tymoshenko:
#### Checklist

There is possible to have a function names collision. I suggest not overcomplicating things and using only generated function names.

### DIFF
--- a/index.js
+++ b/index.js
@@ -297,11 +297,11 @@ function build (schema, options) {
   switch (schema.type) {
     case 'object':
       main = '$main'
-      code = buildObject(location, code, main)
+      code = buildObject(location, code, main, main)
       break
     case 'array':
       main = '$main'
-      code = buildArray(location, code, main)
+      code = buildArray(location, code, main, main)
       schema = location.schema
       break
     case 'string':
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
The reason we use real names for functions is to ease debugging. Maybe just add an integer at the end of name?

ivan-tymoshenko:
It will not solve the problem. I can create a field with the same name as another field name + counter.

ivan-tymoshenko:
I don't know if it's really useful. If you mean useful in the development process, it's pretty clear what this is a function, because you see inlined arguments. If you talking about an error call stack, I would say we should always throw an explainable error and never use a runtime generated call stack to find the reason for the error.

But maybe I don't see some useful cases.

mcollina:
I developed this module by printing out the generated source code, identify which function was causing a problem, and then change the generator. Anonymous functions would prevent this use case, or make it significantly harder.

You can achieve the same by adding a comment in the function itself.

ivan-tymoshenko:
@mcollina What do you think about comment object properties and their types?
```js
    // {"b":{"type":"string"},"t":{"type":"object"}}
    function anonymous2 (input) {}
```

mcollina:
What I need is the "property chain" that lead to that function., what in the original code was `name + key + subkey`. This always indicated the path to a given object.

ivan-tymoshenko:
I will think how to do it clearly. I don't want to pass additional function argument through bunch of core functions to use it in comments in the generated code.

ivan-tymoshenko:
@mcollina I rewrote it. It adds a comment with function name. You can merge it, but I still think that it's overcomplicate the code.

mcollina (approved):
lgtm

PR #439, opened 2022-05-28
fix: large array serialization

ivan-tymoshenko:
Depends on #437 

This PR fixed two bugs:
1. Add support for nullable large arrays.
2. Fix large array string serialization.
If we pass the `asString` function
https://github.com/fastify/fast-json-stringify/blob/4bec21e6ee6951879958247f0299b11a25490d19/index.js#L159
to this map as `mapFnName `
https://github.com/fastify/fast-json-stringify/blob/4bec21e6ee6951879958247f0299b11a25490d19/index.js#L1043
it will receive the array index as the second argument and will skip quotes

### DIFF
--- a/index.js
+++ b/index.js
@@ -409,9 +409,13 @@ const stringSerializerMap = {
   time: 'serializer.asTime.bind(serializer)'
 }
 
-function getStringSerializer (format) {
-  return stringSerializerMap[format] ||
-  'serializer.asString.bind(serializer)'
+function getStringSerializer (format, nullable) {
+  switch (format) {
+    case 'date-time': return nullable ? 'serializer.asDatetimeNullable.bind(serializer)' : 'serializer.asDatetime.bind(serializer)'
+    case 'date': return nullable ? 'serializer.asDateNullable.bind(serializer)' : 'serializer.asDate.bind(serializer)'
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1040,7 +1052,7 @@ function buildArray (location, code, name, key = null) {
   switch (largeArrayMechanism) {
     case 'default':
       code += `
-      return \`[\${obj.map(${result.mapFnName}).join(',')}]\``
+      return \`[\${obj.map((value) => ${result.mapFnName}(value)).join(',')}]\``
```
mcollina, line 1055: let's remove this .map() while we are at it and use the following pattern:

```
cons res = new Array(obj.length)
for (let i = 0 ; i < obj.length; i++) {
...
}
```

it's significantly faster than .map()

mcollina (approved):
lgtm

mcollina:
Does this has an impact on benchmarks?

ivan-tymoshenko:
almost twice faster

Fdawgs:
> almost twice faster

Are you able to post them please?

ivan-tymoshenko:
Снимок экрана 2022-05-29 в 13 06 04

PR #442, opened 2022-05-30
feat!: drop moment lib date format

### LINKED ISSUE #428: Replace `moment` with `luxon`
https://github.com/fastify/fast-json-stringify/blob/03f8714fc2412c3e46cb7bbb947e316cad0681bc/package.json#L42

`moment` is considered to be in "maintenance mode" https://momentjs.com/docs/#/-project-status/
> We now generally consider Moment to be a legacy project in maintenance mode. It is not dead, but it is indeed done.

Until `Temporal` is available, `moment` should be switched out for `luxon` https://moment.github.io/luxon/#/

ivan-tymoshenko:
❗ BREAKING CHANGE

Motivation: https://github.com/fastify/fast-json-stringify/pull/441#discussion_r884803343

Close #428

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "compile-json-stringify": "^0.1.2",
     "inquirer": "^8.2.4",
     "is-my-json-valid": "^2.20.0",
-    "moment": "^2.29.3",
+    "luxon": "^2.4.0",
     "pre-commit": "^1.2.2",
     "proxyquire": "^2.1.3",
     "semver": "^7.1.0",
--- a/test/date.test.js
+++ b/test/date.test.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const test = require('tap').test
-const moment = require('moment')
+const { DateTime } = require('luxon')
 const validator = require('is-my-json-valid')
 const build = require('..')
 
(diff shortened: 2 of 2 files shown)


mcollina (approved):
I'm ok with the change. I'll likely bump the major at the end of this week. If there are other significant changes, let us know.

ivan-tymoshenko:
@mcollina I will try to push everything that I have, but for now, I think there is a lot of work to do. I won't make it by the end of this week. I don't say that we shouldn't release it, just want to let you know it, maybe it would be better to wait another with a major release.

mcollina:
I have an hard date for the fastify v4 release, that's all.

package.json
```diff
@@ -39,7 +39,7 @@
     "compile-json-stringify": "^0.1.2",
     "inquirer": "^8.2.4",
     "is-my-json-valid": "^2.20.0",
-    "moment": "^2.24.0",
+    "moment": "^2.29.3",
```
jsumners, line 42: Should this be removed?

jsumners (approved):

package.json
mcollina: I think so

ivan-tymoshenko: We still use it for tests. I moved it to devDep, but I guess we can replace it with `luxon` there. https://github.com/fastify/fast-json-stringify/issues/428

PR #454, opened 2022-06-05
fix: use ajv for schema ref resolving

ivan-tymoshenko:
❗BREAKING CHANGES

This PR uses ajv addSchema and getSchema methods for resolving json schema references. For this, we need to represent schema and external schemas in terms of json schema draft 7. 

I think it's a good opportunity to fix some issues.

First of all when we pass external schemas as an object.
```js
const externalSchemas = {
 a: { type: 'number' },
 b: { type: 'string' },
}
```

What do keys of the object mean in terms of json schema? Are they base URLs for these schemas? If so how should what should happen if we pass the key and $id?

```js
const externalSchemas = {
 a: { $id: 'id1' },
 b: { $id: 'id2' },
}
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,7 +13,6 @@ const { randomUUID } = require('crypto')
 const validate = require('./schema-validator')
 
 let largeArraySize = 2e4
-let stringSimilarity = null
 let largeArrayMechanism = 'default'
 const validLargeArrayMechanisms = [
   'default',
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

mcollina:
I'm landing this and trying to assemble a release.

mcollina:
Unfortunately I will need to revert this. It breaks quite a few fastify tests that would need to be updated... and we are running out of time.

Note: technically this is correct and there are other things that would need to updated in Fastify regarding encapsulation to make this work properly.

mcollina:
Feel free to send a fresh PR!

PR #456, opened 2022-06-08
refactor: return context function if schema exists

ivan-tymoshenko:
Depends on #455

Return context function if it already exists instead of creating a new function that calls the original one.


mcollina (approved):
lgtm

PR #460, opened 2022-06-10
feat: standalone mode

### LINKED ISSUE #459: breaking `fastify` standalone error serializer build
### Prerequisites

- [X] I have searched existing issues to ensure the regression has not already been reported

### Last working version

3.0.1

### Stopped working in version

4.0.0

### Node.js version

*

### Operating system

Linux

*

### 💥 Regression Report

I think it related to the refactor of https://github.com/fastify/fast-json-stringify/pull/423

Debug mode no longer work like `README.md` describe.

https://github.com/fastify/fastify/issues/3995

### Steps to Reproduce

```
const debugCompiled = fastJson({
  title: 'default string',
  type: 'object',
  properties: {
    firstName: {
      type: 'string'
    }
  }
}, { debugMode: true })

console.log(debugCompiled) // it is an array of functions that can create your `stringify` function
console.log(debugCompiled.toString()) // print a "ready to read" string function, you can save it to a file

const rawString = debugCompiled.toString()
const stringify = fastJson.restore(rawString) // use the generated string to get back the `stringify` function
console.log(stringify({ firstName: 'Foo', surname: 'bar' })) // '{"firstName":"Foo"}'
```

### Expected Behavior

_No response_

climba03003:
1. move `debugMode` in `mode` option.
2. provide `standalone` mode which people can create the code that can directly require and run
3. fixes #459 

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -663,13 +663,37 @@ const debugCompiled = fastJson({
       type: 'string'
     }
   }
-}, { debugMode: true })
+}, { mode: 'debug' })
 
-console.log(debugCompiled) // it is an array of functions that can create your `stringify` function
-console.log(debugCompiled.toString()) // print a "ready to read" string function, you can save it to a file
+console.log(debugCompiled) // it is a object contain code, ajv instance
+const rawString = debugCompiled.code // it is the generated code
+console.log(rawString) 
 
-const rawString = debugCompiled.toString()
-const stringify = fastJson.restore(rawString) // use the generated string to get back the `stringify` function
+const stringify = fastJson.restore(debugCompiled) // use the generated string to get back the `stringify` function
+console.log(stringify({ firstName: 'Foo', surname: 'bar' })) // '{"firstName":"Foo"}'
+```
+
+<a name="standalone"></a>
+### Standalone Mode
+
+The standalone mode is used to compile the code that can be directly run by `node`
+itself. You need to install `fast-json-stringify`, `ajv`, `fast-uri` and `ajv-formats`
+in order to let the standalone code works.
+
+```js
+const fs = require('fs')
+const code = fastJson({
+  title: 'default string',
+  type: 'object',
+  properties: {
+    firstName: {
+      type: 'string'
+    }
+  }
+}, { mode: 'standalone' })
+
+fs.writeFileSync('stringify.js', code)
+const stringify = require('stringify.js')
 console.log(stringify({ firstName: 'Foo', surname: 'bar' })) // '{"firstName":"Foo"}'
 ```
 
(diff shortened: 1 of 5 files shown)


ivan-tymoshenko (commented):
As I understand, we need this when we want to generate stringify function for schemas that we know before runtime.
1. IMHO, it's better to create util that will use fjs to create such modules. I don't really like that one function can return a function or code of a module depending on the params.

2. Can't we create these stringify functions in the runtime? We need to do it only once. I don't think it will affect performance significantly and, more importantly, we won't have to make sure we recompile the module after we fix something in fjs.

If I missed something, please tell me.

climba03003:
> Can't we create these stringify functions in the runtime?

We always can do it. 

> I don't think it will affect performance significantly

I think it will affect cold start time.

mcollina (approved):
lgtm

ivan-tymoshenko:
@climba03003 @mcollina I was writing a long description of why it's not the best solution. Partially my fault, I should have requested changes, probably. The question is what can we do with it now?

This solution brings a dependency on internal code structure/names. It will block a lot of optimizations that will not support function serialization.

The first thing that will not work that I found is ajv schema validation because we use uuids when inserting a schema and then we will regenerate them.

https://github.com/fastify/fast-json-stringify/blob/310325b49d793072d1f0bead6c9b5e0bf6217359/index.js#L621

ivan-tymoshenko:
These are very serious changes, that would take a lot of effort to maintain. I guess we can fix the Fastify issue easier right now. Can we do something with it and then think deeply about what to do with this idea?

climba03003:
I think we focus on fixing the `fastify` error first.
Since `fastify` do not use `$ref` in that code. We are safe from it currently.

Of course, we can provide a truely `fast-json-stringify` dependent free if `standalone`.
But, it require a bit longer time to work on.

ivan-tymoshenko:
@climba03003 can you just rebuild error schema serializer using the last fjs version that did it correctly?

ivan-tymoshenko:
> I think we focus on fixing the fastify error first.
I agree this is a reasonable approach, but creating a new API is serious. We can't just deprecate it if we find a better solution. We will have to support it.

climba03003:
> We can't just deprecate it if we find a better solution. We will have to support it.

Yes, it is. Can we add a experimental flag to indicate the user do not use it in code at this moment.
Since, we are the package maintainer for `fast-json-stringify` and `fastify`. 
I think we are safe to use it first and we know the limitation of it.

mcollina:
The original refactoring should not have shipped as they broke an important use case for Fastify. I'm happy to remove this flag in a semver-major change (if it's the only change) and update Fastify accordingly.

PR #462, opened 2022-06-13
fix: use ajv for schema ref resolving

ivan-tymoshenko:
Reopen #454
Fix #452 #417 #350

The idea behind this solution is to use ajv's addSchema and getSchema methods to resolve json schema references. To do this, as the serializer is built, it keeps track of and updates a reference to the current location. The reference consists of schema id + json pointer. When it needs to validate subschema (if/else, oneof, anyof), it uses the reference.

There are two tests that use an incorrect format of json schema id/ref. They will not work after this fix.
1. It's an external schema that uses a schema key (base URI) that starts with #. As it is considered a schema key if it can start with #. # should be used only as a local anchor inside the schema.
https://github.com/fastify/fast-json-stringify/blob/310325b49d793072d1f0bead6c9b5e0bf6217359/test/ref.test.js#L981

2. It uses a reference combined with the other two references.
https://github.com/fastify/fast-json-stringify/blob/310325b49d793072d1f0bead6c9b5e0bf6217359/test/ref.test.js#L870

And as we have a reference for the current location in the schema we can use it as a comment inside anonymous functions instead of building a custom version of json pointer (locationPath).

PR #464, opened 2022-06-16
fix: ref with same id in properties

climba03003:
I don't think it is a good fix, but it just works.
Fixes https://github.com/fastify/fastify/issues/4028

This PR is a fix before #462 lands. Since #462 changed a lot in the internal ref resolve and may introduce breaking change.
I see the needs of a temporary fix.

@ivan-tymoshenko do you have better idea on this?

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -52,7 +52,7 @@ function mergeLocation (source, dest) {
 const arrayItemsReferenceSerializersMap = new Map()
 const objectReferenceSerializersMap = new Map()
 const schemaReferenceMap = new Map()
-const dedupRefsMap = new Map()
+const dedupRefsSet = new Set()
 
 let ajvInstance = null
 let contextFunctions = null
@@ -61,6 +61,7 @@ function build (schema, options) {
   arrayItemsReferenceSerializersMap.clear()
   objectReferenceSerializersMap.clear()
   schemaReferenceMap.clear()
+  dedupRefsSet.clear()
 
   contextFunctions = []
   options = options || {}
@@ -151,6 +152,7 @@ function build (schema, options) {
   arrayItemsReferenceSerializersMap.clear()
   objectReferenceSerializersMap.clear()
   schemaReferenceMap.clear()
+  dedupRefsSet.clear()
 
   return stringifyFunc
 }
@@ -235,7 +237,7 @@ function addPatternProperties (location) {
     let ppLocation = mergeLocation(location, { schema: pp[regex] })
     if (pp[regex].$ref) {
       ppLocation = refFinder(pp[regex].$ref, location)
-      pp[regex] = dedupIdsRefs(ppLocation.schema)
+      pp[regex] = ppLocation.schema
     }
 
     try {
@@ -254,7 +256,6 @@ function addPatternProperties (location) {
       }
     `
   })
-  dedupRefsMap.clear()
 
   if (schema.additionalProperties) {
     code += additionalProperty(location)
@@ -329,7 +330,7 @@ function refFinder (ref, location) {
 
   if (externalSchema && externalSchema[ref]) {
     return {
-      schema: externalSchema[ref],
+      schema: dedupIdsRefs(ref, externalSchema[ref]),
       root: externalSchema[ref],
       externalSchema
     }
@@ -453,7 +454,7 @@ function buildCode (location, locationPath) {
     const ref = schema.properties[key].$ref
     if (ref) {
       propertyLocation = refFinder(ref, location)
-      schema.properties[key] = dedupIdsRefs(key, ref, propertyLocation.schema)
+      schema.properties[key] = propertyLocation.schema
     }
 
     // Using obj['key'] !== undefined instead of obj.hasOwnProperty(prop) for perf reasons,
@@ -488,7 +489,6 @@ function buildCode (location, locationPath) {
       }
     `
   })
-  dedupRefsMap.clear()
 
   for (const requiredProperty of required) {
     if (schema.properties && schema.properties[requiredProperty] !== undefined) continue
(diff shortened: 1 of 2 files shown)


ivan-tymoshenko:
@climba03003 I think it's better to do it in the fast-json-stringify-compiler.
https://github.com/fastify/fast-json-stringify-compiler/pull/14


We do the same thing in the ajv-compiler.
https://github.com/fastify/ajv-compiler/blob/10879404784f69aa8c5488f77622b52532a28799/index.js#L80

climba03003:
> I think it's better to do it in the fast-json-stringify-compiler.

But the problem still exist in here if you use it directly.

ivan-tymoshenko:
I would say that it shouldn't work. fastify.getSchema creates the wrong schema, so I think I may sense to fix it in the Fastify (fast-json-stringify-compiler).

climba03003:
> I would say that it shouldn't work. fastify.getSchema creates the wrong schema, so I think I may sense to fix it in the Fastify (fast-json-stringify-compiler).

You can see my test, it is nothing related to `fastify.getSchema`. 
People can use the same schema in both `ajv` and `fjs` directly.
It is strange for people should not add `$id` in their external schemas since it is valid json schema.

ivan-tymoshenko:
I mean, it's not a problem of fjs that we pass target schema duplicate in the external schemas. If we continue this logic inside fjs and make schema $id not strict, we will have to also check for duplicates inside external schemas and we don't need that.

climba03003:
The bug actually cause by `fjs` directly replace the schema with schema that contain `$id`.

https://github.com/fastify/fast-json-stringify/blob/12fa1e90a27a27abfcce5be3adac55de208795d3/index.js#L896

And later on in `anyOf` or `oneOf`. It blindly add the schema that may contain duplicate `$id` cause by the line above.

https://github.com/fastify/fast-json-stringify/blob/12fa1e90a27a27abfcce5be3adac55de208795d3/index.js#L951-L971

Which is definitely something wrong within `fjs`.

climba03003:
> This happens because fjs modifies external schemas. If you want to fix this bug as a separate PR, you need to clone external schemas. You don't need to delete any duplications.

It does not work without the `ajv` changes.

ivan-tymoshenko:
Oh, I got it. Yeah, this is one of the reasons why I decided to use ajv. In #462, we pass json pointer to the original schema instead of coping oneOf/anyOf schema. I would say that this is a slightly different problem than https://github.com/fastify/fast-json-stringify-compiler/pull/14.

Your solution fixes two things oneOf/anyOf case and top-level schema duplicate case. IMHO top-level schema duplicate case should be fixed in the fast-json-stringify-compiler. If we merge #462 oneOf/anyOf will be fixed.

Does it work for you?

If you want I can explain in more detail.

climba03003:
> Does it work for you?

Yes, #462 works. 
but I am afraid it is a breaking change and cannot be able to lands on `fastify@4`.

So, this separate PR is needed for `fastify@4`.

ivan-tymoshenko:
Why do you think #462 is a breaking change?

climba03003:
> Why do you think #462 is a breaking change?

Even if the test-case in here are all passed. People in the outside world may already rely on the wrong behavior.
It is more safe to bump a major in this heavy internal changes.

ivan-tymoshenko:
There is said that Fastify supports the  json schema draft 7 standard. I suppose if something is not working it's a bug. Each fix changes something in behavior.

ivan-tymoshenko:
Your solution fixes it only partially. Only top level $ids.

```js
test('ref with same id in properties', (t) => {
  t.plan(2)

  const externalSchema = {
    ObjectId: {
      $id: 'ObjectId',
      type: 'string',
      definitions: {
        p: { $id: 'p' }
      }
    },
    File: {
      $id: 'File',
      type: 'object',
      properties: {
        _id: { $ref: 'ObjectId' },
        name: { type: 'string' },
        owner: { $ref: 'ObjectId' }
      }
    }
  }

  t.test('anyOf', (t) => {
    t.plan(1)

    const schema = {
      $id: 'Article',
      type: 'object',
      properties: {
        _id: { $ref: 'ObjectId' },
        image: {
          anyOf: [
            { $ref: 'File' },
            { type: 'null' }
          ]
        }
      }
    }

    const stringify = build(schema, { schema: externalSchema })
    const output = stringify({ _id: 'foo', image: { _id: 'bar', name: 'hello', owner: 'baz' } })

    t.equal(output, '{"_id":"foo","image":{"_id":"bar","name":"hello","owner":"baz"}}')
  })

  t.test('oneOf', (t) => {
    t.plan(1)

    const schema = {
      $id: 'Article',
      type: 'object',
      properties: {
        _id: { $ref: 'ObjectId' },
        image: {
          oneOf: [
            { $ref: 'File' },
            { type: 'null' }
          ]
        }
      }
    }

    const stringify = build(schema, { schema: externalSchema })
    const output = stringify({ _id: 'foo', image: { _id: 'bar', name: 'hello', owner: 'baz' } })

    t.equal(output, '{"_id":"foo","image":{"_id":"bar","name":"hello","owner":"baz"}}')
  })
})
```

climba03003:
> There is said that Fastify supports the json schema draft 7 standard. I suppose if something is not working it's a bug. Each fix changes something in behavior.

I am not the one releasing this module and it can be land as patch or minor decided who landing it. But, if it's turns out breaking too many people. I think it might revert again. 

I do see landing 462 benefit a lot.

> Your solution fixes it only partially. Only top level $ids.

It will be complicated in fixing nested thing. I think I will stay in this step.

ivan-tymoshenko:
And it should remove the duplicate schema only if schemas are equal. Otherwise, it should throw an error.

climba03003:
> And it should remove the duplicate schema only if schemas are equal. Otherwise, it should throw an error.

This resolution do not mutate the original schema that added to `FJS`.
The current version never be able to resolve the nested schema with same `$id`.

So, when ever we delete the `$id` in `refFinder`. That `$id` schema should be the unique one and safe to delete.

climba03003:
> This shouldn't compile correctly.

Although it is the wrong behavior. But, it is possible currently.
This PR is not fixing this behavior.

```js
const stringify = FJS(
  {
    $id: "https://example.com/schemas/base",
    type: "object",
    properties: {
      p1: { $id: "p1", type: "string" },
      p2: { $ref: "p1" },
    },
  },
  {
    schema: {
      p1: {
        $id: "p1",
        type: "number",
      },
    },
  }
);
console.log(stringify({ p1: "", p2: "" }));
```

ivan-tymoshenko:
> But, it is possible currently.

No, it's not. Why did you change the test?
This test works with your fix and doesn't work without it.

```js
test('anyOf', (t) => {
  t.plan(1)

  const externalSchema = {
    ObjectId: {
      $id: 'ObjectId',
      type: 'string',
    }
  }

  const schema = {
    type: 'object',
    properties: {
      image: {
        anyOf: [
          { $ref: 'ObjectId' },
          {
            $id: 'ObjectId',
            type: 'number',
          }
        ]
      }
    }
  }

  build(schema, { schema: externalSchema })
  t.pass()
})
```

climba03003:
> No, it's not. Why did you change the test?

This fix should handle both case now.

mcollina:
Closing in favor of https://github.com/fastify/fast-json-stringify/pull/462.

PR #472, opened 2022-06-25
fix lgm alert: Useless assignment to local variable

Uzlopak:
Fixes this LGTM alert
https://lgtm.com/projects/g/fastify/fast-json-stringify/?mode=tree&ruleFocus=7900088

### DIFF
--- a/index.js
+++ b/index.js
@@ -309,9 +309,11 @@ function additionalProperty (location) {
 
     return code
   }
-  const apLocation = ap.$ref
-    ? resolveRef(location, ap.$ref)
-    : mergeLocation(location, 'additionalProperties')
+
+  let apLocation = mergeLocation(location, 'additionalProperties')
+  if (apLocation.schema.$ref) {
+    apLocation = resolveRef(apLocation, apLocation.schema.$ref)
+  }
 
   const valueCode = buildValue(apLocation, 'obj[keys[i]]')
 
--- a/index.js
+++ b/index.js
@@ -297,7 +297,7 @@ function addPatternProperties (location) {
 }
 
 function additionalProperty (location) {
-  let ap = location.schema.additionalProperties
+  const ap = location.schema.additionalProperties
   let code = ''
   if (ap === true) {
     code += `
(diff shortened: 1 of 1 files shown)


Eomm (approved):

index.js
```diff
@@ -309,11 +309,9 @@ function additionalProperty (location) {
...
     return code
   }
-  let apLocation = mergeLocation(location, 'additionalProperties')
-  if (ap.$ref) {
-    apLocation = resolveRef(location, ap.$ref)
-    ap = apLocation.schema
-  }
+  const apLocation = ap.$ref
+    ? resolveRef(location, ap.$ref)
+    : mergeLocation(location, 'additionalProperties')
```
ivan-tymoshenko, line 314: There is a bug here. We should always call `mergeLocation` here and pass `apLocation` to the `resolveRef` instead of `location`.

ivan-tymoshenko (changes requested):

index.js
Uzlopak: I dont quite understand your instructions:
I just saw that alert and wanted to fix it. If there is a bug, I would actually need a test to know that I am actually fixing something. Currently all the tests pass.

Probably it is the best to close this PR and open a bug issue.

index.js
```diff
@@ -309,11 +309,9 @@ function additionalProperty (location) {
 
     return code
   }
-  let apLocation = mergeLocation(location, 'additionalProperties')
-  if (ap.$ref) {
-    apLocation = resolveRef(location, ap.$ref)
-    ap = apLocation.schema
-  }
+  const apLocation = ap.$ref
```
ivan-tymoshenko, line 312:
```suggestion
  let apLocation = mergeLocation(location, 'additionalProperties')
  if (apLocation.schema.$ref) {
    apLocation = resolveRef(apLocation, apLocation.schema.$ref)
  }
```

Try this

Uzlopak: done as suggested

ivan-tymoshenko (approved):

index.js
ivan-tymoshenko: It is not really the bug. Since resolveRef use only `schemaId` property of the `location` object, it will not fail, because `location` and `apLocation` have the same `schemaId`. It is just a proper way to do this. There is no test needed.

Eomm (approved):

PR #473, opened 2022-06-25
fix benchmark-cmp-branch, remove chalk and inquirer

Uzlopak:
benchmark-cmp-branch is broken, as inquirer 9.0.0 is now an ESM-Module, like latest chalk.

Instead of setting a lower version of inquirer and adding it into the blacklist of dependabot it like chalk I simply replaced inquirer with the less heavy cli-select and replaced chalk with basic ansi-codes for the color handling. I mean... after all it is only used for the benchmark.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,5 +13,3 @@ on:
 jobs:
   test:
     uses: fastify/workflows/.github/workflows/plugins-ci.yml@v3
-    with:
-      auto-merge-exclude: 'chalk'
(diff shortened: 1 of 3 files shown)


RafaelGSS (approved):

Fdawgs (approved):
Good job dude

Uzlopak:
Thanks. :)

I realized that I forgot to remove chalk from ci.yml. Removed it there now too.

mcollina (approved):
lgtm

PR #475, opened 2022-06-26
perf: replace deepmerge npm-package with a more performant implementation

Uzlopak:
This PR is replacing the very slow deepmerge package with a special implementation derived from the original code. 
Dont get me wrong: I already proposed to TehShrike two PRs to fix the worst performance bottlenecks two months ago. But no reaction. 

See:

https://github.com/TehShrike/deepmerge/pull/246
https://github.com/TehShrike/deepmerge/pull/247

I actually started it by rewriting the deepmerge code to build a function based on the options, like it is done in this project for JSON-stringificatin. But then I realized, that we need for this project just one simple implementation. So I aborted the rewrite and just implemented the only necessary one.

Here the benchmark. merge is this merge, and deepmerge is the package.

```
merge existing simple keys in target at the roots x 1,006,553 ops/sec ±1.21% (88 runs sampled)
deepmerge existing simple keys in target at the roots x 193,263 ops/sec ±0.79% (89 runs sampled)
merge existing simple keys in target at the roots x 7,392,616 ops/sec ±1.56% (89 runs sampled)
deepmerge existing simple keys in target at the roots x 1,591,627 ops/sec ±1.14% (86 runs sampled)
merge nested objects into target x 5,219,415 ops/sec ±1.09% (87 runs sampled)
deepmerge nested objects into target x 27,331 ops/sec ±1.23% (58 runs sampled)
```

```javascript
var Benchmark = require('benchmark');
var deepmerge = require('deepmerge')
var merge = require('./merge')
var suite = new Benchmark.Suite;

var srcSimple = { key1: 'changed', key2: 'value2' }
var targetSimple = { key1: 'value1', key3: 'value3' }


var srcNested = {
  key1: {
    subkey1: 'subvalue1',
    subkey2: 'subvalue2',
  },
}
var targetNested = {
  key1: 'value1',
  key2: 'value2',
}

var ArraySrc = [{ ...srcSimple }, { ...srcSimple }, { ...srcSimple }, { ...srcSimple }, { ...srcSimple }]
var ArrayTarget = [{ ...targetSimple }, { ...targetSimple }, { ...targetSimple }, { ...targetSimple }, { ...targetSimple }]


// add tests
suite.add('merge existing simple keys in target at the roots', function () {
  var res = merge(ArrayTarget, ArraySrc)
})
// add tests
suite.add('deepmerge existing simple keys in target at the roots', function () {
  var res = deepmerge(ArrayTarget, ArraySrc)
})

suite.add('merge existing simple keys in target at the roots', function () {
  var res = merge(targetSimple, srcSimple)
})
// add tests
suite.add('deepmerge existing simple keys in target at the roots', function () {
  var res = deepmerge(targetSimple, srcSimple)
})
suite.add('merge nested objects into target', function () {
  var res = merge(targetNested, srcNested)
})
suite.add('deepmerge nested objects into target', function () {
  var res = deepmerge(targetNested, srcNested)
})
// add listeners
suite.on('cycle', function (event) {
  console.log(String(event.target));
})
// run async
suite.run({ async: true });
```

### DIFF
--- a/merge.js
+++ b/merge.js
@@ -6,6 +6,10 @@ function isMergeableObject (value) {
   return typeof value === 'object' && value !== null && !(value instanceof RegExp) && !(value instanceof Date)
 }
 
+function map (entry) {
+  return isMergeableObject(entry) ? merge(Array.isArray(entry) ? [] : {}, entry) : entry
+}
+
 function mergeObject (target, source) {
   const result = {}
   if (isMergeableObject(target)) {
@@ -13,10 +17,10 @@ function mergeObject (target, source) {
     let i, il
     for (i = 0, il = keys.length; i < il; ++i) {
       const key = keys[i]
-      result[key] = isMergeableObject(target[key]) ? merge(Array.isArray(target[key]) ? [] : {}, target[key]) : target[key]
+      result[key] = map(target[key])
     }
   } else if (typeof target !== 'object' || target === null) {
-    return isMergeableObject(source) ? merge(Array.isArray(source) ? [] : {}, source) : source
+    return map(source)
   }
 
   const keys = Object.keys(source)
@@ -32,13 +36,9 @@ function mergeObject (target, source) {
         continue
       }
     }
-    result[key] = isMergeableObject(source[key]) ? merge(Array.isArray(source[key]) ? [] : {}, source[key]) : source[key]
+    result[key] = map(source[key])
   }
   return result
-}
-
-function map (entry) {
-  return isMergeableObject(entry) ? merge(Array.isArray(entry) ? [] : {}, entry) : entry
 }
 
 function merge (target, source) {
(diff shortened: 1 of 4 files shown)


merge.js
```diff
@@ -0,0 +1,67 @@
+'use strict'
+
+// based on https://github.com/TehShrike/deepmerge
```
mcollina, line 3: please include the original license of that code

test/merge.test.js
```diff
@@ -0,0 +1,533 @@
+// based on https://github.com/TehShrike/deepmerge/tree/master/test
```
mcollina, line 1: Please include the original license of that code

mcollina (changes requested):
Good work! Only a couple of nits

mcollina:
I think we could use a faster version of that modules in a couple more places.

Maybe we should put it in a module? cc @fastify/core

dalisoft:
@mcollina I am vote for make modules within @fastify organization to separate module and keep this module readable

Fdawgs (commented):
Needs rebasing as it's reintroducing chalk, that you just recently removed in another PR 😊

Uzlopak:
If you want to have it in a separate module, then I would recommend to create a new project in the org, like we did with the accept-negotatior ;).

Uzlopak:
Closing in favor of #483

Issue #479, opened 2022-06-29
Schema that worked in v4 no longer works in v5


smonn:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.2.0

### Plugin version

5.0.3

### Node.js version

16.15.1

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.4

### Description

I encountered this error after upgrading to Fastify v4.2.0 which included v5 of fast-json-stringify. This error is thrown:

```
Error: no schema with key or ref "schema#/properties/union/anyOf/0"
    at Ajv.validate (.../node_modules/ajv/lib/core.ts:359:21)
    at anonymous0 (eval at build (.../node_modules/fast-json-stringify/index.js:177:23), <anonymous>:45:20)
    at main (eval at build (.../node_modules/fast-json-stringify/index.js:177:23), <anonymous>:6:15)
    at Object.<anonymous> (.../src/index.ts:80:5)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Module.m._compile (.../node_modules/ts-node/src/index.ts:1597:23)
    at Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Object.require.extensions.<computed> [as .ts] (.../node_modules/ts-node/src/index.ts:1600:12)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
```

For reference, I also reported an issue over in the TypeBox repo: https://github.com/sinclairzx81/typebox/issues/184

### Steps to Reproduce

Use the following schema to validate `{ name: 'foo', union: 'a8f1cc50-5530-5c62-9109-5ba9589a6ae1' }` and you should see the error. In v4 there is no error and it stringifies things correctly.

```json
{
  "$id": "schema",
  "type": "object",
  "allOf": [
    {
      "$id": "base",
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        }
      },
      "required": [
        "name"
      ]
    },
    {
      "$id": "inner_schema",
      "type": "object",
      "properties": {
        "union": {
          "$id": "id",
          "anyOf": [
            {
              "format": "uuid",
              "$id": "guid",
              "type": "string"
            },
            {
              "format": "email",
              "$id": "email",
              "type": "string"
            }
          ]
        }
      },
      "required": [
[...]

PR #480, opened 2022-06-29
fix: reference after merge allOf

Eomm:
fixes https://github.com/fastify/fast-json-stringify/issues/479

The issue:

after the merge of the `allOf` object, we lose the `$id`s of the merged objects, so it is not possible to call `ajv.validate` when one of the `allOf` objects contains an `anyOf` or `allOf` item.

This fix adds a reference to the merged object to use it as a validator.

Any better ideas?

### DIFF
--- a/index.js
+++ b/index.js
@@ -489,6 +489,10 @@ function mergeAllOfSchema (location, schema, mergedSchema) {
     }
   }
   delete mergedSchema.allOf
+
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
let's ship the fix

zekth (approved):
LGTM. I don't have any better implementation in mind

PR #482, opened 2022-06-29
fix: reset json pointer in merged allOf schema

### LINKED ISSUE #481: Ajv validation inside allOf
### Prerequisites

### Issue

There is a fastify issue https://github.com/fastify/fastify/issues/4092
I reduced the test example.

```js
test('anyOf inside allOf', (t) => {
  t.plan(1)

  const schema = {
    anyOf: [
      {
        type: 'object',
        allOf: [
          {
            properties: {
              a: {
                anyOf: [
                  { const: 'A1' },
                  { const: 'A2' }
                ]
              }
            }
          },
          {
            properties: {
              b: { const: 'B' }
            }
          }
        ]
      }
    ]
  }

  const object = { a: 'A1', b: 'B' }
  const output = stringify(object)

  t.equal(output, JSON.stringify(object))
})
```

I'm working on that.

ivan-tymoshenko:
#### Checklist

Fix #481
Fix https://github.com/fastify/fastify/issues/4092

#480 sets merged schema as a location, but doesn't set json pointer to the root of this merged schema.

### DIFF
--- a/index.js
+++ b/index.js
@@ -493,6 +493,7 @@ function mergeAllOfSchema (location, schema, mergedSchema) {
   mergedSchema.$id = `merged_${randomUUID()}`
   ajvInstance.addSchema(mergedSchema)
   location.schemaId = mergedSchema.$id
+  location.jsonPointer = '#'
 }
 
 function buildInnerObject (location) {
--- a/test/ref.test.js
+++ b/test/ref.test.js
@@ -1891,3 +1891,38 @@ test('input schema is not mutated', (t) => {
   t.equal(output, '{"obj":"test"}')
   t.same(schema, clonedSchema)
 })
+
+test('anyOf inside allOf', (t) => {
+  t.plan(1)
+
+  const schema = {
+    anyOf: [
+      {
+        type: 'object',
+        allOf: [
+          {
+            properties: {
+              a: {
+                anyOf: [
+                  { const: 'A1' },
+                  { const: 'A2' }
+                ]
+              }
+            }
+          },
+          {
+            properties: {
+              b: { const: 'B' }
+            }
+          }
+        ]
+      }
+    ]
+  }
+
+  const object = { a: 'A1', b: 'B' }
+  const stringify = build(schema)
+  const output = stringify(object)
+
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -493,6 +493,7 @@ function mergeAllOfSchema (location, schema, mergedSchema) {
   mergedSchema.$id = `merged_${randomUUID()}`
   ajvInstance.addSchema(mergedSchema)
   location.schemaId = mergedSchema.$id
+  location.jsonPointer = '#'
```
climba03003, line 496: Will it be problem of deeply nested schema?

ivan-tymoshenko: It should work with nested schemas. Can you provide an example of what you mean?

climba03003: I get my answer when I create the example.
Here comes one problem related to previous fix.

Recursive `$ref` will be a problem when using `allOf`.

```js
{
  $id: 'base',
  anyOf: [
    {
      type: 'object',
      allOf: [
        {
          properties: {
            a: {
              anyOf: [
                { const: 'A1' },
                { const: 'A2' }
              ]
            }
          }
        },
        {
          properties: {
            b: { const: 'B' }
          }
        },
        {
          properties: {
            c: {
              oneOf: [
                { $ref: 'base' },
                { type: 'null' }
              ]
            }
          }
        }
      ]
    }
  ]
}
```

climba03003 (approved):
LGTM.

index.js
ivan-tymoshenko: I guess we can fix this. But there are some other problems with merging schemas that we should also. We can't use deepMerge for merging. I was trying to make it better #453, but it's not enough. We can check how [json-schema-merge-allof](https://www.npmjs.com/package/json-schema-merge-allof) works.

In general, we have some places where we have troubles (if/then, oneOf, anyOf, allOf, type array), bc json schema format designed for validation and described input types and we use json schema for serialization and ask people to describe output types.

climba03003: I think it worth an issue for discussion.

mcollina (approved):
lgtm

PR #483, opened 2022-06-30
perf: replace deepmerge npm-package with @fastify/deepmerge

Uzlopak:
I have just one worry: @fastify/deepmerge has prototype protection handling. What happens if in the json schema an type: object has the key prototype or constructor? It would mean that those keys are not merged/carried over. should we change deepmerge to have an option where we create Objects by using Object.create(null) so that a prototype pollution is not possible?

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,7 +2,7 @@
 
 /* eslint no-prototype-builtins: 0 */
 
-const merge = require('deepmerge')
+const merge = require('@fastify/deepmerge')()
 const clone = require('rfdc')({ proto: true })
 const fjsCloned = Symbol('fast-json-stringify.cloned')
 const { randomUUID } = require('crypto')
--- a/package.json
+++ b/package.json
@@ -50,9 +50,9 @@
     "webpack": "^5.40.0"
   },
   "dependencies": {
+    "@fastify/deepmerge": "^1.0.0",
     "ajv": "^8.10.0",
     "ajv-formats": "^2.1.1",
-    "deepmerge": "^4.2.2",
     "fast-uri": "^2.1.0",
     "rfdc": "^1.2.0"
   },


ivan-tymoshenko:
I know this PR should just update to the fastest merge version, but if we touch on this topic, the normal merge function is not suitable for schema merging. There are two places now where we are merging schema if/then and allOf, and both of them should be merged according to the rules of the schemes. #453 fixed some cases, but, I guess it doesn't cover all cases. We can take a look at [json-schema-merge-allof](https://www.npmjs.com/package/json-schema-merge-allof) and use it or take some ideas.

Uzlopak:
@ivan-tymoshenko 

Wow thats crazy that they wrote that in lodash :D.

Eomm (approved):
> There are two places now where we are merging schema if/then and allOf, and both of them should be merged according to the rules of the schemes

I think this is another codebase aspect and does not relate to this specific dep change

Uzlopak:
@Eomm 
I am currently thinking if I should implement a mode where we create Objects without prototype (Object.create(null)) and then deepmerge can handle keys like prototype and constructor. Should I focus on this next?

Eomm:
> Should I focus on this next?

If you would like to work it, extending the `@fastify/deepmerge` go for it.
In this case we should measure it against rfdc: https://www.npmjs.com/package/rfdc

> Object.create(null)

Regarding this, I was thinking quite the same about the `fastify` repo. We have some `{}` around that could be a simple object without a prototype and I wonder how much memory benefit the server instance could get from this change

Uzlopak:
created a PR for deepmerge to handle also nullPrototype 
https://github.com/fastify/deepmerge/pull/2

ivan-tymoshenko (approved):

Uzlopak:
I really think we should merge and publish first the nullPrototype PR and then use the nullPrototype mode for merging Schema.

mcollina:
I don't think we need null prototype protection at all here. Schemas are considered as safe as code is.

Uzlopak:
If we dont need null prototype, then this PR should be ready to merge ;)

mcollina (approved):
lgtm

PR #484, opened 2022-07-02
fix: string date format validation

ivan-tymoshenko:
#### Checklist

### DIFF
--- a/ajv.js
+++ b/ajv.js
@@ -7,29 +7,22 @@ const ajvFormats = require('ajv-formats')
 module.exports = buildAjv
 
 function buildAjv (options) {
-  const ajvInstance = new Ajv({ ...options, strictSchema: false, validateSchema: false, uriResolver: fastUri })
+  const ajvInstance = new Ajv({
+    ...options,
+    strictSchema: false,
+    validateSchema: false,
+    allowUnionTypes: true,
+    uriResolver: fastUri
+  })
+
   ajvFormats(ajvInstance)
 
-  const validateDateTimeFormat = ajvFormats.get('date-time').validate
-  const validateDateFormat = ajvFormats.get('date').validate
-  const validateTimeFormat = ajvFormats.get('time').validate
-
   ajvInstance.addKeyword({
-    keyword: 'fjs_date_type',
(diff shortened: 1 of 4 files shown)


ivan-tymoshenko:
Please make sure that this test works as expected.
https://github.com/fastify/fast-json-stringify/blob/e083ca6778b6cfeb09be24dd15bae5c6c99791ee/test/date.test.js#L492

serializer.js
```diff
@@ -69,36 +69,36 @@ module.exports = class Serializer {
...
-  asDatetime (date) {
-    const quotes = '"'
+  asDateTime (date) {
+    if (date === null) return '""'
     if (date instanceof Date) {
-      return quotes + date.toISOString() + quotes
+      return '"' + date.toISOString() + '"'
     }
-    return this.asString(date)
+    throw new Error(`The value "${date}" cannot be converted to a date-time.`)
```
climba03003, line 77: What if I passed a string that fulfill `date-time`?
It seems like it only accept `Date` object after this change.

ivan-tymoshenko: If you pass the string it will be stringified in the `asString` method.

climba03003 (approved):
LGTM.

ajv.js
```diff
@@ -7,29 +7,22 @@ const ajvFormats = require('ajv-formats')
...
-        return validateDateTimeFormat(date)
-      }
-      if (schema === 'date') {
-        return validateDateFormat(date)
-      }
-      if (schema === 'time') {
-        return validateTimeFormat(date)
-      }
-      return false
+    keyword: 'fjs_type',
```
mcollina, line 21:
```suggestion
    keyword: 'fjs_date_type',
```


I think we should try not to change this.

ivan-tymoshenko: @mcollina I agree, but it is a completely different meaning of this property now. It's not a date format anymore. It describes how ajv should validate custom type at all. For example, for string type, there would be not only the Date object validation but also Regexp objects and some other exceptions.

mcollina (approved):
lgtm

Eomm (approved):

PR #493, opened 2022-07-13
Fix stringify for non-standard, non-Date objects like luxon.DateTime

### LINKED ISSUE #492: `toString` existence check on objects does not work using Object.hasOwnProperty if the object is a class instance
### Prerequisites

### Fastify version

4.2.1

### Plugin version

5.1.0

### Node.js version

18.4.0

### Operating system

macOS

12.4

### Description

The generated code for creating strings from objects checks for the existence of a `toString` function, but through Object.hasOwnProperty.call(object, 'toString') (https://github.com/fastify/fast-json-stringify/blob/master/index.js#L904). This does not work for class instances.

An example for where this goes south: if you're using Luxon DateTimes in your code for dates instead of the regular JavaScript Date object and need it as a string in your API response, the conversion fails (but the error message references a stringified Luxon DateTime because there is a toString present on the class, just not as a "own property".

### Steps to Reproduce

dependencies:
```
npm install fastify@4 luxon@3
```

index.js:
```
const l = require('luxon')
const f = require('fastify')

const app = new f.fastify()

const opts = {
  schema: {
    response: {
      200: {
        type: 'object',
        properties: {
          date: { type: 'string', format: 'date-time' }
        }
      }
    }
  }
}

app.get('/', opts, () => {
  return { date: l.DateTime.now() }
})

app.listen({ port: 8000 })
```

call route
```
curl localhost:8000
```

error: 
```
{"statusCode":500,"error":"Internal Server Error","message":"The value \"2022-07-13T16:03:06.938+02:00\" does not match schema definition."}
```

Note that the value reported is a valid date-time string

### Expected Behavior

The string conversion should succeed by calling toString() on the instance.

pvanagtmaal:
#### Checklist

- [ ] documentation is changed or added (no changes required in this case)

Fixes #492 

The check has become quite ugly, mainly because

- All Objects have a toString and in that case we need to hasOwnProperty check, but only for objects whose constructor is not that of Object
- Logic for Date objects is different so we need to filter these out

Benchmark:

```
$ npm run bench

> fast-json-stringify@5.1.0 bench
> node ./benchmark/bench.js

short string............................................. x 46,201,257 ops/sec ±0.71% (593 runs sampled)
long string.............................................. x 11,675 ops/sec ±0.42% (594 runs sampled)
number................................................... x 988,806,479 ops/sec ±0.52% (594 runs sampled)
integer.................................................. x 149,494,519 ops/sec ±0.66% (593 runs sampled)
formatted date-time...................................... x 4,269,928 ops/sec ±0.43% (594 runs sampled)
formatted date........................................... x 2,846,144 ops/sec ±0.26% (595 runs sampled)
formatted time........................................... x 2,850,038 ops/sec ±0.27% (594 runs sampled)
short array of numbers................................... x 103,756 ops/sec ±0.48% (592 runs sampled)
short array of integers.................................. x 95,856 ops/sec ±0.69% (594 runs sampled)
short array of short strings............................. x 26,121 ops/sec ±0.32% (595 runs sampled)
short array of long strings.............................. x 26,071 ops/sec ±0.42% (594 runs sampled)
short array of objects with properties of different types x 14,203 ops/sec ±0.57% (592 runs sampled)
object with number property.............................. x 995,475,152 ops/sec ±0.28% (596 runs sampled)
object with integer property............................. x 151,170,986 ops/sec ±0.71% (594 runs sampled)
object with short string property........................ x 29,965,328 ops/sec ±0.31% (594 runs sampled)
object with long string property......................... x 11,737 ops/sec ±0.45% (594 runs sampled)
object with properties of different types................ x 3,142,251 ops/sec ±0.54% (594 runs sampled)
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -901,7 +901,7 @@ function buildValue (location, input) {
           switch (type) {
             case 'string': {
               code += `
-                ${statement}(${input} === null || typeof ${input} === "${type}" || ${input} instanceof RegExp || (typeof ${input} === "object" && Object.hasOwnProperty.call(${input}, "toString")))
+                ${statement}(${input} === null || typeof ${input} === "${type}" || ${input} instanceof RegExp || (typeof ${input} === "object" && ((!(${input} instanceof Date) && typeof ${input}["toString"] === "function" && ${input}.__proto__.constructor.name !== "Object") || Object.hasOwnProperty.call(${input}, "toString"))))
                   ${nestedResult}
               `
               break
--- a/test/date.test.js
+++ b/test/date.test.js
@@ -525,3 +525,18 @@ test('nullable date', (t) => {
 
   t.same(result, `"${DateTime.fromJSDate(data).toISODate()}"`)
 })
+
+test('Luxon Date stringify', (t) => {
+  t.plan(1)
+
+  const schema = {
+    type: 'string',
+    format: 'date-time'
+  }
+
+  const date = DateTime.now()
+  const stringify = build(schema)
+  const output = stringify(date)
+
+  t.equal(output, `"${date.toISO()}"`)
+})


ivan-tymoshenko (changes requested):
I think we drop support for any custom date format such as moment or luxon here. @mcollina @jsumners what do you think?

pvanagtmaal:
@ivan-tymoshenko please note that this change is not only about luxon, it is about **any** class that should be converted to a string, **except** the Date class

mcollina:
This change seems pretty simple compared to what was removed in https://github.com/fastify/fast-json-stringify/pull/442. Specifically, this is not calling any specific method of moment or luxon  (like we did before).

@ivan-tymoshenko do you see any specific issues with this particular change? It seems a good middle-ground solution.

ivan-tymoshenko:
It will cause a validation problem. How Ajv will find out if it's a date or not after these changes?

After all, we should specify what we mean by object type. I thought this comment explains it https://github.com/fastify/fast-json-stringify/pull/441#discussion_r884798650.

pvanagtmaal:
The logic for Date objects did not change with these changes. The only actual change is that a different check for toString existence is performed on objects.

I don't see why you would want to tell ajv about the object being any specific instance of a class and I still don't see why you think this is specific to Dates.

mcollina:
@ivan-tymoshenko could you produce a small test case that would fail for this change?

ivan-tymoshenko:
We have the way we defined date type (not js Date object/type) in the schema.
```js
{ type: 'string', format: 'date' }
```

We use this information not only to serialize but also for validation in some cases. If we add the logic you suppose, we need to change the validation logic to be consistent. Now the logic is next: date type it's either string formatted as a date or instance of Date. You suppose to call every object that has the toString method as a date type. I'm not sure it's a good idea.

Here is an example of when we use date validation.

```js
test('DateTime object as a date', (t) => {
  t.plan(1)

  const schema = {
    anyOf: [
      {
        type: 'string',
        format: 'date-time'
      }
    ]
  }

  const data = DateTime.now()
  const result = stringify(data)

  t.same(result, `"${data.toString()}"`)
})
```

pvanagtmaal:
@ivan-tymoshenko I finally found what you meant... Your test case does indeed fail. There is a ajv format defined in ajv.js that checks for Date instances if the fjs_type is an object.

Removing the ajv format altogether does not cause any present tests to fail, and neither does your test case. Of course, it is easy to create a test that doesn't pass without the format (just pass a random object to stringify).

The question is, is it the responsibility of fast-json-stringify to check for a valid date object? I can see why you dropped support for Luxon / MomentJS in #441, but wouldn't it make sense to drop support for Date objects as well and make it the user's responsibility to pass an object that stringifies to a date-time / date / time string? The resulting string is the one that is important, the object that results in that string really isn't...

mcollina:
`Date` is supported by JSON.stringify() natively.

pvanagtmaal:
So is luxon DateTime: 
```
JSON.stringify(luxon.DateTime.now())
'"2022-07-14T12:24:44.670+02:00"'
```

ivan-tymoshenko:
> So is luxon DateTime:
> 
> ```
> JSON.stringify(luxon.DateTime.now())
> '"2022-07-14T12:24:44.670+02:00"'
> ```

Yes, if you set up it as a string type without date format fjs will serialize it. But it's not a date. It's your custom object that have toString method.

PR #497, opened 2022-07-31
test: add test for issue #342

### LINKED ISSUE #342: When no schemas of a oneOf schema element match null is returned rather than raising an error.
### Prerequisites

### Fastify version

3.18.1

### Plugin version

_No response_

### Node.js version

14.17.3

### Operating system

macOS

11.4

### Description

Assume using Fastify, there is a response schema for a 200 status code to a route that has a the following JSON schema specified:

```json
{
  "type": "object",
  "additionalProperties": false,
  "required": ["data"],
  "properties": {
    "data": {
      "type": "array",
      "minItems": 1,
      "items": {
        "oneOf": [
          {
            "type": "string"
          },
          {
            "type": "number"
          }
        ]
      }
    }
  }
}
```

This says that the response should be an object with a property called "data" which consists of an array where items can be a string or a number.

If the function generating the response returns a response like `{"data": [false, "testing"]}` the response generated by `fast-json-stringify` will be `{"data": [null]}` which violates the schema of the response.  This will likely surprise the callers of the route.

**No exception will be raised, which seems unsound.**

This is because of this logical branch:

https://github.com/fastify/fast-json-stringify/blob/ab95d2c4f6f577e3808f055e8fb94f75f7944bcd/index.js#L1217

If none of the available oneOf schemas match, an error isn't raised a null is just returned for the the produced portion of the schema.

### Steps to Reproduce

Run the example above.

### Expected Behavior

I think the right thing to do is raise an exception that the response does not match the expected schema and return a different HTTP status code such as 500 rather than returning a response that violates the schema.

ivan-tymoshenko:
#### Checklist

Close #342
Issue #342 was fixed by #445. This PR adds a test that reproduces the test case.

### DIFF
--- a/test/nullable.test.js
+++ b/test/nullable.test.js
@@ -464,3 +464,37 @@ test('nullable type in the schema', (t) => {
   t.same(stringify(data), JSON.stringify(data))
   t.same(stringify(null), JSON.stringify(null))
 })
+
+test('throw an error if the value doesn\'t match the type', (t) => {
+  t.plan(2)
+
+  const schema = {
+    type: 'object',
+    additionalProperties: false,
+    required: ['data'],
+    properties: {
+      data: {
+        type: 'array',
+        minItems: 1,
+        items: {
+          oneOf: [
+            {
+              type: 'string'
+            },
+            {
+              type: 'number'
+            }
+          ]
+        }
+      }
+    }
+  }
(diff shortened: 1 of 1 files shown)


Uzlopak (approved):
LGTM

PR #504, opened 2022-08-14
refactor: create separate schema for Ajv validation

ivan-tymoshenko:
How it's working now:
It takes a user-defined schema, adds necessary adjustments for Ajv (now it's only for date type), and then works with modified schema.

How it will work:
It will clone a user-defined schema, modify it and pass it to the Ajv. Serialization will work with the original schema. The only requirement we must follow is that the schema must not change its structure during the modification. In other words, the JSON pointers must match in two schemas, but the schemas themselves (type, format, keywords, ... etc.) may differ.

Motivation:
It's too complicated to maintain one syntax for serialization and validation rules. Our code depends on Ajv validation syntax.

Depends on #507


Uzlopak:
It makes sense to me.

mcollina (approved):
lgtm

ivan-tymoshenko:
There are two updates.
1. I moved all Ajv logic to the Validator class. The Validator is a class Adapter. It has two public methods `addSchema` and `validate`. It should validate schema according to our rules. It encapsulates all Ajv validation logic in one place. 
2. There is a small change in API. If you call the build method in the debug mode, one of the return parameters would be the Validator instance instead of the Ajv instance.

ref-resolver.js
```diff
@@ -11,8 +11,10 @@ class RefResolver {
     if (schema.$id !== undefined && schema.$id.charAt(0) !== '#') {
       schemaId = schema.$id
     }
-    this.insertSchemaBySchemaId(schema, schemaId)
-    this.insertSchemaSubschemas(schema, schemaId)
+    if (this.getSchema(schemaId) === undefined) {
```
climba03003, line 14: Is this changes itself self contained?
Do we need a test case for it?

climba03003 (approved):
The code itself look good to me and the changes on `debug` output is breaking.

@ivan-tymoshenko
You are the one focusing on this package. Can I ask how many breaking changes are you expected in the near future? I am considering should we group a bunch of them before another major.

ivan-tymoshenko:
@climba03003 I can leave the Ajv instance and remove it in the next major release.

ref-resolver.js
ivan-tymoshenko: I didn't add this check. I just moved it inside the class.
https://github.com/fastify/fast-json-stringify/blob/b0bb1a699e5faed3657ad8c1d665920e29304be5/index.js#L113

Issue #502, opened 2022-08-07
Worse performance with response schema when 'const' keyword is used


DanieleFedeli:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.3.0

### Node.js version

16.13.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

13.0

### Description

When specifying the response schema with 'const' I noticed that the RPS decreased. 

### Steps to Reproduce

Options used: 
```
{
  schema: {
    response: {
        200: {
          type: "object",
          properties: {
            msg: { type: "string" },
          }
      }
   }
}
```
Schema benchmark: 		
<img width="718" alt="Screenshot 2022-08-07 alle 12 06 23" src="https://user-images.githubusercontent.com/37077048/183285845-a2fd096e-b03f-411a-87b8-b729f961de5d.png">

Options used: 
```
{
  schema: {
    response: {
        200: {
          type: "object",
          properties: {
            msg: { const: "hello world" },
          }
      }
   }
}
```
Schema with const:
<img width="675" alt="Screenshot 2022-08-07 alle 12 08 13" src="https://user-images.githubusercontent.com/37077048/183285902-c532e8ed-07a7-49f8-ac9a-72e81ff78b83.png">

Repository with the code: https://github.com/DanieleFedeli/fastify-response-schema-issue

### Expected Behavior

I expect at least the same performance when specifying const in the schema.

Uzlopak:
How is the performance if we also use type: "string" or if we use enum instead of const?

DanieleFedeli:
With type string

```javascript
{
  type: "object",
  properties: {
    msg: { type: "string", const: "hello world" },
  }
}
```
<img width="765" alt="Screenshot 2022-08-07 alle 16 18 25" src="https://user-images.githubusercontent.com/37077048/183295181-9b4c6a90-9e1f-4e37-9f50-e461bc3aeed6.png">

With enum

``` javascript
{
  type: "object",
  properties: {
    msg: { enum: ["hello world"] },
  }
}
```
<img width="741" alt="Screenshot 2022-08-07 alle 16 20 00" src="https://user-images.githubusercontent.com/37077048/183295248-813cde28-b707-45ee-b029-089ca21a7399.png">


It improves drastically

climba03003:
I would say it is expected. 
[...]

PR #505, opened 2022-08-16
Improve performance for const schemas

DanieleFedeli:
#### Checklist

As per issue #502, when using a const schema, the performance degrades.

Benchmark pre fix:
Screenshot 2022-08-16 alle 10 50 37

Benchmark post fix:
Screenshot 2022-08-16 alle 11 02 14

As you can see, object const schemas did not improve. I am working on that.

### DIFF
--- a/index.js
+++ b/index.js
@@ -346,11 +346,11 @@ function buildCode (location) {
       schema.properties[key] = propertyLocation.schema
     }
 
-    // Using obj['key'] !== undefined instead of obj.hasOwnProperty(prop) for perf reasons,
-    // see https://github.com/mcollina/fast-json-stringify/pull/3 for discussion.
-
     const sanitized = JSON.stringify(key)
     const asString = JSON.stringify(sanitized)
+
+    // Using obj['key'] !== undefined instead of obj.hasOwnProperty(prop) for perf reasons,
+    // see https://github.com/mcollina/fast-json-stringify/pull/3 for discussion.
 
     code += `
       if (obj[${sanitized}] !== undefined) {
@@ -901,7 +901,7 @@ function buildValue (location, input) {
           switch (type) {
             case 'string': {
               code += `
-                ${statement}(${input} === null || typeof ${input} === "${type}" || ${input} instanceof RegExp || (typeof ${input} === "object" && Object.hasOwnProperty.call(${input}, "toString")))
+                ${statement}(${input} === null || typeof ${input} === "${type}" || ${input} instanceof RegExp || (typeof ${input} === "object" && Object.prototype.hasOwnProperty.call(${input}, "toString")))
                   ${nestedResult}
               `
               break
@@ -987,7 +987,7 @@ function extendDateTimeType (schema) {
 function isEmpty (schema) {
   // eslint-disable-next-line
   for (var key in schema) {
-    if (schema.hasOwnProperty(key) && schema[key] !== undefined) {
+    if (Object.prototype.hasOwnProperty.call(schema, key) && schema[key] !== undefined) {
       return false
     }
   }


mcollina:
wow, good numbers! Is there anything missing here (the PR is in draft)?

DanieleFedeli:
I am trying to figure out if I can improve const object schemas as well,

DanieleFedeli:
@Uzlopak  I tried but object schemas are not serialized correctly.
test/const.test.js fails
Screenshot 2022-08-16 alle 11 31 11

DanieleFedeli:
Manage to improve object as well 🚀
Screenshot 2022-08-16 alle 12 40 23

DanieleFedeli:
Improve const strings and const objects 🚀
Screenshot 2022-08-16 alle 13 46 53

mcollina (approved):
lgtm

index.js
```diff
@@ -805,159 +818,205 @@ function buildValue (location, input) {
...
-  switch (type) {
-    case 'null':
-      code += 'json += serializer.asNull()'
-      break
-    case 'string': {
-      funcName = nullable ? 'serializer.asStringNullable.bind(serializer)' : 'serializer.asString.bind(serializer)'
-      code += `json += ${funcName}(${input})`
-      break
+  if ('const' in schema) {
+    const stringifiedSchema = JSON.stringify(schema.const)
```
Uzlopak, line 822: I am not so convinced by this, as it is not deterministic and would break probably on circular references and bigint.

Uzlopak: So probably returns false if you set const: { a: 1, b: 2 } then it would return false on { b: 2, a: 1} 

Maybe we need here rfdc instead of stringification for objects 

Also what happens when we provide a Date.

E.g. 

const: { }

and JSON.stringify on Date is also returning { }

So probably it would pass on Dates.

Uzlopak: sry, my mistake JSON.stringify(Date) returns the ISODateTime. But still a potential pitfall.

Uzlopak (changes requested):
I would expect this to pass.

```js
test('schema with const object', (t) => {
  t.plan(1)

  const schema = {
    type: 'object',
    properties: {
      foo: { const: { a: 1, b: 2 } }
    }
  }

  t.doesNotThrow(() => stringify({
    foo: { b: 2, a: 1 }
  }))
})
```

Uzlopak:
tbh. I thought the first commit was good enough as it was improving the performance of primitives. 

And I actually dont like the notation of passing an Object simply with const.

Because this works perfect and is the JSONSchema-way

```js
test('schema with const object', (t) => {
  t.plan(2)

  const schema = {
    type: 'object',
    properties: {
      foo: {
        type: 'object',
        additionalProperties: false,
        required: ['a', 'b'],
        properties: {
          a: { const: 1 },
          b: { const: 2 }
        }
      }
    }
  }

  t.doesNotThrow(() => stringify({
    foo: { b: 2, a: 1 }
  }))
  t.doesNotThrow(() => stringify({
    foo: { a: 1, b: 2 }
  }))
})
```

DanieleFedeli:
Ok then I will revert the last commit

Uzlopak:
Please wait for the feedback of @ivan-tymoshenko. Maybe he has some idea.

index.js
```diff
@@ -805,159 +818,205 @@ function buildValue (location, input) {
...
 
-  switch (type) {
-    case 'null':
-      code += 'json += serializer.asNull()'
-      break
-    case 'string': {
-      funcName = nullable ? 'serializer.asStringNullable.bind(serializer)' : 'serializer.asString.bind(serializer)'
-      code += `json += ${funcName}(${input})`
-      break
+  if ('const' in schema) {
```
Eomm, line 821: I would add an option to turn this feature on, or it will be a major version with a harder distribution across userland

ivan-tymoshenko:
@Uzlopak @climba03003 Is there a reason we match data with a const before inserting the const? Can't we just insert the const? We have a vague explanation of when we validate data and when we don't, but it looks like we can avoid it for performance purposes. WDYT?

index.js
Eomm: Agree on `rfdc` to remove any `__proto__`

We know that schemas should be treated as code, but adding this `const` check gives to the user the possibility to inject bad code

Eomm (changes requested):
I want to start a discussion here: why not coerce the output value to the `const` value instead?
A json schema with `const` is telling me ([standard quotes](https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-6.1.3) :

> An instance validates successfully against this keyword if its value
>   is equal to the value of the keyword.

So, a serialization perspective should be "no matter what the input is, the output will be the const value."

I think `fast-json-stringify` is not a validator, it is an information filter (with rare exceptions such as `anyOf` and `allOf`)

Uzlopak:
i have no strong opinion on this. I just think that he first commit was a nice patch, which had only the issue, that it maybe could have resulted in some hickups like what if the const is a bigint or function. But it would have improved the performance significantly, if somebody would have provided some json schema without type information. 

I assume that the following commits were under the impression that we want to improve the performance under any circumstances. Thats why I wrote, that I liked the first commit, as it was small and very effective. It does not even make fast-json-stringify to a validator as you mentioned @Eomm .

I think the first commit is a good base for a PR. But I would actually also not simply return typeof value.

more like 
```js
function getJSONSchemaType(value) {
  const type = typeof value
  if (type === 'string') {
    return type;
  } else if (type === 'boolean') {
    return type
  } else if (type === 'number') {
    // NaN and Infinity and -Infinity are not serializable
    if (isFinite(value)) {
      return type
    }
  }
}
```

ivan-tymoshenko:
> i have no strong opinion on this. I just think that he first commit was a nice patch, which had only the issue, that it maybe could have resulted in some hickups like what if the const is a bigint or function. But it would have improved the performance significantly, if somebody would have provided some json schema without type information.
> 
> I assume that the following commits were under the impression that we want to improve the performance under any circumstances. Thats why I wrote, that I liked the first commit, as it was small and very effective. It does not even make fast-json-stringify to a validator as you mentioned @Eomm .
> 
> I think the first commit is a good base for a PR. But I would actually also not simply return typeof value.
> 
> more like
> 
> ```js
> function getJSONSchemaType(value) {
>   const type = typeof value
>   if (type === 'string') {
>     return type;
>   } else if (type === 'boolean') {
>     return type
>   } else if (type === 'number') {
>     // NaN and Infinity and -Infinity are not serializable
>     if (isFinite(value)) {
>       return type
>     }
>   }
> }
> ```

@Uzlopak, I think we've got a slight misunderstanding. What I and @Eomm mean is that FJS already validates the data when inserting a constant, and it's not necessary. Without validation, it would be even simpler than a first commit and would work with any type.

test/const.test.js
```diff
@@ -65,3 +145,53 @@ test('schema with const and invalid object', (t) => {
     t.ok(err)
   }
 })
+
+test('schema with const and invalid number', t => {
```
Eomm, line 149: Let me make an example @Uzlopak 

talking about the behaviour, I think this test should be always successful:

```
const x = stringify({
      foo: 2
    })
// x should be '{"foo":1}'
```

No matter what the input object is, the output should be always equals to `const`

climba03003:
> So, a serialization perspective should be "no matter what the input is, the output will be the const value."

Just one question, how do we handle array of types that include `null`.

```json
{
  type: ["string", "null"],
  const: "foo"
}
```

It is a special case that user may expect the `const` can be null-able.

Uzlopak:
That is kind of contradicting.

climba03003:
> That is kind of contradicting.

Yes, I thought of it. But, I can't really think of a way to optionally parse the `const`.

```js
const schema = {
  "type": "object",
  "properties": {
    "foo": {
      "const": "bar"
    }
  }
}

stringify({ "foo": "bar" }) // { "foo": "bar" }
stringify({ "foo": "baz" }) // { "foo": "bar" }
stringify({ "foo": 1 }) // { "foo": "bar" }

stringify({}) // { "foo": "bar" } ??? what if I want { } only
```

DanieleFedeli:
So what is the plan? 
Should I revert to [68ff81c](https://github.com/DanieleFedeli/fast-json-stringify/tree/68ff81cc29979413b1db7ed9b13bd5cbfac8db32) and then open an issue for objects?

Eomm:
> So what is the plan? Should I revert to [68ff81c](https://github.com/DanieleFedeli/fast-json-stringify/tree/68ff81cc29979413b1db7ed9b13bd5cbfac8db32) and then open an issue for objects?

No, that commit is not a problem.
The discussion here is that the following line applies a validation processing that is unwanted:

```
throw new Error(\`Item $\{JSON.stringify(${input})} does not match schema definition.\`)`
```

I think that we should force the `const` value into the serialized object whatever the `input` object is.


The @climba03003 https://github.com/fastify/fast-json-stringify/pull/505#issuecomment-1217390075 is the perfect summary of how a serializer should use a `const`.
_The `type: null` case could be added in a follow PR because it is an edge case_

Moreover, this feature should be hidden by an option flag: `coerceConst: Bool` if you need to ship this feature in the current major version.

If someone does not agree with my opinion, please explain.

DanieleFedeli:
@Eomm In this case I agree with you. 
I added validation because I saw the way how const schema was handled:
```javascript
case undefined:
.
.
.
.
else if ('const' in schema) {
 code += ` if(ajv.validate(${JSON.stringify(schema)}, ${input}))
    json += '${JSON.stringify(schema.const)}'
  else
    throw new Error(\`Item $\{JSON.stringify(${input})} does not match schema definition.\`)`
}
```

But if it not needed I am happy to remove it

ivan-tymoshenko:
> @Eomm In this case I agree with you. I added validation because I saw the way how const schema was handled:
> 
> ```js
> case undefined:
> .
> .
> .
> .
> else if ('const' in schema) {
>  code += ` if(ajv.validate(${JSON.stringify(schema)}, ${input}))
>     json += '${JSON.stringify(schema.const)}'
>   else
>     throw new Error(\`Item $\{JSON.stringify(${input})} does not match schema definition.\`)`
> }
> ```
> 
> But if it not needed I am happy to remove it

Please don't forget about null case.

PR #507, opened 2022-08-21
refactor: create json schema ref resolver

### LINKED ISSUE #506: possible performance impact when updating from v4.2.0 to v5.0.0
### Prerequisites

### Issue

Hi!

Today I was doing some dependency updates for my fastify apps, when I noticed a bigger InitDuration for my AWS Lambdas.
To have an idea my [await app.ready()](https://github.com/fastify/aws-lambda-fastify#lower-cold-start-latency) call went from approx. 400ms to approx. 1600ms _(on AWS Lambda this went up to 18000ms and also caused some timeouts)_ by updating from fastify [v4.1.0 to v4.2.0](https://github.com/fastify/fastify/compare/v4.1.0...v4.2.0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519).

After some testing I'm assuming that probably fast-json-stringify is causing this because of [this commit](https://github.com/fastify/fast-json-stringify/commit/a5e307987e3220dc93692cb7b0eca1f88c46da04).

I'm not so familiar with the internals, but this simple test, helped me to compare a bit more.

For example by wrapping the `refFinder(schema.$ref, location)` call here: https://github.com/fastify/fast-json-stringify/blob/v4.2.0/index.js#L108 in v4.2.0
i.e. like this:
```js
const d = Date.now()
location = refFinder(schema.$ref, location)
console.log('refFinder', Date.now() - d)
```
I basically get always 0 ms for the time measurements for all invocations, but compared to the newer `ajvInstance.getSchema(schemaRef)` call here: https://github.com/fastify/fast-json-stringify/blob/v5.0.0/index.js#L64 in v5.0.0
i.e. like this:
```js
const d = Date.now()
ajvSchema = ajvInstance.getSchema(schemaRef)
console.log('ajvInstance.getSchema', Date.now() - d)
```
I'm now getting 2-20ms for the time measurements for the different invocations.

image
image

Like said, I'm not familiar with these internals, but I assume [ajv.getSchema](https://ajv.js.org/api.html#ajv-getschema-key-string-undefined-data-any-boolean-promise-any) is getting the [fastify.ready()](https://www.fastify.io/docs/latest/Reference/Server/#ready) call much slower.

I'm sorry for my unprofessional reproduction steps, but I hope this information may help to understand the issue.

ivan-tymoshenko:
Fix #506 

I want to return a custom JSON schema reference resolver with some changes. It will only resolve JSON schema references only at serializer compilation time. All refs used for validation still will be resolved by Ajv. It's a proof of concept. I need to read the JSON schema doc carefully before merging it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -111,8 +111,14 @@ function build (schema, options) {
       }
 
       if (refResolver.getSchema(schemaKey) === undefined) {
-        ajvInstance.addSchema(externalSchema, schemaKey)
         refResolver.addSchema(externalSchema, key)
+      }
+
+      if (
+        ajvInstance.refs[schemaKey] === undefined &&
+        ajvInstance.schemas[schemaKey] === undefined
+      ) {
+        ajvInstance.addSchema(externalSchema, schemaKey)
       }
     }
   }
--- a/index.js
+++ b/index.js
@@ -9,6 +9,7 @@ const { randomUUID } = require('crypto')
 
 const validate = require('./schema-validator')
 const Serializer = require('./serializer')
+const RefResolver = require('./ref-resolver')
 const buildAjv = require('./ajv')
 
 let largeArraySize = 2e4
@@ -57,20 +58,12 @@ function resolveRef (location, ref) {
   const schemaId = ref.slice(0, hashIndex) || location.schemaId
   const jsonPointer = ref.slice(hashIndex) || '#'
 
-  const schemaRef = schemaId + jsonPointer
+  const schema = refResolver.getSchema(schemaId, jsonPointer)
 
-  let ajvSchema
-  try {
-    ajvSchema = ajvInstance.getSchema(schemaRef)
-  } catch (error) {
+  if (schema === undefined) {
     throw new Error(`Cannot find reference "${ref}"`)
   }
 
-  if (ajvSchema === undefined) {
-    throw new Error(`Cannot find reference "${ref}"`)
-  }
-
-  const schema = ajvSchema.schema
   if (schema.$ref !== undefined) {
     return resolveRef({ schema, schemaId, jsonPointer }, schema.$ref)
   }
@@ -83,6 +76,7 @@ const objectReferenceSerializersMap = new Map()
 
 let rootSchemaId = null
 let ajvInstance = null
+let refResolver = null
 let contextFunctions = null
 
 function build (schema, options) {
@@ -95,11 +89,13 @@ function build (schema, options) {
   options = options || {}
 
   ajvInstance = buildAjv(options.ajv)
+  refResolver = new RefResolver()
   rootSchemaId = schema.$id || randomUUID()
 
   isValidSchema(schema)
   extendDateTimeType(schema)
   ajvInstance.addSchema(schema, rootSchemaId)
+  refResolver.addSchema(schema, rootSchemaId)
 
   if (options.schema) {
     const externalSchemas = clone(options.schema)
@@ -114,8 +110,9 @@ function build (schema, options) {
         schemaKey = key + externalSchema.$id // relative URI
       }
 
-      if (ajvInstance.getSchema(schemaKey) === undefined) {
+      if (refResolver.getSchema(schemaKey) === undefined) {
         ajvInstance.addSchema(externalSchema, schemaKey)
+        refResolver.addSchema(externalSchema, key)
       }
     }
   }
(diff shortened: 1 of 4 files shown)


ivan-tymoshenko:
@adrai Can you confirm that these changes fix your case?

adrai:
Yes, seems much better, but still slower than before.

v4.1.0 156ms
v4.2.0 687ms
v4.2.0 with this fix 241ms

adrai:
fyi: https://github.com/adrai/fastify-slower-schema-resolving/pull/1#issuecomment-1221493117

ivan-tymoshenko:
> Yes, seems much better, but still slower than before.

Another optimization that I want to do is to add external schemas to the RefResolver only once instead of adding them for each route. But it would be a part of another PR. I want to move the fast-json-stringify-compiler to the fast-json-stringify.

I guess the problem is not in refResolver.getSchema. It should work as fast as refFinder worked. @adrai Can you check that, please?

adrai:
Do you mean this?

https://github.com/adrai/fastify-slower-schema-resolving/blob/main/simple-benchmarks/4.2.0/getSchema.output

https://github.com/adrai/fastify-slower-schema-resolving/blob/main/simple-benchmarks/4.1.0/findRef.output

mcollina (approved):
lgtm

index.js
```diff
@@ -114,8 +110,9 @@ function build (schema, options) {
         schemaKey = key + externalSchema.$id // relative URI
       }
 
-      if (ajvInstance.getSchema(schemaKey) === undefined) {
+      if (refResolver.getSchema(schemaKey) === undefined) {
```
Eomm, line 113: I would try to check the ajv's readonly properties.
This does not trigger the compiling and we could avoid the deep/equal logic

```
(x.schemas[schemaKey] || x.refs[schemaKey]) === undefined
```

```
const Ajv = require('ajv')

const x = new Ajv()

const externalSchema = {
  $id: 'root',
  definitions: {
    subschema: {
      $id: 'subschema',
      definitions: {
        anchorSchema: {
          $id: '#anchor',
          type: 'string'
        }
      }
    }
  }
}

x.addSchema(externalSchema)

{
  const id = 'subschema'
  console.log(x.schemas[id] || x.refs[id])
}

{
  const id = 'root'
  console.log(x.schemas[id] || x.refs[id])
}

{
  const id = 'subschema#anchor'
  console.log(x.schemas[id] || x.refs[id])
}
```

ivan-tymoshenko: The only concern I have is that refs and schemas is not a part of public API and can be changed without major release.

Eomm: Isn't it public?
https://github.com/ajv-validator/ajv/blob/master/lib/core.ts#L281

They would have declared as `private` instead

index.js
```diff
@@ -114,7 +110,14 @@ function build (schema, options) {
...
+        refResolver.addSchema(externalSchema, key)
+      }
+
+      if (
+        ajvInstance.refs[schemaKey] === undefined &&
+        ajvInstance.schemas[schemaKey] === undefined
+      ) {
```
Eomm, line 120: I meant that this check should be enough to replace the new `RefResolver` holder as both (ajv and refresolver) are storing the same values

I will try it locally

ivan-tymoshenko: It won't work for nested schemas like 'subschema' and 'subschema#anchor'. I will work if we want ti know if Ajv has a schema, but we can't get it.

ivan-tymoshenko: Oh, I see. Ajv resolves references in that cases.

Another reason why I want to have a separate abstraction for this is because Ajv does two things: validates schemas at runtime and resolves references at compile time. It's not good. So in #504 I will have to create two Ajv instances (for schema validation and reference resolving), because there would be two different syntaxes.

And since we own this code it's easy to optimize it. For example in case if we want to reuse external schemas.

ivan-tymoshenko: @Eomm I rebased #504 to this branch. You can see there that RefResolver/Validator have different responsibilities and different reasons to be changed.

mcollina:
@Eomm is this good to land?

Eomm (approved):
Lgtm

adrai:
Will this land in the next fastify version?

mcollina:
Ideally yes.

PR #511, opened 2022-08-25
Improve performance for const schemas

DanieleFedeli:
#### Checklist

As per issue https://github.com/fastify/fast-json-stringify/issues/502, when using a const schema, the performance degrades. This is the continuation of #505.

### Benchmark 
Pre PR
Screenshot 2022-08-16 alle 10 50 37

Post PR
Screenshot 2022-08-25 alle 18 46 08

This PR supports nullable values:
```javascript
test('schema with const string and no input', t => {
  t.plan(2)

  const schema = {
    type: 'object',
    properties: {
      foo: { const: 'bar' }
    }
  }

  const validate = validator(schema)
  const output = stringify({})

  t.equal(output, '{}')
})
```

Or it coerces the value to the const property if it is required
```javascript

test('schema with const string and no input but required property', t => {
  t.plan(2)

  const schema = {
    type: 'object',
    properties: {
      foo: { const: 'bar' }
    },
    required: ['foo']
  }

  const validate = validator(schema)
  const output = stringify({})

  t.equal(output, '{"foo":"bar"}')
})
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -240,6 +240,11 @@ function inferTypeByKeyword (schema) {
   for (var keyword of numberKeywords) {
     if (keyword in schema) return 'number'
   }
+
+  if (schema.const && typeof schema.const !== 'object') {
+    return typeof schema.const
+  }
+
   return schema.type
 }
 
--- a/index.js
+++ b/index.js
@@ -352,9 +352,6 @@ function buildCode (location) {
     // Using obj['key'] !== undefined instead of obj.hasOwnProperty(prop) for perf reasons,
     // see https://github.com/mcollina/fast-json-stringify/pull/3 for discussion.
 
-    const isRequired = schema.required !== undefined && schema.required.indexOf(key) !== -1
-    const isConst = schema.properties[key].const !== undefined
-
     code += `
       if (obj[${sanitized}] !== undefined) {
         ${addComma}
@@ -364,24 +361,14 @@ function buildCode (location) {
     code += buildValue(propertyLocation, `obj[${JSON.stringify(key)}]`)
 
     const defaultValue = schema.properties[key].default
-    const constValue = schema.properties[key].const
 
     if (defaultValue !== undefined) {
       code += `
       } else {
         ${addComma}
-        json += ${asString} + ':' + ${JSON.stringify(
-        JSON.stringify(defaultValue)
-      )}
+        json += ${asString} + ':' + ${JSON.stringify(JSON.stringify(defaultValue))}
       `
-    } else if (isRequired && isConst) {
-      code += `
-      } else {
-        json += ${asString} + ':' + ${JSON.stringify(
-        JSON.stringify(constValue)
-      )}
-      `
-    } else if (isRequired) {
+    } else if (required.includes(key)) {
       code += `
       } else {
         throw new Error('${sanitized} is required!')
@@ -810,7 +797,7 @@ function buildValue (location, input) {
   }
 
   let type = schema.type
-  const nullable = schema.nullable === true
+  const nullable = schema.nullable === true || (Array.isArray(type) && type.includes('null'))
 
   let code = ''
   let funcName
@@ -820,6 +807,12 @@ function buildValue (location, input) {
   }
 
   if ('const' in schema) {
+    if (nullable) {
+      code += `
+        json += ${input} === null ? 'null' : '${JSON.stringify(schema.const)}'
+      `
+      return code
+    }
     code += `json += '${JSON.stringify(schema.const)}'`
     return code
   }
@@ -1013,6 +1006,6 @@ module.exports.validLargeArrayMechanisms = validLargeArrayMechanisms
 module.exports.restore = function ({ code, ajv }) {
   const serializer = new Serializer()
   // eslint-disable-next-line
-  return Function.apply(null, ['ajv', 'serializer', code])
-    .apply(null, [ajv, serializer])
+  return (Function.apply(null, ['ajv', 'serializer', code])
+    .apply(null, [ajv, serializer]))
 }
--- a/index.js
+++ b/index.js
@@ -359,7 +359,7 @@ function buildCode (location) {
 
     const sanitized = JSON.stringify(key)
     const asString = JSON.stringify(sanitized)
-    const isRequired = schema.required !== undefined && schema.required.findIndex(item => item === key) !== -1
+    const isRequired = schema.required !== undefined && schema.required.indexOf(key) !== -1
     const isConst = schema.properties[key].const !== undefined
     // Using obj['key'] !== undefined instead of obj.hasOwnProperty(prop) for perf reasons,
     // see https://github.com/mcollina/fast-json-stringify/pull/3 for discussion.
--- a/index.js
+++ b/index.js
@@ -13,7 +13,10 @@ const buildAjv = require('./ajv')
 
 let largeArraySize = 2e4
 let largeArrayMechanism = 'default'
-const validLargeArrayMechanisms = ['default', 'json-stringify']
+const validLargeArrayMechanisms = [
+  'default',
+  'json-stringify'
+]
 
 const addComma = `
   if (addComma) {
@@ -31,9 +34,7 @@ function isValidSchema (schema, name) {
       name = ''
     }
     const first = validate.errors[0]
-    const err = new Error(
-      `${name}schema is invalid: data${first.instancePath} ${first.message}`
-    )
+    const err = new Error(`${name}schema is invalid: data${first.instancePath} ${first.message}`)
     err.errors = isValidSchema.errors
     throw err
   }
@@ -121,9 +122,7 @@ function build (schema, options) {
 
   if (options.rounding) {
     if (!['floor', 'ceil', 'round'].includes(options.rounding)) {
-      throw new Error(
-        `Unsupported integer rounding method ${options.rounding}`
-      )
+      throw new Error(`Unsupported integer rounding method ${options.rounding}`)
     }
   }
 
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -348,7 +359,8 @@ function buildCode (location) {
 
     const sanitized = JSON.stringify(key)
     const asString = JSON.stringify(sanitized)
-
+    const isRequired = schema.required !== undefined && schema.required.findIndex(item => item === key) !== -1
```
Uzlopak, line 362: can we do an indexOf?

index.js
```diff
@@ -805,159 +821,184 @@ function buildValue (location, input) {
...
-      } else if (schema.format === 'time') {
-        funcName = nullable ? 'serializer.asTimeNullable.bind(serializer)' : 'serializer.asTime.bind(serializer)'
-      } else {
-        funcName = buildObject(location)
+  if ('const' in schema) {
+    const stringifiedSchema = JSON.stringify(schema.const)
+
+    code += `
+            json += '${stringifiedSchema}'
+        `
```
ivan-tymoshenko, line 829: Can you just return the `code` from here and don't create `switchTypeSchema` nested function?

index.js
```diff
@@ -240,6 +240,17 @@ function inferTypeByKeyword (schema) {
   for (var keyword of numberKeywords) {
     if (keyword in schema) return 'number'
   }
+
+  if (schema.const) {
```
ivan-tymoshenko, line 244: I don't understand why we need to get a schema type if it's a const. Shouldn't we just substitute the value of the schema.const?

DanieleFedeli: You right, I forgot to remove that

mcollina (approved):
lgtm

index.js
ivan-tymoshenko: @DanieleFedeli Please don't close the conversation before resolving it.

index.js
```diff
@@ -805,159 +811,185 @@ function buildValue (location, input) {
...
-                ${statement}(${input} === null || typeof ${input} === "${type}" || ${input} instanceof RegExp || (typeof ${input} === "object" && Object.prototype.hasOwnProperty.call(${input}, "toString")))
+          const locationClone = clone(location)
+          sortedTypes.forEach((type, index) => {
+            const statement = index === 0 ? 'if' : 'else if'
+            locationClone.schema.type = type
+            const nestedResult = buildValue(locationClone, input)
+            switch (type) {
+              case 'string': {
+                code += `
+                ${statement}(${input} === null || typeof ${input} === "${type}" || ${input} instanceof RegExp || (typeof ${input} === "object" && Object.hasOwnProperty.call(${input}, "toString")))
```
ivan-tymoshenko, line 937: Please rebase your branch to the master before merging it. Otherwise, you will overwrite #510.

index.js
```diff
@@ -805,159 +811,185 @@ function buildValue (location, input) {
...
         `
-      } else if (isEmpty(schema)) {
-        code += `
+        } else if (isEmpty(schema)) {
+          code += `
           json += JSON.stringify(${input})
         `
-      } else if ('const' in schema) {
-        code += `
+        } else if ('const' in schema) {
```
ivan-tymoshenko, line 900: This should be removed.

index.js
```diff
@@ -360,14 +362,18 @@ function buildCode (location) {
...
+    }
+
     if (defaultValue !== undefined) {
       code += `
       } else {
         ${addComma}
         json += ${asString} + ':' + ${JSON.stringify(JSON.stringify(defaultValue))}
       `
-    } else if (required.includes(key)) {
+    } else if (required.includes(key) && !isConst) {
```
Uzlopak, line 376: maybe?

```suggestion
    } else if (isRequired && !isConst) {
```

index.js
```diff
@@ -360,14 +362,18 @@ function buildCode (location) {
 
     code += buildValue(propertyLocation, `obj[${JSON.stringify(key)}]`)
 
-    const defaultValue = schema.properties[key].default
+    let defaultValue = schema.properties[key].default
+    if (isRequired) {
+      defaultValue = defaultValue !== undefined ? defaultValue : schema.properties[key].const
```
ivan-tymoshenko, line 367: const shouldn't work as a default value. If you define a schema constant property, it must const value, whatever value the user passes.

ivan-tymoshenko (changes requested):

index.js
DanieleFedeli: Applied.

index.js
```diff
@@ -361,13 +370,24 @@ function buildCode (location) {
...
       code += `
       } else {
         ${addComma}
-        json += ${asString} + ':' + ${JSON.stringify(JSON.stringify(defaultValue))}
+        json += ${asString} + ':' + ${JSON.stringify(
+        JSON.stringify(defaultValue)
+      )}
       `
-    } else if (required.includes(key)) {
+    } else if (isRequired && isConst) {
```
ivan-tymoshenko, line 383: This should work in a different way.
1. You should handle value serialization only in one place otherwise you will have code duplication and can make an error. In your case, you miss the null check here. So please handle all schema serialization logic inside buildValue function.
2. We don't need to do this check for const https://github.com/fastify/fast-json-stringify/blob/844eba5f4956230e1ec64bfaea18d0e595d066d4/index.js#L365.

DanieleFedeli: Null checks you mean this?

```javascript
const schema = {
  "type": "object",
  "properties": {
    "foo": {
      "const": "bar"
    }
  }
}


stringify({ "foo": "bar" }) // { "foo": "bar" }
stringify({ "foo": "baz" }) // { "foo": "bar" }
stringify({ "foo": 1 }) // { "foo": "bar" }
stringify({}) // { }
```

DanieleFedeli: To answer 2. 
If I don't have that checks how can I handle this?

```javascript
const schema = {
  "type": "object",
  "properties": {
    "foo": {
      "const": "bar"
    }
  }
}

stringify({}) // { }
```

Uzlopak:
I read the code. Now it is like 80% reformatting. 19% tests and 1% change. And somehow i dont see any code which is responsible for comparing const with the actual value. I guess ajv does the validation? I cant follow, lol. How i read it, it is always serializing const no matter what the actual value is.

I am not that into the codebase so i guess i am out of here.

DanieleFedeli:
@Uzlopak Yes my bad, I don't know why I have the format changes since I don't even have the formats on save options.

ivan-tymoshenko:
@Uzlopak I think the main idea of https://github.com/fastify/fast-json-stringify/pull/505#pullrequestreview-1074429503 was to ignore the actual value and always insert the const value. I agree with this point of view because fjs is not a validator and if we can do something without validation we don't have to do validation. If you don't agree please share your thoughts.

ivan-tymoshenko:
@DanieleFedeli Null check.

```js
const schema = {
  "type": "object",
  "properties": {
    "foo": {
      "nullable": true
      "const": "bar"
    }
  }
}

stringify({foo:null}) // {foo:null}
```

```js
const schema = {
  "type": "object",
  "properties": {
    "foo": {
      "type": [...whatever, 'null']
      "const": "bar"
    }
  }
}

stringify({foo:null}) // {foo:null}
```

index.js
ivan-tymoshenko: I think that the const value shouldn't work as a default value. It shouldn't have any other additional behavior.

```js
const schema = {
  "type": "object",
  "properties": {
    "foo": {
      "const": "bar"
    }
  },
  "required": ['foo']
}

stringify({}) // {}
```

DanieleFedeli: I am OK with that, but I remember that the majority agree on this [comment](https://github.com/fastify/fast-json-stringify/pull/505#issuecomment-1217390075).

ivan-tymoshenko: @Eomm @climba03003 I have some thoughts on why const shouldn't duplicate the default value. These are two different responsibilities. const is one of the keys that determine what values a property can take aka type/set of value. It shouldn't have any other behavioral side effects.

I will give you an example. I can define an integer property with constraints min: 1 and max: 1. The only possible value is 1, but it doesn't mean that I should serialize this property if user doesn't pass the value.

ivan-tymoshenko:
@DanieleFedeli, Can you revert the formatting changes? It would be easier to review for other people.

DanieleFedeli:
I can try

index.js
DanieleFedeli: > I think that the const value shouldn't work as a default value. It shouldn't have any other additional behavior.
> 
> ```js
> const schema = {
>   "type": "object",
>   "properties": {
>     "foo": {
>       "const": "bar"
>     }
>   },
>   "required": ['foo']
> }
> 
> stringify({}) // {}
> ```

@ivan-tymoshenko Why that should not raise the error `"foo" is required!` ?

ivan-tymoshenko: Yes, you are right, it should. From my point of view.

mcollina: If you are waiting for me, I'm leaving this to you.

ivan-tymoshenko: I would like to hear @climba03003 opinion as it was his suggestion https://github.com/fastify/fast-json-stringify/pull/505#issuecomment-1217390075.

climba03003: When the schema said the property is required, and missing one from the input.
It just convenient to always put the value there since it is `required`.

Just like why we always put the `const` value regardless the input.

climba03003: I do not block if it turns to throw.
It consistence with the current `required` handling for other properties.

ivan-tymoshenko (approved):

mcollina (approved):
lgtm

mcollina:
@climba03003 are you ok for this to land?

climba03003 (approved):
Yes.
LGTM

PR #513, opened 2022-08-27
fix: toString method validation check

### LINKED ISSUE #492: `toString` existence check on objects does not work using Object.hasOwnProperty if the object is a class instance

ivan-tymoshenko:
Fix #492 

I think that this is a good opportunity to discuss this check. It says that if an object has a defined custom `toString` method, we should validate it as a string. There are two points why I think this is a bad check.

1. We don't coerce types during type validation because, otherwise, everything can be coerced to the string. Numbers and booleans also have the toString method, but we don't count them as a string type.

2. Ajv also does not use type coercion during schema validation, so the `toString` method won't be checked if you have anyOf/oneOf instead of a type array.

I want to drop this check in some major release of FJS/Fastify. I would like to heat your thoughts about it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -901,7 +901,17 @@ function buildValue (location, input) {
           switch (type) {
             case 'string': {
               code += `
-                ${statement}(${input} === null || typeof ${input} === "${type}" || ${input} instanceof RegExp || (typeof ${input} === "object" && Object.prototype.hasOwnProperty.call(${input}, "toString")))
+                ${statement}(
+                  typeof ${input} === "string" ||
+                  ${input} === null ||
+                  ${input} instanceof RegExp ||
+                  (
+                    typeof ${input} === "object" &&
+                    typeof ${input}.toString === "function" &&
+                    ${input}.toString !== Object.prototype.toString &&
+                    !(${input} instanceof Date)
+                  )
+                )
                   ${nestedResult}
               `
               break
--- a/test/typesArray.test.js
+++ b/test/typesArray.test.js
@@ -438,6 +438,35 @@ test('object that is simultaneously a string and a json switched', (t) => {
   t.equal(valueObj, '{"simultaneously":{"foo":"hello"}}')
 })
 
+test('class instance that is simultaneously a string and a json', (t) => {
+  t.plan(2)
+
+  const schema = {
+    type: 'object',
+    properties: {
+      simultaneously: {
+        type: ['string', 'object'],
+        properties: {
+          foo: { type: 'string' }
+        }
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
I think this can ship as a minor as it's not really breaking any tests. If we want to be a bit more cautious, we can ship a major of this and a major of our compiler wrapper so there is a way to revert to the original behavior for Fastify users.

ivan-tymoshenko:
> I think this can ship as a minor as it's not really breaking any tests. If we want to be a bit more cautious, we can ship a major of this and a major of our compiler wrapper so there is a way to revert to the original behavior for Fastify users.

I need to explain.
1. In this PR, I fixed the broken behavior.
2. My suggestion is to remove this behavior in the future, and it will break tests like that.
https://github.com/fastify/fast-json-stringify/blob/cd586bf065b45903e6bfa8508e7bba7c966b3f27/test/typesArray.test.js#L389

RafaelGSS (approved):

mcollina:
ah yes! Maybe open a PR and we'll land it in the future.

Issue #517, opened 2022-08-31
LGTM-Alert Prototype pollution?


Uzlopak:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

We should just check if this is a valid alert or not.

https://lgtm.com/projects/g/fastify/fast-json-stringify?mode=tree&ruleFocus=1513136283260

According to lgtm it was introduced with #504

mcollina:
It's not a problem: https://github.com/fastify/fast-json-stringify#security

climba03003:
Even the alert itself should be false positive.

`fjsCloned` is a symbol, not user provided string. It can never be `__proto__` and trigger the problem described.

PR #518, opened 2022-08-31
refactor: remove array items schema cloning

ivan-tymoshenko:
Even if #517 is not an issue, we don't need to clone and modify the schema here. It's enough to reassign the variable.


mcollina (approved):
lgtm

PR #519, opened 2022-09-01
refactor: remove empty schema handler

ivan-tymoshenko:
The empty schema handler is the same as the default handler.


mcollina (approved):
lgtm

PR #520, opened 2022-09-02
feat: drop nullable const support

ivan-tymoshenko:
❗BREAKING CHANGES

I didn't check it here #511 but turns out if the schema has const property, it can't be nullable.

If I have a schema like that: 
```json
{ "type": ["string", "null"], "const": "abc" }
```
or 
```json
{ "type": "string", "nullable": true, "const": "abc" }
```
it will validate the `null` value as false. That means if we serialize the `null` value as `null`, the serialized object would fail the validation.

P.S. I was confused because I saw the validation check, but it turns out that `is-my-json-valid` doesn't handle `const` check at all.
https://github.com/fastify/fast-json-stringify/blob/ab900c85f1dd047da1a7a94b5b0b69f782434fe1/test/const.test.js#L242

### DIFF
--- a/test/const.test.js
+++ b/test/const.test.js
@@ -238,7 +238,7 @@ test('schema with const and null as type', (t) => {
     foo: null
   })
 
-  t.equal(output, '{"foo":null}')
+  t.equal(output, '{"foo":"baz"}')
   t.ok(validate(JSON.parse(output)), 'valid schema')
 
   const output2 = stringify({ foo: 'baz' })
@@ -262,7 +262,7 @@ test('schema with const as nullable', (t) => {
     foo: null
   })
 
-  t.equal(output, '{"foo":null}')
+  t.equal(output, '{"foo":"baz"}')
   t.ok(validate(JSON.parse(output)), 'valid schema')
 
   const output2 = stringify({
--- a/index.js
+++ b/index.js
@@ -765,12 +765,6 @@ function buildValue (location, input) {
   let funcName
 
   if ('const' in schema) {
-    if (nullable) {
-      code += `
-        json += ${input} === null ? 'null' : '${JSON.stringify(schema.const)}'
-      `
-      return code
-    }
     code += `json += '${JSON.stringify(schema.const)}'`
     return code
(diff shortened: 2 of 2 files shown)


test/const.test.js
```diff
@@ -238,7 +238,7 @@ test('schema with const and null as type', (t) => {
     foo: null
   })
 
-  t.equal(output, '{"foo":null}')
+  t.equal(output, '{"foo":"baz"}')
```
Uzlopak, line 241: `JSON.stringify({foo: null })`  is `{"foo":null}`

So how can it be that stringifying `{ foo: null }` results in `{"foo": "baz} `

ivan-tymoshenko: FJS output should be equal to JSON.stringify only when you pass data that match the data schema. You shouldn't pass `null` as a data input for schema { "type": ["string", "null"], "const": "abc" }, because null doesn't match the schema.

ivan-tymoshenko: We want this to work:
```js
const schema = { /*some json schema draft 7*/ }
const data = { /*data that match to schema or can be coerced according to FJS rules */ }

const json = fjs(data)
const parsedData = JSON.parse(json)

validate(schema, data) // should be always true
```

And if we serialize `null` as `null`, 
```js
validate({ "type": ["string", "null"], "const": "abc" }, null)
```
will return false.

ivan-tymoshenko: For the same reason, if you pass `null` to the serializer with schema `{ "type": "number" }`, you will get `0`, instead of `null`.

ivan-tymoshenko:
We would have caught this error if we had checked the schema correctly. For example, Ajv does not allow the const and nullable keywords to be used in a schema at the same time. It will throw an error if you add schema with both properties. FJS validates input schemas with JSON schema draft 7 but uses an extended version of the specification. "nullable" keyword is not a part of JSON schema spec.

mcollina:
Unfortunately I think we should land this in a `next` branch and wait for the Fastify v5 release train (2023).

PR #523, opened 2022-09-05
refactor: handle nullable schema in one place

ivan-tymoshenko:
I'm sorry that there are so many changes. Most of them are decomposition into functions without changes. This PR does two things:
1. Handles nullable property in one place instead of checking it all over the code.
2. Optimize handling multitype schemas. 

I can't separate these two changes. They are related to each other.


mcollina (approved):
lgtm

Eomm (approved):

PR #524, opened 2022-09-05
fix: remove array items type checking

ivan-tymoshenko:
Type checking in this place is not only unnecessary (FJS is not a validator), it blocks type coercion for array items for no reason.

### DIFF
--- a/test/array.test.js
+++ b/test/array.test.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const test = require('tap').test
+const { DateTime } = require('luxon')
 const validator = require('is-my-json-valid')
 const build = require('..')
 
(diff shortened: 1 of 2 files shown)


test/array.test.js
```diff
@@ -152,28 +153,56 @@ buildTest({
   '@data': ['test']
 })
 
-test('invalid items throw', (t) => {
+test('coerce number to string type item', (t) => {
```
Eomm, line 156: Is `patternProperties` still a thing?

ivan-tymoshenko: I don't know why there is a patternProperties in this test. It works without them ('invalid' string always doesn't match the object schema). The test will work differently after this PR, but not because of patternProperties.

How the test works now:
```js
test('invalid items throw', (t) => {
  t.plan(1)
  const schema = {
    type: 'object',
    properties: {
      args: {
        type: 'array',
        items: [
          {
            type: 'object',
            patternProperties: {
              '.*': {
                type: 'string'
              }
            }
          }
        ]
      }
    }
  }
  t.throws(() => stringify({ args: ['invalid'] })) // throws an error because 'invalid' is not an object
})
```

How it will work:
```js
test('invalid items throw', (t) => {
  t.plan(1)
  const schema = {
    type: 'object',
    properties: {
      args: {
        type: 'array',
        items: [
          {
            type: 'object',
            patternProperties: {
              '.*': {
                type: 'string'
              }
            }
          }
        ]
      }
    }
  }
  t.equal(stringify({ args: ['invalid'] }), '{"args":[{"0":"i","1":"n","2":"v","3":"a","4":"l","5":"i","6":"d"}]}') // it coerces 'invalid' string to the object schema
})
```

It will work just like the same schema with patternProperties works now without array.
```js
test('invalid items throw', (t) => {
  t.plan(1)
  const schema = {
    type: 'object',
    patternProperties: {
      '.*': {
        type: 'string'
      }
    }
  }
  t.equal(stringify('invalid'), '{"0":"i","1":"n","2":"v","3":"a","4":"l","5":"i","6":"d"}')
})

```

ivan-tymoshenko: @Eomm  I decided to remove the test instead of updating it because this coercion looks like nonsense. If you want I can add the test like that.

climba03003:
I have merged `master` to `next` branch.
This PR should be target to `next`.

test/array.test.js
Eomm: I like the idea of removing all the "validation" logic here, but I would move it to the `next` branch.

I think the new output could warn a user that does not understand what is going on: I would prefer an empty output instead of `{"0":"i","1":"n","2":"v","3":"a","4":"l","5":"i","6":"d"}`, but we can't do it without keeping the "validation" logic tho

mcollina:
I'm ok for landing this in the next branch.

test/array.test.js
ivan-tymoshenko: The output you see isn't related to the array at all. It's all about how FJS coerces string to object.

ivan-tymoshenko:
For me, it's more like a fix, but if you all think it should be merged into the next, I'm ok with it. I want to clarify, why did we merge #511 into the master branch instead of the next branch? It has the same type of changes.

mcollina:
Ah ok! Then I'm ok with a fix/

climba03003:
> For me, it's more like a fix, but if you all think it should be merged into the next, I'm ok with it. I want to clarify, why did we merge #511 into the master branch instead of the next branch? It has the same type of changes.

It is hard to tell.
For me, if the PR is adding more support and do not break the current test (test change due to the support is acceptable). It should be safe to land in `master`.

And here, we know that it will certainly break most of the people. It is more appropriate to wait for a major bump, not one PR for one major bump.
For example, #520, #504 (update on debug mode) can be group inside `next` branch.

PR #525, opened 2022-09-06
fix: check all types for toJSON function

ivan-tymoshenko:
@bmenant I've taken your tests from #414 (except two). Can you take a look at this PR?

All meaningful changes are here: https://github.com/fastify/fast-json-stringify/compare/add-tojson-function-handler?expand=1#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R877. All other is necessary refactoring.

Снимок экрана 2022-09-06 в 13 59 15

### DIFF
--- a/test/toJSON.test.js
+++ b/test/toJSON.test.js
@@ -172,3 +172,1020 @@ test('throw an error on non nullable null object', (t) => {
   })
   t.throws(() => { stringify(null) })
 })
+
+test('recursive toJSON call', (t) => {
+  t.plan(1)
+
+  const schema = {
+    type: 'object',
+    properties: {
+      firstName: {
+        type: 'string'
+      },
+      lastName: {
+        type: ['string', 'null']
+      },
+      age: {
+        type: 'integer',
+        minimum: 0
+      }
+    }
+  }
+
+  const stringify = build(schema)
+
+  const data = {
+    props: {
+      firstName: { toJSON () { return 'Matteo' } },
+      lastName: { toJSON () { return 'Collina' } },
+      age: { toJSON () { return 32 } }
+    },
+    toJSON () {
+      return {
+        firstName: this.props.firstName,
+        lastName: this.props.lastName,
+        age: this.props.age
+      }
+    }
+  }
+
+  t.equal(stringify(data), JSON.stringify(data))
+})
+
+test('toJSON - ref - properties', (t) => {
+  t.plan(2)
+
+  const schema = {
+    title: 'object with $ref',
+    definitions: {
+      def: {
+        type: 'object',
+        properties: {
+          str: {
+            type: 'string'
+          }
+        }
+      }
+    },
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

test/toJSON.test.js
```diff
@@ -172,3 +172,1020 @@ test('throw an error on non nullable null object', (t) => {
...
+          type: 'number'
+        }
+      }
+    }
+  }
+
+  const dataNull = { toJSON () { return null } }
+
+  const stringify = build(schema)
+  t.throws(() => stringify({ arr: dataNull }))
```
Eomm, line 890: Which error?

test/toJSON.test.js
```diff
@@ -172,3 +172,1020 @@ test('throw an error on non nullable null object', (t) => {
...
+    toJSON () {
+      return { obj: this.props.obj }
+    }
+  }
+
+  const stringify = build(schema)
+  const output = stringify(object)
+
+  JSON.parse(output)
+  t.pass()
```
Eomm, line 253: I would move this line into the `toJSON` method, or it seems not informative here

test/toJSON.test.js
ivan-tymoshenko: 'The value \'null\' does not match schema definition.'

ivan-tymoshenko:
This feature will cause some performance downgrade. I can add the initial parameter that would turn it off. Should I add it?

climba03003:
> I can add the initial parameter that would turn it off. Should I add it?

Around 5% performance, I think it better turn off by default. Or at least make it configurable.
It can be done on another PR if you like.

climba03003 (approved):
LGTM.

Eomm (commented):
> I think it better turn off by default

Does it not make it braking change setting the option to `false` by default?
Do we want to ship it in the current major?

ivan-tymoshenko:
> > I think it better turn off by default
> 
> 
> 
> Does it not make it braking change setting the option to `false` by default?
> 
> Do we want to ship it in the current major?
> 
> 

Yes, it will be a breaking change if we turn it off by default. That's why I changed the target branch to the next.

I thought that the next major release for the FJS would be possible only with a Fastify major release. Then I don’t see a problem with shipping it. It I’m not right, please explain it to me.

Eomm (approved):
> Yes, it will be a breaking change if we turn it off by default. That's why I changed the target branch to the next.

Oh, I oversight that!

> I thought that the next major release for the FJS would be possible only with a Fastify major release. Then I don’t see a problem with shipping it. It I’m not right, please explain it to me.

Ni, we can ship a new major for `@fastify/fast-json-schema-serializer`, and then users can use it very easily! So it is not necessary to wait for fastify v5 😄

I'm quite proud of that component, even if it is quite tricky sometimes

PR #526, opened 2022-09-06
chore: move top level modules into folders

ivan-tymoshenko:
There are too many files in the root folder. I moved them into three new folders: lib, build, and examples.


Uzlopak (approved):

build/build-schema-validator.js
```diff
@@ -23,4 +23,4 @@ const validationCode = standaloneCode(ajv, validate)
 
 const moduleCode = `/* CODE GENERATED BY '${path.basename(__filename)}' DO NOT EDIT! */\n${validationCode}`
 
-fs.writeFileSync(path.join(__dirname, 'schema-validator.js'), moduleCode)
+fs.writeFileSync(path.join(__dirname, './../lib/schema-validator.js'), moduleCode)
```
climba03003, line 26:
```suggestion
fs.writeFileSync(path.join(__dirname, '../lib/schema-validator.js'), moduleCode)
```

climba03003 (approved):

mcollina (approved):
lgtm

examples/example.js
```diff
@@ -1,4 +1,4 @@
-const fastJson = require('.')
+const fastJson = require('..')
```
Eomm, line 1: this file is linked in the package.json to support runkit
```suggestion
const fastJson = require('fast-json-stringify')
```

Uzlopak: Then we have to create one for runkit

ivan-tymoshenko: @Eomm  I updated path in require call and path to example.js in the package.json.

Eomm (approved):

PR #527, opened 2022-09-07
refactor: move global variables into class context

ivan-tymoshenko:
Through the building serializer process, different function use the same context that is stored in global. I think it would be better to share this context using the class.

Also, I separated the `build` function call into two different calls: 
1. Class constructor initiates all context variables, including external schemas.
2. The `build` method: should build the schema serializer.
You can't call the `build` method of the SerializerFactory with different schemas yet. I want to add this functionality, but in another PR.

The API of the library is still the same.

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,13 +6,12 @@ const merge = require('@fastify/deepmerge')()
 const clone = require('rfdc')({ proto: true })
 const { randomUUID } = require('crypto')
 
+const buildStandaloneCode = require('./lib/standalone')
 const validate = require('./lib/schema-validator')
 const Serializer = require('./lib/serializer')
 const Validator = require('./lib/validator')
 const RefResolver = require('./lib/ref-resolver')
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

serializer.js
```diff
@@ -12,9 +12,11 @@ module.exports = class Serializer {
       case 'round':
         this.parseInteger = Math.round
         break
-      default:
+      case undefined:
         this.parseInteger = Math.trunc
         break
+      default:
```
climba03003, line 18: I would split this changes to another PR.

climba03003 (changes requested):

serializer.js
climba03003: Nevermind, I see where it comes from.

climba03003 (approved):
LGTM.

mcollina:
Can you rebase? This now conflicts.

PR #533, opened 2022-09-10
fix: apply if/then/else check for any schema

ivan-tymoshenko:
We have to check for if/then/else all schemas. Not only objects.

### DIFF
--- a/test/if-then-else.test.js
+++ b/test/if-then-else.test.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const t = require('tap')
+const { DateTime } = require('luxon')
 const build = require('..')
 
 const schema = {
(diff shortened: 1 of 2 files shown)


test/if-then-else.test.js
```diff
@@ -368,3 +369,52 @@ t.test('nested if/then', t => {
...
+    type: 'integer',
+    if: { minimum: 42 },
+    then: { const: 66 },
+    else: { const: 33 }
+  }
+
+  const stringify = build(schema)
+
+  t.equal(stringify(66.32), '66')
+  t.equal(stringify(33.12), '33')
```
Eomm, line 403: just to be sure that the rounding is not triggered

```suggestion
  t.equal(stringify(100.32), '66')
  t.equal(stringify(10.12), '33')
```

mcollina (approved):
lgtm

Eomm (approved):

Issue #508, opened 2022-08-21
Move the fast-json-stringify-compiler to the fast-json-stringify


ivan-tymoshenko:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

I want to move the fast-json-stringify-compiler logic to the fast-json-stringify.

Motivation: The fast-json-stringify-compiler binds external schemas to the functional context. What we can do is precompile as many things as we can: clone schemas, add schemas to the RefResolver, modify schemas for Ajv and maybe precompile serializers for those external schemas that don't depend on route schema. Now we do these operations for all external schemas each time when we build a serializer for a route. All these optimizations should be done inside the library because they are based on the implementation and not on the interface.

@Eomm @mcollina WDYT?

Eomm:
The main topic is that fjs has always provided a simple function that does all the work.

To do what you propose we need to move to an interface like ajv (creating an instance calling new)

I think that the fjs simplicity is one of its key element. So I would keep it while planning the its future steps

ivan-tymoshenko:
We can keep both interfaces.

mcollina:
fast-json-stringify can be used outside of Fastify. fast-json-stringify-compiler is fastify-specific. I'd prefer to keep the two separate unless we need them not to be for a specific reason.

ivan-tymoshenko:
@Eomm I want to discuss with you this ajv-compilet issue https://github.com/fastify/ajv-compiler/issues/81, but according to FJS. You said that creating an ajv container for each route is overkilling. We actually do it for FJS. I want to change it, but the trick with [addUsedSchema](https://ajv.js.org/options.html#addusedschema) wouldn't work in FJS because we need to validate different subschemas instead of the schema itself. Do you think it is possible to add a constraint not to create conflicting global links? We will throw an error if you try to set up two different routes that have different schemas with the same global id.

Tony133:
Closing as stale — no activity since 2022.

For the record, the position expressed in this thread was to keep `fast-json-stringify` a simple single-function API and leave the Fastify-specific caching to `fast-json-stringify-compiler`.

If there's still interest in this, feel free to open a new Discussion.

PR #538, opened 2022-09-12
perf: add only required schemas to the validator

ivan-tymoshenko:
This will optimize a cold start for those schemas that have validation parts (anyOf, oneOf, if/then) and others external schemas in the context. 
\+ It might help with resolving schema conflicts in the future for #508.

### DIFF
--- a/index.js
+++ b/index.js
@@ -44,7 +44,8 @@ function mergeLocation (location, key) {
   return {
     schema: location.schema[key],
     schemaId: location.schemaId,
-    jsonPointer: location.jsonPointer + '/' + key
+    jsonPointer: location.jsonPointer + '/' + key,
+    isValidated: location.isValidated
   }
 }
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -138,12 +141,12 @@ function build (schema, options) {
...
-    const externalSchemas = options.schema || {}
-    const validatorSchemas = { ...externalSchemas, ...mergeAllOfSchemas }
-    for (const [schemaKey, schema] of Object.entries(validatorSchemas)) {
-      validator.addSchema(schema, schemaKey)
+  for (const schemaId of validatorSchemasIds) {
+    const schema = refResolver.getSchema(schemaId)
+    if (schema) {
+      validator.addSchema(schema, schemaId)
+    } else {
+      throw new Error(`Cannot find schema "${schemaId}"`)
```
Eomm, line 149: This should never happen, right?

ivan-tymoshenko: Yes, add added it just in case) I would be easier to catch.

Uzlopak: Can we have a test for the else case please?

image

ivan-tymoshenko: This else branch should never happen. I don't know the case that would throw an error. I can remove that if/else, but it would be much harder to catch the error if something goes wrong or if I missed something.

Uzlopak: Let me check it.

ivan-tymoshenko: I think you catch another error.
https://github.com/fastify/fast-json-stringify/blob/68f7b695b2804778d5aba38a31676828db5ac845/index.js#L63

I should change the message of the new error to something different.

mcollina (approved):
lgtm

index.js
Uzlopak: I know. It caches another Error. Thats why I wrote that if we refactor and dont catch it with the refresolver we can catch it anyway.

Actually we catch the missing reference already in resolveRef. So we can remove the if clause.

ivan-tymoshenko: @Uzlopak done

Eomm (approved):

PR #542, opened 2022-09-13
fix ajv strict warnings in if/else-tests

Uzlopak:
#### Checklist

### DIFF
--- a/test/if-then-else.test.js
+++ b/test/if-then-else.test.js
@@ -391,8 +391,8 @@ t.test('if/else with const integers', (t) => {
   t.plan(2)
 
   const schema = {
-    type: 'integer',
-    if: { minimum: 42 },
+    type: 'number',
+    if: { type: 'number', minimum: 42 },
     then: { const: 66 },
     else: { const: 33 }
   }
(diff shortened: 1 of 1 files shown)


test/if-then-else.test.js
```diff
@@ -391,8 +391,8 @@ t.test('if/else with const integers', (t) => {
   t.plan(2)
 
   const schema = {
-    type: 'integer',
```
climba03003, line 394: `integer` should be a valid `type`. Why change it to `number`?

ttps://json-schema.org/understanding-json-schema/reference/numeric.html#id4

Uzlopak: test/if-then-else.test.js 2> strict mode: type "number" not allowed by context "integer" at "9f255981-9e81-41ed-8a47-ec359679ccd7#/if" (strictTypes)

Uzlopak: @climba03003 
Does this answer your question?

climba03003: It does answer my question, but I would like to hear why it is added like this in the beginning.
This test is added in https://github.com/fastify/fast-json-stringify/pull/533

cc @ivan-tymoshenko

ivan-tymoshenko: > I would like to hear why it is added like this in the beginning

It made some sense before this change. https://github.com/fastify/fast-json-stringify/pull/533#discussion_r967678324

It's ok to change it now.

ivan-tymoshenko (approved):

Eomm (approved):

PR #546, opened 2022-09-18
fix: merged schemas reference resolving

### LINKED ISSUE #540: Cannot find reference with allOf
### Prerequisites

### Fastify version

4.5.3

### Plugin version

5.3.0

### Node.js version

16

### Operating system

macOS

12.2

### Description

I suspect it comes from https://github.com/fastify/fastify/pull/4049 shipped with fastify 4.2.0 but I cannot figure what is wrong with my case.
It used to work with  fastify 4.1.0 .

Maybe you have a clue.

### Steps to Reproduce

Schema is :
```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "Test",
  "definitions": {
    "Problem": {
      "type": "object",
      "required": [
        "type"
      ],
      "additionalProperties": true,
      "properties": {
        "type": {
          "type": "string"
        }
      }
    },
    "ValidationFragment": {
      "type": "string"
    },
    "ValidationErrorProblem": {
      "type": "object",
      "allOf": [
        {
          "$ref": "#/definitions/Problem"
        },
        {
          "type": "object",
          "properties": {
            "validation": {
              "$ref": "#/definitions/ValidationFragment"
            }
          }
        }
      ]
    }
  },
  "additionalProperties": false,
  "type": "object"
}
```
Code is
```ts
import fastify from 'fastify';

const server = fastify();

const schema = require('./schema.json');
server.addSchema(schema);

server.post<{
}>(
  '/dummy',
  {
    schema: {
      response: {
        '400': {$ref: 'Test#/definitions/ValidationErrorProblem'},
      },
    }
  },
  async (_request, _reply) => {
    return 'OK'
  });

server.ready();
```
Error is :
```
node_modules/fastify/lib/route.js:326
              throw new FST_ERR_SCH_SERIALIZATION_BUILD(opts.method, url, error.message)
                    ^
FastifyError [Error]: Failed building the serialization schema for POST: /dummy, due to error Cannot find reference "#/definitions/ValidationFragment"
```

### Expected Behavior

It used to work with fastify 4.1.0. I would expect FJS to compile the schema.

ivan-tymoshenko:
Fix #540

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ const validate = require('./lib/schema-validator')
 const Serializer = require('./lib/serializer')
 const Validator = require('./lib/validator')
 const RefResolver = require('./lib/ref-resolver')
+const Location = require('./lib/location')
 
 let largeArraySize = 2e4
 let largeArrayMechanism = 'default'
@@ -40,57 +41,13 @@ function isValidSchema (schema, name) {
   }
 }
 
-class Location {
-  constructor (schema, schemaId, jsonPointer = '#', isValidated = false) {
-    this.schema = schema
-    this.schemaId = schemaId
-    this.jsonPointer = jsonPointer
-    this.isValidated = isValidated
-
-    this.isMerged = false
-    this.mergedSchemaId = null
-    this.mergedJsonPointer = null
-  }
-
-  getPropertyLocation (propertyName) {
-    const propertyLocation = new Location(
-      this.schema[propertyName],
-      this.schemaId,
-      this.jsonPointer + '/' + propertyName,
-      this.isValidated
-    )
-
-    if (this.isMerged) {
-      propertyLocation.addMergedSchema(
-        this.mergedSchemaId,
-        this.mergedJsonPointer + '/' + propertyName
-      )
-    }
-
-    return propertyLocation
-  }
-
-  getSchemaRef () {
-    if (this.isMerged) {
-      return this.mergedSchemaId + this.mergedJsonPointer
-    }
-    return this.schemaId + this.jsonPointer
-  }
-
-  addMergedSchema (schemaId, jsonPointer = '#') {
-    this.isMerged = true
-    this.mergedSchemaId = schemaId
-    this.mergedJsonPointer = jsonPointer
-  }
-}
-
 function resolveRef (location, ref) {
   let hashIndex = ref.indexOf('#')
   if (hashIndex === -1) {
     hashIndex = ref.length
   }
 
-  const schemaId = ref.slice(0, hashIndex) || location.schemaId
+  const schemaId = ref.slice(0, hashIndex) || location.getOriginSchemaId()
   const jsonPointer = ref.slice(hashIndex) || '#'
 
   const schema = refResolver.getSchema(schemaId, jsonPointer)
@@ -299,7 +256,7 @@ function buildExtraObjectPropertiesSerializer (location) {
       try {
         RegExp(propertyKey)
       } catch (err) {
-        const jsonPointer = propertyLocation.schema + propertyLocation.jsonPointer
+        const jsonPointer = propertyLocation.getSchemaRef()
         throw new Error(`${err.message}. Invalid pattern property regexp key ${propertyKey} at ${jsonPointer}`)
       }
 
@@ -349,9 +306,6 @@ function buildInnerObject (location) {
 
   let code = ''
 
-  if (!(location instanceof Location)) {
-    console.log(location)
-  }
   const propertiesLocation = location.getPropertyLocation('properties')
   Object.keys(schema.properties || {}).forEach((key) => {
     let propertyLocation = propertiesLocation.getPropertyLocation(key)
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -40,12 +40,47 @@ function isValidSchema (schema, name) {
   }
 }
 
-function mergeLocation (location, key) {
-  return {
-    schema: location.schema[key],
-    schemaId: location.schemaId,
-    jsonPointer: location.jsonPointer + '/' + key,
-    isValidated: location.isValidated
+class Location {
```
Eomm, line 43: Did you evaluate the `URL` class?

ivan-tymoshenko: Sorry, I didn't get you.

test/allof.test.js
```diff
@@ -437,3 +437,48 @@ test('object with external $refs in allOf', (t) => {
...
+          type: 'object',
+          allOf: [
+            {
+              $ref: '#/definitions/Problem'
+            },
+            {
+              type: 'object',
+              properties: {
+                validation: {
+                  $ref: '#/definitions/ValidationFragment'
```
Eomm, line 469: what about an absolute ref like `schema:app:foo#/definitions/hello`?

index.js
Eomm: The [URL](https://nodejs.org/api/url.html#class-url) class has many convenient methods to calculate if the URL is absolute and relative and compose it.

Looking at this `Location` class, I have afraid that we are going to replicate the `URL` class in future.

So I wanted to point it out to reduce the burden you could face by developing it

test/allof.test.js
ivan-tymoshenko: Yes, they are working. This PR doesn't change anything for absolute refs. The problem was that when we create merged schema we should have two locations to resolve local refs:
1. We need location inside new "merged" schema in case if we have anyof/oneof .., inside merged schema.
2. We original location to be able to resolve local reference in this schema, but outside of merged schema.

It's hard to explain simply) #480 Fixed the first case, but broke the second. This PR should support both.

ivan-tymoshenko:
@Eomm I can't replace the hole Location class with an URL, we've got some custom logic here. I can use URL instead of split/join for schemaId and jsonPointer, but I'm not sure that it's a good idea. All operation that we've got pretty simple and URL is super slow in compare with them.
I can't replace the hole Location class with an URL, we've got some custom logic here. I can use URL instead of split/join for schemaId and jsonPointer, but I'm not sure that it's a good idea. All operation that we've got pretty simple and URL is super slow in compare with them.

index.js
```diff
@@ -40,12 +40,47 @@ function isValidSchema (schema, name) {
...
+  constructor (schema, schemaId, jsonPointer = '#', isValidated = false) {
+    this.schema = schema
+    this.schemaId = schemaId
+    this.jsonPointer = jsonPointer
+    this.isValidated = isValidated
+
+    this.isMerged = false
+    this.mergedSchemaId = null
+    this.mergedJsonPointer = null
```
climba03003, line 52: It is currently hard to follow why does it fix the issue and how it is fixed.
Would you mind to add some comment of step flow why it works? and how the usage of `mergedJsonPointer`?

index.js
climba03003: I don't think they are the same and `URL` is really slow.
The usage here are only `string` concatenation.

index.js
```diff
@@ -40,12 +40,47 @@ function isValidSchema (schema, name) {
...
+      return this.mergedSchemaId + this.mergedJsonPointer
+    }
+    return this.schemaId + this.jsonPointer
+  }
+
+  addMergedSchema (schemaId, jsonPointer = '#') {
+    this.isMerged = true
+    this.mergedSchemaId = schemaId
+    this.mergedJsonPointer = jsonPointer
   }
```
mcollina, line 84: Can you move this class into its own file under lib?

index.js
ivan-tymoshenko: @climba03003 @Eomm  I will explain with an example. Sorry, it will not be simple because the case is not simple.

```js
{
  $id: 'root',
  type: 'object',
  properties: {
    property1: { type: 'string' },
    property2: {
      type: 'object',
      allOf: [
        {
          properties: {
            property21: { $ref: '#/properties/property1' }
          }
        },
        {
          properties: {
            property22: {
              anyOf: [
                { type: 'string' },
                { type: 'number' }
              ]
            }
          }
        }
      ]
    }
  }
}
```

1. When we meet property `allOf`, we create a merged schema and create a new unique id for it. 
https://github.com/fastify/fast-json-stringify/blob/e6b3c5fb6e1dc47f1c71560adf2ef3363bbf2b70/index.js#L458
If we meet anyOf/oneOf/if inside the merged schema, we create a reference using the merged schema id.
`property22` will have the next reference: `merged_12...223#/properties/property22`. To create this reference, we should know the merged schema id.

2. If somewhere inside the `allOf` schemas we meet a local reference like  `'#/properties/property1'` we have to resolve it. To do this, we should know the original schema reference, which is `root` in this case.

So that is why we need to keep two schemas ids when we build a serializer inside the `allOf` property.

ivan-tymoshenko: The `Location` class now has two methods for getting schema id.
1. `getSchemaId` - you should use it when you want to CREATE a new reference to some place in the schema. For example, for validation.
2. `getOriginSchemaId` - you should use it when you want to RESOLVE already existing references.

I added corresponding comments in the code.

climba03003 (approved):
LGTM. 
Thanks for the explanation, it's now more clear to me.

remidewitte:
Thanks. I will test asap :)

PR #547, opened 2022-09-18
chore: add linting check to npm test command

ivan-tymoshenko:


Uzlopak (changes requested):
We removed linting from tests on purpose. What actually has to be done is adding

```
"pre-commit": [
  "lint",
  "test"
]
```

to package.json so that pre-commit runs linting before checking in.

Uzlopak:
@Fdawgs

ivan-tymoshenko:
IMHO it's just inconvenient. When I have written some code and want to check that everything is fine, I will have to run two commands instead of one.

Uzlopak:
I get your point.  but lets discuss this with Fdawgs.

mcollina (approved):
lgtm

Uzlopak (approved):

PR #553, opened 2022-10-15
perf: use regexp in string escaping

ivan-tymoshenko:
Do you want some perf upgrades? I've got something for you.

### DIFF
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -1,5 +1,8 @@
 'use strict'
 
+// eslint-disable-next-line
+const STR_ESCAPE = /[\u0000-\u001f\u0022\u005c\ud800-\udfff]|[\ud800-\udbff](?![\udc00-\udfff])|(?:[^\ud800-\udbff]|^)[\udc00-\udfff]/
+
 module.exports = class Serializer {
   constructor (options = {}) {
     switch (options.rounding) {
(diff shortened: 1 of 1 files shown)


ivan-tymoshenko:
Снимок экрана 2022-10-15 в 15 22 11
Снимок экрана 2022-10-15 в 15 22 24

lib/serializer.js
```diff
@@ -94,6 +97,10 @@ module.exports = class Serializer {
       str = str.toString()
     }
 
+    if (!STR_ESCAPE.test(str)) {
```
Uzlopak, line 100: I was wondering, if following has a perf impact?
```suggestion
    if (STR_ESCAPE.test(str) === false) {
```

Could you please bench it?

lib/serializer.js
```diff
@@ -1,5 +1,8 @@
 'use strict'
 
+// eslint-disable-next-line
+const STR_ESCAPE = /[\u0000-\u001f\u0022\u005c\ud800-\udfff]|[\ud800-\udbff](?![\udc00-\udfff])|(?:[^\ud800-\udbff]|^)[\udc00-\udfff]/
```
Uzlopak, line 4: Did you check with safe-regex2 package if it is a safe regex?

lib/serializer.js
RafaelGSS: It doesn't.

RafaelGSS, line 100: Could you please include a comment above this line telling why it's needed?

lib/serializer.js
ivan-tymoshenko: It's safe.

lib/serializer.js
Uzlopak: Thx

mcollina (approved):
lgtm

climba03003 (approved):
LGTM.

PR #558, opened 2022-11-07
fast-json-stringify is not a validator

mcollina:
Let's make it clear that fast-json-stringify is not a validator and the caller takes all the responsibilities on the safety of the data being rendered.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -650,6 +650,9 @@ While the `schema` is currently validated for any developer errors,
 there is no guarantee that supplying user-generated schema could not
 expose your application to remote attacks.
 
+The _data_ that `fast-json-stringify` serializes is trusted too. In other terms,
+fast-json-stringify makes no claim to validate the incoming data against the schema. 
+
 <a name="debug"></a>
 ### Debug Mode
 
--- a/README.md
+++ b/README.md
@@ -650,8 +650,9 @@ While the `schema` is currently validated for any developer errors,
 there is no guarantee that supplying user-generated schema could not
 expose your application to remote attacks.
 
-The _data_ that `fast-json-stringify` serializes is trusted too. In other terms,
-fast-json-stringify makes no claim to validate the incoming data against the schema. 
+Users are responsibile for sending trusted data. `fast-json-stringify` guarantees that you will get
+a valid output only if your input matches the schema or can be coerced to the schema. If your input
+doesn't match the schema, you will get undefined behavior. 
 
 <a name="debug"></a>
 ### Debug Mode


ivan-tymoshenko:
Does it mean that we can drop all unnecessary validation checks in the next fastify major release? Because if not, then it's better to specify which validation check we do.

mcollina:
It means that we offer no guarantees we will do any checks. If we want to do checks it's up to us.

ivan-tymoshenko:
Why do we want to make some validation checks that are not used in serialization (they are just validation checks) and don't tell users about them?

ivan-tymoshenko:
I want to say that validation in the fast-json-stringify is the main and one of the most difficult questions for me right now. We have a [thread](https://github.com/fastify/fast-json-stringify/discussions/532#discussioncomment-3609675) where we discuss how it should work in the next major.

I agree that FJS doesn't give you any guarantees on the output, and I'm ok with such a comment. But the question is much deeper, and if we say that "fast-json-stringify makes no claim to validate the incoming data against the schema" it's a little bit strange for me that we throw an error if your data doesn't match the schema in some cases.

README.md
```diff
@@ -650,6 +650,9 @@ While the `schema` is currently validated for any developer errors,
 there is no guarantee that supplying user-generated schema could not
 expose your application to remote attacks.
 
+The _data_ that `fast-json-stringify` serializes is trusted too. In other terms,
+fast-json-stringify makes no claim to validate the incoming data against the schema.
```
climba03003, line 654:
```
The _data_ that `fast-json-stringify` serializes should be a trusted source. 
The user should take their responsibility to sanitize the _data_ before passing 
to `fast-json-stringify`. The validation in this package is only against the `schema` 
but not `malicious code`.
```

climba03003: Would it be better? @ivan-tymoshenko

mcollina: That would work for me.

ivan-tymoshenko: What do you mean by `malicious code`? Whatever it means I agree with this point, because FJS doesn't check your data for any `malicious code`. But I think you talking about something different. Sanitizing is not validation.

ivan-tymoshenko: Ok, I saw the original security report. I would say that: fast-json-stringify guarantees that you will get a valid output only if your input matches the schema or can be coerced to the schema. If your input doesn't match the schema, you will get undefined behavior.

Ry0taK: See https://github.com/fastify/fast-json-stringify/issues/559 for more context

Fdawgs:
Is it also worth using [GitHub's info or warning highlighting syntax](https://github.com/community/community/discussions/16925), to draw more attention to it?

jsumners:
> Is it also worth using [GitHub's info or warning highlighting syntax](https://github.com/community/community/discussions/16925), to draw more attention to it?

I would rather not rely on GitHub specific rendering syntax. Callouts with generic Markdown are easy enough -- https://github.com/fastify/fastify/blob/675b00d6ca7a7eeaad073e6469c70b50fcd532eb/docs/Guides/Getting-Started.md?plain=1#L94-L121

delvedor (approved):

README.md
mcollina: PTAL

ivan-tymoshenko:
LGTM

ivan-tymoshenko (approved):

mcollina:
@climba03003 PTAL

climba03003 (approved):

Uzlopak:
Ah the good old undefined behaviour :)

Uzlopak (approved):

PR #562, opened 2022-11-24
perf: replace joins with fast-array-join

anonrig:


Uzlopak:
I dont like to fragment our code with more npm dependencies. We should focus on improving the perf in the runtime. For me unfortunately a -1.

anonrig:
@Uzlopak Would you change your decision if we move this to fastify organization?

mcollina:
Does this improve things at all?

Uzlopak:
I looked the changes and the places are in generating the serializer-functions. So the performence improvements should be at startup but not at runtime.

anonrig:
> Does this improve things at all?

The benchmarks did not show any improvements and were highly inconsistent across different runs.

Eomm:
I think the improvement is proportional to how many response schema the user is processing.

I recall a benchmark about it, i must searched it

climba03003 (commented):
Cold start is not really a consideration.
I would say that it is good to have but not a must.

mcollina:
I don't think we should pull in another dep without a clear performance improvement.

Thanks @anonrig anyway!

jsumners:
> I don't think we should pull in another dep without a clear performance improvement.
> 
> Thanks @anonrig anyway!

Agreed. Even if there were an improvement, I think the module's code is small enough to inline.

PR #569, opened 2023-01-09
fix: fix error message when type is array and object is object

yukha-dw:
When we attempt to stringify Object as an Array, the Object does not displayed properly (`'[object Object]'`). To simplify debugging process, we should display better error message to pinpoint where the problem using `schemaRef` property of `location` object.

Example:
```
// Schema
const schema = {
    type: 'object',
    properties: {
      arr: {
        type: 'array',
        items: {
          type: 'number'
        }
      }
    }
  };
const stringify = build(schema);
stringify({ arr: { foo: 'hello' } });
```

Original:
```
{
 "type": "TypeError",
 "message": "The value '[object Object]' does not match schema definition."
}
```
Proposed:
```
{
 "type": "TypeError",
 "message": "The value of '#/properties/arr' does not match schema definition."
}
```

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -558,7 +558,7 @@ function buildArray (location) {
 
   functionCode += `
     if (!Array.isArray(obj)) {
-      throw new TypeError(\`The value of '${location.jsonPointer}' does not match schema definition.\`)
+      throw new TypeError(\`The value of '${schemaRef}' does not match schema definition.\`)
     }
     const arrayLength = obj.length
   `
--- a/index.js
+++ b/index.js
@@ -558,7 +558,7 @@ function buildArray (location) {
 
   functionCode += `
     if (!Array.isArray(obj)) {
-      throw new TypeError(\`The value '$\{obj}' does not match schema definition.\`)
+      throw new TypeError(\`The value '$\{JSON.stringify(obj)}' does not match schema definition.\`)
     }
     const arrayLength = obj.length
   `
--- a/index.js
+++ b/index.js
@@ -558,7 +558,7 @@ function buildArray (location) {
 
   functionCode += `
     if (!Array.isArray(obj)) {
-      throw new TypeError(\`The value '$\{JSON.stringify(obj)}' does not match schema definition.\`)
+      throw new TypeError(\`The value of '${location.jsonPointer}' does not match schema definition.\`)
     }
     const arrayLength = obj.length
   `
--- a/test/typesArray.test.js
+++ b/test/typesArray.test.js
@@ -482,10 +482,10 @@ test('should throw an error when type is array and object is null', (t) => {
   }
 
   const stringify = build(schema)
-  t.throws(() => stringify({ arr: null }), new TypeError('The value \'null\' does not match schema definition.'))
+  t.throws(() => stringify({ arr: null }), new TypeError('The value of \'#/properties/arr\' does not match schema definition.'))
 })
 
-test('should throw an error when type is array and object is object', (t) => {
+test('should throw an error when type is array and object is not at an array', (t) => {
   t.plan(1)
   const schema = {
     type: 'object',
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -558,7 +558,7 @@ function buildArray (location) {
 
   functionCode += `
     if (!Array.isArray(obj)) {
-      throw new TypeError(\`The value '$\{obj}' does not match schema definition.\`)
+      throw new TypeError(\`The value '$\{JSON.stringify(obj)}' does not match schema definition.\`)
```
Uzlopak, line 561: I dont like it. Imagine that obj is a deeply nested Object. We would then JSON.stringify the whole object, just to say, that it is invalid.

Uzlopak (changes requested):
I dont like it.

index.js
yukha-dw: Indeed, how about we truncate the stringified object like this?
```
const obj = { fieldA: 'valueA', fieldB: 123, fieldC: 'valueC', fieldD: [1, 2, 3, 4, 5] };
const str = JSON.stringify(obj);
if (str.length > 24) {
    throw new TypeError(`The value '${str.slice(0, 23)}...' does not match schema definition.`);
}
throw new TypeError(`The value '${str}' does not match schema definition.`);
```

I'm not sure how to do this efficiently/neatly

Uzlopak: Its the same issue. I would probably prefer following error message:

throw new TypeError(`The value of ${key} does not match schema definition.`);

To determine key, you would need to check if you can derive the key from the location object.

climba03003: The major problem is `object` can be recursively reference itself and `JSON.stringify` do not deal with this situation.
Sometimes it is not the fault of developer but the library it using.

Referencing the `location` would be a better choice.

ivan-tymoshenko: You can put a location data here.

yukha-dw: From what I see, we could display `jsonPointer` as key?

Uzlopak: schemaId is probably not relevant  as it is just a UUID generated at runtime. So it doesnt help to track down the error. 

jsonPointer is helpful. So I would prefer jsonPointer.

jsumners:
> I dont like it.

This is not an appropriate review. Explain your issue with the PR without such negativity.

index.js
yukha-dw: I've added 8a6b27f3ccc1bc49f97e22e27daa549ef198e076 to display `jsonPointer` instead.
https://github.com/fastify/fast-json-stringify/blob/8a6b27f3ccc1bc49f97e22e27daa549ef198e076/index.js#L560-L562

Uzlopak:
@jsumners 

Relax friend ;). I first wrote the comment corresponding to the line. and then I finished the review with requesting for changes, and needed to write something, so I just duplicated the first sentence of my comment.

Uzlopak (approved):
LGTM.

ivan-tymoshenko:
@yukha-dw We already solved a problem similar to this. 
https://github.com/fastify/fast-json-stringify/blob/850cc177a106ae00f22c09aee3d08c50f9961bfa/index.js#L503

ivan-tymoshenko (changes requested):
comment above

ivan-tymoshenko (approved):
LGTM. Can you add a test when it fails from an external schema, so there would be an external schema id in the error?

yukha-dw:
> LGTM. Can you add a test when it fails from an external schema, so there would be an external schema id in the error?

I've added the test case here c1d5c43bb1332e9519c4179713a9a11812f29eef. Is it right?
https://github.com/fastify/fast-json-stringify/blob/c1d5c43bb1332e9519c4179713a9a11812f29eef/test/typesArray.test.js#L508-L532

climba03003 (approved):

PR #570, opened 2023-01-11
Array should only have "length"-check if items is an Array

Uzlopak:
If Array as additionalItems undefined or false, then it generates following code:

```
      if (arrayLength > undefined) {
        throw new Error(`Item at undefined does not match schema definition.`)
      }
```

This does not make any sense. Ajv is ignoring additionalItems if items is not an array. So we just add the additionalItems check if the index is an Array of Schemas

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -563,7 +563,7 @@ function buildArray (location) {
     const arrayLength = obj.length
   `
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #576, opened 2023-01-20
invalid value for patternproperties are validated by ajv

Uzlopak:
We can never catch that error because ajv is already validating that pattern properties are valid regex values. 

improves code coverage. 

#### Checklist


climba03003 (commented):
I would like to keep the check.
The error throw from `ajv` is not good for debugging. It tells you where the problem is but not exactly which one is problematic.

Uzlopak:
The problem is, that the schema is already rejected by ajv before we even apply the check. So there is no way to ever catch this error.

mcollina (approved):
lgtm

ivan-tymoshenko (approved):
lgtm

Eomm (approved):

PR #585, opened 2023-01-20
remove luxon

Uzlopak:
We dont need luxon when we can replace them with three liners...

#### Checklist


jsumners (commented):
I would much rather rely on luxon to deal with all of the non-trivial date calculations than attempt to do it ourselves. At least until Temporal is widely available. 

👎

Uzlopak:
I set the date to a fixed timestamp. So there is actually no transformation needed anymore ;)

mcollina (approved):
lgtm

Eomm (approved):
Simpler to read IMHO

PR #596, opened 2023-01-22
avoid intermediate strings. use const and let

Uzlopak:
I dont know why we use var for those variables, because in the generated code we use const and let in other places.

It reduces some pressure on the GC if we avoid some intermediate string variables. 

#### Checklist


mcollina (approved):
lgtm

(the use of `var` it's of the times when it was faster than const and let)

PR #599, opened 2023-01-23
discussion: optimize for very short arrays? 

Uzlopak:
I know that there is a bug still in there, but before i work further on this, I wanted to show this as a discussion point.

I believe that most arrays are very short. Like 0 to 10 entries. Of course we can have like Arrays with 1e4 and more entries. We can add a fast path for short arrays. I benched this approach and up to 3 elements it is much faster than using the for loop. 

this branch wth minSample 100:

```sh
node benchmark/bench-array.js
empty array................... x 240,717,906 ops/sec ±0.15% (193 runs sampled)
array with one number value... x 122,848,208 ops/sec ±0.58% (188 runs sampled)
array with two number values.. x 52,701,578 ops/sec ±0.12% (195 runs sampled)
array with three number values x 28,192,651 ops/sec ±0.11% (195 runs sampled)
array with three four values.. x 15,970,103 ops/sec ±0.14% (194 runs sampled)
```

same benchmark against master:

```sh
 node benchmark/bench-array.js
empty array................... x 243,649,239 ops/sec ±0.10% (189 runs sampled)
array with one number value... x 87,170,806 ops/sec ±0.35% (191 runs sampled)
array with two number values.. x 35,999,960 ops/sec ±0.65% (193 runs sampled)
array with three number values x 21,870,022 ops/sec ±0.42% (192 runs sampled)
array with three four values.. x 16,138,385 ops/sec ±0.14% (195 runs sampled)
```

So there are maybe some perf gains possible. If you give green light, I would work further on it.

#### Checklist

### DIFF
--- a/benchmark/bench-array.js
+++ b/benchmark/bench-array.js
@@ -1,0 +1,83 @@
+'use strict'
+
+const path = require('path')
+const { Worker } = require('worker_threads')
+
+const BENCH_THREAD_PATH = path.join(__dirname, 'bench-thread.js')
+
+const benchmarks = [
+  {
+    name: 'empty array',
+    schema: {
+      type: 'array',
+      items: { type: 'number' }
+    },
+    input: []
+  },
+  {
+    name: 'array with one number value',
+    schema: {
+      type: 'array',
+      items: { type: 'number' }
+    },
+    input: [1]
+  },
+  {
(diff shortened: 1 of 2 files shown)


Uzlopak:
For now I think it is more important to decide if we want a strict system.

PR #606, opened 2023-02-23
use Number.isInteger instead of bitwise operation

Uzlopak:
I forgot that the | 0 trick only handles numbers up to 2^32 properly

#### Checklist

### DIFF
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -27,7 +27,7 @@ module.exports = class Serializer {
       if (i === Infinity || i === -Infinity) {
         throw new Error(`The value "${i}" cannot be converted to an integer.`)
       }
-      if ((i | 0) === i) {
+      if (Number.isInteger(i)) {
         return '' + i
       }
       if (Number.isNaN(i)) {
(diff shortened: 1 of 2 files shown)


lib/serializer.js
```diff
@@ -27,7 +27,7 @@ module.exports = class Serializer {
       if (i === Infinity || i === -Infinity) {
         throw new Error(`The value "${i}" cannot be converted to an integer.`)
       }
-      if ((i | 0) === i) {
+      if (Number.isInteger(i)) {
```
jsumners, line 30: Shouldn't there be a test that exposes why this method is used over the other (which I agree this is the better method)? Although, I'm not sure how to represent a `BigInt` to trigger this.

Uzlopak: bigint would jump to the branch in line 39. I added a test, which theoretically should jump to that branch.

Uzlopak: It is not breaking existing code, because it would anyway be casted to an integer.

jsumners: Test clears the bar on my system 👍

jsumners (approved):

mcollina (approved):
lgtm

PR #611, opened 2023-02-26
fix: ignore defined type if anyOf and oneOf are defined

### LINKED ISSUE #290: oneOf/anyOf doesn't seem to work with objects
## 🐛 Bug Report

oneOf/anyOf doesn't seem to work with objects.

## To Reproduce

The following schema models a union type, i.e. in TypeScript terms:

```ts
type Foo = { foo: string, baz: string };
type Bar = { bar: string, baz: string };
type FooBar = Foo | Bar;
```

```js
const fastJson = require("fast-json-stringify");

const schemaFoo = {
  properties: {
    foo: { type: "string" },
    baz: { type: "string" },
  },
  required: ["foo", "baz"],
};

const schemaBar = {
  properties: {
    bar: { type: "string" },
    baz: { type: "string" },
  },
  required: ["bar", "baz"],
};

const stringify = fastJson({
  type: "object",
  anyOf: [schemaFoo, schemaBar],
});

console.log(stringify({ foo: "foo", baz: "baz" }));
console.log(stringify({ bar: "bar", baz: "baz" }));
```

Output:

```
{}
{}
```

## Expected behavior

Since both objects match the schema, both should be stringified.

## Your Environment

- *node version*: v14.16.0
- *fastify version*: 3.13.0
- *fast-json-stringify*: 2.4.2
- *os*: Ubuntu 18.04

Uzlopak:
Closes #290 

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -835,7 +835,7 @@ function buildValue (context, location, input) {
 
   let code = ''
 
-  if (type === undefined && (schema.anyOf || schema.oneOf)) {
+  if (schema.anyOf || schema.oneOf) {
     context.validatorSchemasIds.add(location.getSchemaId())
 
     const type = schema.anyOf ? 'anyOf' : 'oneOf'
--- a/test/issue-290.test.js
+++ b/test/issue-290.test.js
@@ -1,0 +1,60 @@
+'use strict'
+
+const { test } = require('tap')
+const build = require('..')
+
+test('should work with anyOf', (t) => {
+  t.plan(2)
+
+  const schemaFoo = {
+    properties: {
+      foo: { type: 'string' },
+      baz: { type: 'string' }
+    },
+    required: ['foo', 'baz']
+  }
+
+  const schemaBar = {
+    properties: {
+      bar: { type: 'string' },
+      baz: { type: 'string' }
+    },
+    required: ['bar', 'baz']
+  }
+
+  const stringify = build({
+    type: 'object',
+    anyOf: [schemaFoo, schemaBar]
+  })
+
+  t.equal(stringify({ foo: 'foo', baz: 'baz' }), '{"foo":"foo","baz":"baz"}')
+  t.equal(stringify({ bar: 'bar', baz: 'baz' }), '{"bar":"bar","baz":"baz"}')
+})
+
+test('should work with oneOf', (t) => {
+  t.plan(2)
+
+  const schemaFoo = {
+    properties: {
+      foo: { type: 'string' },
+      baz: { type: 'string' }
+    },
+    required: ['foo', 'baz']
+  }
+
+  const schemaBar = {
+    properties: {
+      bar: { type: 'string' },
+      baz: { type: 'string' }
+    },
+    required: ['bar', 'baz']
+  }
+
+  const stringify = build({
+    type: 'object',
+    oneOf: [schemaFoo, schemaBar]
+  })
+
+  t.equal(stringify({ foo: 'foo', baz: 'baz' }), '{"foo":"foo","baz":"baz"}')
+  t.equal(stringify({ bar: 'bar', baz: 'baz' }), '{"bar":"bar","baz":"baz"}')
+})
--- a/test/issue-290.test.js
+++ b/test/issue-290.test.js
@@ -58,3 +58,18 @@ test('should work with oneOf', (t) => {
   t.equal(stringify({ foo: 'foo', baz: 'baz' }), '{"foo":"foo","baz":"baz"}')
   t.equal(stringify({ bar: 'bar', baz: 'baz' }), '{"bar":"bar","baz":"baz"}')
 })
+
+test('should merge schemas properly', (t) => {
+  t.plan(1)
+
+  const schema = {
+    properties: {
+      foo: { type: 'string' }
+    },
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

ivan-tymoshenko:
We should catch a case when there is a `type` keyword at the top level or if a type is derived from other keywords and it's not equal to type in oneOf/anyOf/allOf.

Example1:
```js
{
  type: 'object',
  oneOf: [
    { type: 'string' },
    { type: 'object' }
  ]
}
```

Example2:
```js
{
  required: ['foo', 'bar'],
  oneOf: [
    { type: 'string' },
    { type: 'object' }
  ]
}
```

Ideally, we should catch this case with a [isShemaValid](https://github.com/fastify/fast-json-stringify/blob/d662aeece0c47e1e6010e07df40080a61e0fec6b/index.js#L32), although it's hard to catch it there.

Uzlopak:
isnt ajv already validating it?

ivan-tymoshenko:
It shows a warning. IMHO, it should throw an error. Anyway, we should have tests for cases

ivan-tymoshenko (changes requested):
To fix this issue, you should merge (look how we merge allOf subschemas) each oneOf/anyOf option with a parent schema. Otherwise, you will fix one problem but create another.

Complex test:
```js
test('should merge oneOf schema with parent schema', (t) => {
  t.plan(1)

  const schema = {
    properties: {
      bat: { type: 'string' }
    },
    oneOf: [
      {
        type: 'object',
        properties: {
          foo: { type: 'string' },
          bar: { type: 'string' }
        }
      },
      {
        type: 'object',
        properties: {
          foo: { type: 'string' },
          baz: { type: 'string' }
        }
      }
    ],
    required: ['foo']
  }

  t.equal(stringify({ bat: 'bat', foo: 'foo' }), '{"bat":"bat","foo":"foo"}')
  t.throws(() => stringify({ bat: 'bar' })) // foo is required
})
```

A simple test to show the breaking change:
```js
test('breaking change', (t) => {
  t.plan(1)

  const schema = {
    properties: {
      foo: { type: 'string' }
    },
    oneOf: [{ type: 'object' }]
  }

  t.equal(stringify({ foo: 'foo' }), '{"foo":"foo"}')
})
```

ivan-tymoshenko:
If you have some time to deal with it, you can think about merging nested `oneOf`/`anyOf`.
```js
test('should merge nested oneOf schemas with parent schema', (t) => {
  t.plan(1)

  const schema = {
    type: 'object',
    oneOf: [
      {
        oneOf: [
          { required: ['test1'] },
          { required: ['test2'] }
        ]
      }
    ],
    required: ['test0']
  }

  t.throws(() => stringify({ test1: 'data' })) // test0 is missing
  t.throws(() => stringify({ test0: 'data' })) // test1 is missing
})
```

Uzlopak:
That is basically a total rewrite of anyOf and oneOf logic.

ivan-tymoshenko:
That is why this issue has been waiting so long)

ivan-tymoshenko:
I would start by creating a separate module with the `mergeSchema` function. Maybe, even a separate npm package. It should deeply merge two or more schemas. And then use it here for merging oneOf/anyOf/allOf cases.

Uzlopak:
I was under the shower and had the same thought while washing my hair.

We basically should have some schema helper which creates disjunct, union and intersections of two or more schemas. 

So if we have an allOf, it should merge the schemas together. 

But in case of oneOf we have to check if actually only one schema is matching, so we basically have to call every validation function per schema and check if we have exactly one matching schema and call its corresponding serializer fn.

In case of anyOf, we have to first create a matrix of merged schemas and then have to check which schemas are matching and then call based on the determined matching schemas the correct merged schema from the matrix

E.g. we have three schemas. We need 7 variations. We can store them in an array use the bitwise positions for the lookup. So

0. Error
1. (=001b) schema 1
2. (=010b) schema 2
3. (=011b) schema 1 + schema 2
4. (=100b) schema 3
5. (=101b) schema3 + schema 1
6. (=110b) schema 3 + schema 2
7. (=111b) schema 3 + schema 2 + schema 1

and in the for loop we validate them like

let serializerArrayPos = 0
for (let i = 0, il = schemas.length; i < il; ++i) {
  if (validateBySchema(schemas[i]) (
    serializerArrayPos += (1 >> i)
  }
}

ivan-tymoshenko:
1. We treat the `oneOf` keyword as the `anyOf`. Trying to check that only one schema matches the data is a validation task. It's not necessary to have it here. It's part of another discussion.
2. Not sure that you need disjunct, union, and intersections. You need only a union. If it's an `allOf` you need to create a union of all schemas (parent schema, child1 schema, child2 schema). If it's a `oneOf` or `anyOf`, you need to create a few unions of (parent schema, child1 schema), (parent schema, child2 schema).
3. The mergeSchema should have access to the `refResolver`, because you might have references inside the schema.
4. When you merge schemas, you should track both current locations: location in the original schema and location in the merged schema. Check https://github.com/fastify/fast-json-stringify/pull/546.

Uzlopak:
Not the solution we need.

PR #636, opened 2023-07-28
improve benchmark workflow

Uzlopak:
It doesnt make sense to make a benchmark accross libs. What we need is the benchmark between branches.

This changes the benchmark to benchmark between two branches when adding the benchmark label.

Also reducing the maxSample to 100, as doing 500 samples takes a million years and is not more reliable.
 

#### Checklist

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,4 +1,4 @@
-name: Benchmark
+name: Benchmark PR
 
 on:
   pull_request_target:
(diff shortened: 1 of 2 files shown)


Uzlopak:
To see the output, see

https://github.com/Uzlopak/fast-json-stringify/pull/3

.github/workflows/benchmark.yml
```diff
@@ -31,7 +22,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 18
```
Fdawgs, line 25: Set this to lts/* so we don't have to keep bumping it?

Uzlopak: I will anyway open an issue regarding general benchmark wf.

Issue #626, opened 2023-06-01
Support TypedArrays


joshkel:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

fast-json-stringify does not work with TypedArray objects (Float32Array, etc.). Normally, using `JSON.stringify` to serialize a TypedArray results in writing it as an object (e.g., `new Float32Array([0.1,0.2])` gets serialized as `{"0":0.1,"1":0.2}`). However, since fast-json-stringify uses a JSON Schema to guide its output, it can know that a given property is an array; it seems reasonable and consistent to allow it to serialize TypedArrays as well as regular arrays.

Proposed implementation:

* The [Array.isArray](https://github.com/fastify/fast-json-stringify/blob/v5.7.0/index.js#LL557C10-L557C18) check could be broadened to also allow [TypedArrays](https://stackoverflow.com/q/15251879/25507).
* TypedArray can only hold (floating point or integer) numbers, so it would not work with many JSON Schemas. One option would be to only allow TypedArray if the JSON Schema is compatible with TypedArray (e.g., only number items); another option would be to allow it as is and then let individual elements fail validation.
* If a TypedArray is in use, largeArrayMechanism is ignored (treated as `default`).

### Motivation

Our application uses TypedArray for several purposes. This is fast internally but is noticeably slower to serialize. Using fast-json-stringify, locally patched to support TypedArray, is over 2.5x faster for a Float32Array of 4000 randomly generated numbers.

### Example

```js
const array = new Float32Array(new Array(4000).fill(1));

const stringify = fastJson({
  title: 'Example Schema',
  type: 'array',
  items: {
    type: 'number'
  }
})

console.log(stringify(array));
```

Eomm:
Thanks for reporting!
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

PR #640, opened 2023-08-15
Adding support for TypedArrays

calumrussell:
I have added support for `Uint8Array` with a test showing that it works as expected. No additional types have been created, this works by simply treating a `Uint8Array` like a normal `array`. Afaik, you cannot create a nested array this way so the changes are quite limited.

Support for more types of `TypedArray` is easy to add but as I worked on this issue further I realised that there may be wider implications for this:
- Does this conflict with JSONSchema?
- TypedArrays appear to be strongly typed (for example, creating a Uint8Array with values longer than 8 bits will always be cast down to 8 bit values) so would it be possible to improve perf by removing type checks (I looked at this briefly but I didn't see how it could be done without broader changes).
- Which `TypedArray` should be supported?

Related to - https://github.com/fastify/fast-json-stringify/issues/626 - I haven't verified the performance difference either, I am assuming this can only come from removing type checks?

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -27,10 +27,6 @@ const validLargeArrayMechanisms = [
   'json-stringify'
 ]
 
-const supportedTypedArrays = [
-  'Uint8Array'
-]
-
 const addComma = '!addComma && (addComma = true) || (json += \',\')'
 
 function isValidSchema (schema, name) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -586,7 +590,7 @@ function buildArray (context, location) {
   `
 
   functionCode += `
-    if (!Array.isArray(obj)) {
+    if (!Array.isArray(obj) && !(obj != null && (${supportedTypedArrays.map(type => ' obj.constructor.name === \'' + type + '\' ').join('||')}) )) {
```
Uzlopak, line 593: ugh. It would be better to put the array in a "global" scope and make an indexOf by the obj.constructor.name than this imho. 

I think this will slow down array significantly.

ivan-tymoshenko (changes requested):
This would not work in places where we use Ajv. To test it put an array under anyOf, oneOf or if.

index.js
calumrussell: I am not familiar with perf implications here.

Would Array.includes also be worth testing?

Tony133:

PR #641, opened 2023-08-20
feat: add validationErrors to schema mismatch errors

MoLow:
when using a schema that includes `anyOf` or `oneOf` and it fails to find a match the error is pretty worthless (in my case it is usually `'The value of \'#\' does not match schema definition.'`)
and it is not trivial to debug since stack trace points to anonymous functions composed at runtime.
this sets the validation errors on the error before throwing it. if the performance penalty is too big I suggest it should be behind a `verbose` option

### DIFF
--- a/index.js
+++ b/index.js
@@ -866,6 +866,7 @@ function buildValue (context, location, input) {
 
   if ((type === undefined || type === 'object') && (schema.anyOf || schema.oneOf)) {
     context.validatorSchemasIds.add(location.getSchemaId())
+    code = 'const errors = []\n'
 
     if (schema.type === 'object') {
       context.wrapObjects = false
@@ -885,8 +886,9 @@ function buildValue (context, location, input) {
       const schemaRef = optionLocation.getSchemaRef()
       const nestedResult = buildValue(context, optionLocation, input)
       code += `
-        ${index === 0 ? 'if' : 'else if'}(validator.validate("${schemaRef}", ${input}))
+        ${index === 0 ? 'if' : 'else if'}(validator.validate("${schemaRef}", ${input}, errors)) {
           ${nestedResult}
+        }
       `
     }
 
@@ -896,7 +898,9 @@ function buildValue (context, location, input) {
     }
 
     code += `
-      else throw new TypeError(\`The value of '${schemaRef}' does not match schema definition.\`)
+      else throw Object.assign(new TypeError(\`The value of '${schemaRef}' does not match schema definition.\`), {
+        validationErrors: errors
+      })
     `
     if (schema.type === 'object') {
       code += `
--- a/index.js
+++ b/index.js
@@ -866,7 +866,7 @@ function buildValue (context, location, input) {
 
   if ((type === undefined || type === 'object') && (schema.anyOf || schema.oneOf)) {
     context.validatorSchemasIds.add(location.getSchemaId())
-    code = 'const errors = []\n'
+    code += 'const errors = []\n'
 
     if (schema.type === 'object') {
       context.wrapObjects = false
--- a/lib/validator.js
+++ b/lib/validator.js
@@ -47,8 +47,10 @@ class Validator {
     }
   }
 
-  validate (schemaRef, data) {
-    return this.ajv.validate(schemaRef, data)
+  validate (schemaRef, data, errors) {
+    const valid = this.ajv.validate(schemaRef, data)
+    if (this.ajv.errors && Array.isArray(errors)) errors.push(...this.ajv.errors)
+    return valid
   }
 
   // Ajv does not support js date format. In order to properly validate objects containing a date,
--- a/test/any.test.js
+++ b/test/any.test.js
@@ -154,7 +154,7 @@ test('empty schema on anyOf', (t) => {
 })
 
 test('should throw a TypeError with the path to the key of the invalid value /1', (t) => {
-  t.plan(1)
+  t.plan(3)
 
   // any on Foo codepath.
   const schema = {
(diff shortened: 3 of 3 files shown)


MoLow:
CC @mcollina @ivan-tymoshenko

ivan-tymoshenko (commented):
I like the idea, but I don't uunderstand the purpose of accumulating errors. Why we can't throw them if validation fails without accumulating?

MoLow:
> I like the idea, but I don't uunderstand the purpose of accumulating errors. Why we can't throw them if validation fails without accumulating?

with `anyOf` or `oneOf` we go over all the possible schemas and if validation passes we use that schema for serialization. for the case where non of the possibilities matches - each of the errors is relevant I think

ivan-tymoshenko:
Oh, I see. Can you modify the `validate` function to also return errors when validation fails and put the error acumulation directly into [anyof/oneOf block](https://github.com/fastify/fast-json-stringify/blob/920317fc218cc84d94e10c5ca1e03da398bce261/index.js#L887) please? Reseting errors globaly might be a problem for some cases in the future.

test/any.test.js
```diff
@@ -186,7 +186,12 @@ test('should throw a TypeError with the path to the key of the invalid value /1'
 
   const stringify = build(schema)
 
-  t.throws(() => stringify({ kind: 'Baz', value: 1 }), new TypeError('The value of \'#\' does not match schema definition.'))
+  t.throws(() => stringify({ kind: 'Baz', value: 1 }), Object.assign(new TypeError('The value of \'#\' does not match schema definition.'), {
+    validationErrors: [
+      { message: 'must be equal to one of the allowed values', schemaPath: '#/properties/kind/enum', instancePath: '/kind' },
+      { message: 'must be equal to one of the allowed values', schemaPath: '#/properties/kind/enum', instancePath: '/kind' }
+    ]
+  }))
```
Eomm, line 194: Could you split this check into multiple ones?

```
try{
stringify(...)
fail('should not pass')
} catch(err){
// inspect err
}
```

lib/validator.js
```diff
@@ -47,8 +47,10 @@ class Validator {
     }
   }
 
-  validate (schemaRef, data) {
-    return this.ajv.validate(schemaRef, data)
+  validate (schemaRef, data, errors) {
+    const valid = this.ajv.validate(schemaRef, data)
+    if (this.ajv.errors && Array.isArray(errors)) errors.push(...this.ajv.errors)
```
Eomm, line 52: Maybe we should add some reference to `schemaRef` or the errors will be relative to the schema position in the `anyOf` array

Eomm (commented):
Nice addition 👍🏼

lib/validator.js
MoLow: the error already contains a `schemaPath` property

mcollina (approved):
lgtm

MoLow:
@ivan-tymoshenko @mcollina is there anything else missing to get this merged?

index.js
```diff
@@ -866,6 +866,7 @@ function buildValue (context, location, input) {
 
   if ((type === undefined || type === 'object') && (schema.anyOf || schema.oneOf)) {
     context.validatorSchemasIds.add(location.getSchemaId())
+    code = 'const errors = []\n'
```
Uzlopak, line 869:
```suggestion
    code += 'const errors = []\n'
```

MoLow: That doesn't really make any difference since https://github.com/MoLow/fast-json-stringify/blob/35f3539115d54ed495f610814ac95835407c9b7d/index.js#L865

index.js
```diff
@@ -896,7 +898,9 @@ function buildValue (context, location, input) {
     }
 
     code += `
-      else throw new TypeError(\`The value of '${schemaRef}' does not match schema definition.\`)
+      else throw Object.assign(new TypeError(\`The value of '${schemaRef}' does not match schema definition.\`), {
+        validationErrors: errors
+      })
```
Uzlopak, line 903: Why not use `cause` and avoid the Object.assign?

```suggestion
      else throw new TypeError(\`The value of '${schemaRef}' does not match schema definition.\`, {
        cause: errors
      })
```

lib/validator.js
Uzlopak, line 52: could this be faster?

```suggestion
    Array.isArray(errors) && Arrry.prototype.push.apply(errors, this.ajv.errors)
```

index.js
Uzlopak: In the future we maybe add some more steps before tis `code =` call, and then we need to debug to realize it was not `=` and not `+=`

test/any.test.js
```diff
@@ -186,11 +186,21 @@ test('should throw a TypeError with the path to the key of the invalid value /1'
 
   const stringify = build(schema)
 
-  t.throws(() => stringify({ kind: 'Baz', value: 1 }), new TypeError('The value of \'#\' does not match schema definition.'))
+  try {
+    stringify({ kind: 'Baz', value: 1 })
+    t.fail('should throw')
+  } catch (err) {
+    t.equal(err.message, 'The value of \'#\' does not match schema definition.')
+    t.same(err.validationErrors, [
```
Uzlopak, line 194: This would be then:

```suggestion
    t.same(err.cause, [
```

test/any.test.js
```diff
@@ -227,5 +237,15 @@ test('should throw a TypeError with the path to the key of the invalid value /2'
 
   const stringify = build(schema)
 
-  t.throws(() => stringify({ data: { kind: 'Baz', value: 1 } }), new TypeError('The value of \'#/properties/data\' does not match schema definition.'))
+  try {
+    stringify({ data: { kind: 'Baz', value: 1 } })
+    t.fail('should throw')
+  } catch (err) {
+    t.equal(err.message, 'The value of \'#/properties/data\' does not match schema definition.')
+    t.same(err.validationErrors, [
```
Uzlopak, line 245: dito
```suggestion
    t.same(err.cause, [
```

Uzlopak (changes requested):
I hope my remarks make sense :)
Looking forward

test/any.test.js
Eomm: i think the rationale here was to call it as fastify with ajv implementation:

https://github.com/fastify/fastify/blob/9b3ca58097954c4df20b253906ed6697461e2486/lib/handleRequest.js#L109

Tony133:

Issue #642, opened 2023-08-29
Schema serialization of anyOf/oneOf properties are parsed incorrectly


royhamulak:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.22.0

### Node.js version

18.x

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

not relevant

### Description

[PR#630](https://github.com/fastify/fast-json-stringify/pull/630) introduced new serialization for `anyOf` or `oneOf`  in object schemas.
When these are defined on an object schema, the parsed value leaves a trailing comma which breaks the JSON standard.

### Steps to Reproduce

```JS
import fastJson from 'fast-json-stringify'

// A schema that defines that `prop1` is mandatory and either `prop2` or `prop3` must be provided
// but both of these props can be provided as well
const schema = {
  type: 'object',
  properties: {
    prop1: {type: 'string'},
    prop2: {type: 'string'},
    prop3: {type: 'string'},
  },
  required: ['prop1'],
  anyOf: [{required: ['prop2']}, {required: ['prop3']}],
}

const stringify = fastJson(schema)

console.log(stringify({prop1: 'test', prop2: 'test2'})) // output: {"prop1":"test","prop2":"test2",}
console.log(stringify({prop1: 'test', prop3: 'test3'})) // output: {"prop1":"test","prop3":"test3",}
console.log(stringify({prop1: 'test', prop2: 'test2', prop3: 'test3'})) // output: {"prop1":"test","prop2":"test2","prop3":"test3",}
```
it's the same for `oneOf` instead of `anyOf`

### Expected Behavior

return a proper stringified version of the object provided without the trailing comma.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

PR #630, opened 2023-07-23
fix: support anyOf inside allOf

MoLow:
Fixes https://github.com/fastify/fastify/issues/4920

PR #645, opened 2023-09-14
fix: anyOf produces invalid json due to extra comma

samzilverberg:
PR to address https://github.com/fastify/fast-json-stringify/issues/642

It simple removes the extra comma that was introduced since https://github.com/fastify/fast-json-stringify/pull/630 .

I spent around an hour trying a better approach where the comma is not added in the first place.
but the code is hard to follow and I don't have enough time to get to the bottom of it.

added test case from the issue (was red before fix, green after)

### DIFF
--- a/index.js
+++ b/index.js
@@ -900,6 +900,8 @@ function buildValue (context, location, input) {
     `
     if (schema.type === 'object') {
       code += `
+        if (json.endsWith(','))
+          json = json.substr(0, json.length - 1)
         json += '}'
       `
       context.wrapObjects = true
--- a/test/anyof.test.js
+++ b/test/anyof.test.js
@@ -644,3 +644,22 @@ test('object with ref and validated properties', (t) => {
   const stringify = build(schema, { schema: externalSchemas })
   t.equal(stringify({ id: 1, reference: 'hi' }), '{"id":1,"reference":"hi"}')
 })
+
+test('anyOf required props', (t) => {
+  t.plan(3)
+
+  const schema = {
+    type: 'object',
+    properties: {
+      prop1: { type: 'string' },
+      prop2: { type: 'string' },
+      prop3: { type: 'string' }
+    },
+    required: ['prop1'],
+    anyOf: [{ required: ['prop2'] }, { required: ['prop3'] }]
+  }
+  const stringify = build(schema)
+  t.equal(stringify({ prop1: 'test', prop2: 'test2' }), '{"prop1":"test","prop2":"test2"}')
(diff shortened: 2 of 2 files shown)


samzilverberg:
I spent a little more time on figuring out a better way that would simple not add the commas if not needed.
It didn't work out.
The problem was with this code:
```
      code += `
        ${index === 0 ? 'if' : 'else if'}(validator.validate("${schemaRef}", ${input}))
          ${nestedResult}
      `
```

for the "anyOf" kind of block that contains "required", the "nestedResult" is just a block of code that either produces an err or adds nothing to the json.
but when the anyOf contains other schema, the nestedResult can also add something to the json.

this means the comma is dependand on what the "nestedResult" does.

the nestedResult returns something like `json += blabla` (or json += "") , and not just the value, so I can't put a condition on it to add the comma :(

a major refactor would be needed to make this work....

posting all of this to convince you this fix is the only way I see forward, unless you are interested in a major refactor, or maybe you can think of another fix...

Uzlopak:
Yeah, I see that you also dislike the solution. Maybe you try to fix it from the wrong angle?

If I remove this line

https://github.com/fastify/fast-json-stringify/blob/4073e5b6c9b2f4c6b4a3cea1f0f1d6f0ba89e769/index.js#L876

Then your unit test works without your changes.

But it will fail 'object with anyOf nested inside allOf' in allof.test.js, because then a comma is missing. So maybe, it has to be fixed for that case and not for the case you implemented. So maybe we can detect better, when we need to add a comma for this allof case instead of generally setting the comma and being forced to remove the comma in an edge case.

samzilverberg:
yes,  I ran both tests and ran some experiments with adding commas based on different conditions (and not doing the substring i did in this pr).
It always boiled down to this if condition i posted and whether the "nestedResult" added something to the json or not (which is not detectable atm :( )

I don't have a better idea atm.

sloops77:
I've played around with this locally and I may have a solution. All tests pass including the new one that is  a part of this PR

```
...
  if (schema.type === 'object') {
      context.wrapObjects = false
      const funcName = buildObject(context, location)
      code += `
        json += '{'
        json += ${funcName}(${input})       
      `
    }

    const type = schema.anyOf ? 'anyOf' : 'oneOf'
    const anyOfLocation = location.getPropertyLocation(type)

    if (schema.type === 'object' && location.schema[type].length > 0) {
      code += `
        json += ','
      `
    }
    ...
```

mcollina:
@sloops77 could you send a fresh PR?

samzilverberg:
closing in favor of more proper solution in https://github.com/fastify/fast-json-stringify/pull/651

PR #651, opened 2023-10-14
fix:  anyOf produces invalid json due to extra comma

sloops77:
PR to address https://github.com/fastify/fast-json-stringify/issues/642
Superseeds #645 

It simple removes the extra comma that was introduced since https://github.com/fastify/fast-json-stringify/pull/630 .

Inspired by previous PR and tests added by @samzilverberg in https://github.com/fastify/fast-json-stringify/pull/645

In the end even my original proposed solution did not work, instead mirrored the introduction of `context.wrapObjects` with `context.addComma` for this case. `context.addComma` serves as a default value for the properties being added when `context.wrapObjects` is false

#### Checklist


index.js
```diff
@@ -888,6 +890,9 @@ function buildValue (context, location, input) {
         ${index === 0 ? 'if' : 'else if'}(validator.validate("${schemaRef}", ${input}))
           ${nestedResult}
       `
+      if (schema.type === 'object') {
```
Uzlopak, line 893: can you explain this part please?

sloops77: all this logic is only required when an anyOf/oneOf was within an object. If you look at #630, you can see the logic that was introduced for these kinds of objects. In that case they braces are defined above, but then any of the subsequent entries of the anyOf/oneOf will need the comma added. In advance we dont know which one, so it needs to be added as a prefix to the anyOf-entry producing function when and if it executes.

mcollina (approved):
lgtm

ivan-tymoshenko (changes requested):
@sloops77 Hi, thanks for your PR. I see the problem. It has more neggative effects except adddind a comma and requires a complex collution. Read my commant [here](https://github.com/fastify/fastify/issues/4920#issuecomment-1645233089). #630 Was a workaround, not a real solution. But let's focus on adding a commas.

Decision to add or not to add a comma for an object should be made during a serialization (not in a compile time). This decision should based on knowladge was another property already added to the object or not. In this PR you move some of this logic to the compile time, which means there is a case when it wouldn't work.

I don't have a good solution here, but I would start from adding another param to the compiled funtion that would tell is there a need to add a first comma or not.

Here is a test to show the problem.
```js
test('anyOf required props', (t) => {
  t.plan(1)

  const schema = {
    type: 'object',
    properties: {
      prop1: { type: 'string' }
    },
    anyOf: [
      {
        properties: {
          prop2: { type: 'string' }
        }
      },
      {
        properties: {
          prop3: { type: 'string' }
        }
      }
    ]
  }
  t.equal(stringify({ prop2: 'test2' }), '{"prop2":"test2"}')
})

```

sloops77:
@ivan-tymoshenko a third attempt that adds a runtimeContext to the nested function calls. I've also added a new test to cover the one you pointed to as would fail.

FYI when testing this new test I added a new separate expectation that failed. 

```js
test('anyOf required props', (t) => {
  t.plan(1)

  const schema = {
    type: 'object',
    properties: {
      prop1: { type: 'string' }
    },
    anyOf: [
      {
        properties: {
          prop2: { type: 'string' }
        }
      },
      {
        properties: {
          prop3: { type: 'string' }
        }
      }
    ]
  }
  t.equal(stringify({ prop1: 'test1', prop3: 'test3' }), '{"prop1":"test1","prop3":"test3"}')
})
```

The output was
```
      diff: |
        --- expected
        +++ actual
        @@ -1,1 +1,1 @@
        -{"prop1":"test1","prop3":"test3"}
        +{"prop1":"test1"}
```

I did some console lines and it was because the wrong  `anyOf` function was called. (The function param2 instead of param3)

This seems to be unrelated to the comma bug and is instead related to the runtime `validate` function not executing as expected (its matching the first schema instead of the second)

samzilverberg:
@sloops77 IMO your last test case works as expected.
anyOf stops matching at the first schema it find that "fits".
`{ prop1: 'test1', prop3: 'test3' }` is fine against the "combined schema" that defines prop1 and prop2 (the first anyOf).
because it allows additionalProperties (prop3 being the additional).

if you want it to behave as expected maybe you need to change the anyOf to something like:
```
anyOf: [
      {
        properties: {
          prop2: { type: 'string' }
        },
        required: ['prop2']
      },
      {
        properties: {
          prop3: { type: 'string' }
        },
        required: ['prop3']
      }
```

ivan-tymoshenko (commented):
The idea is correct now. Can you please simplify it a bit? Just add an `addComma` second argument with a default `false` value and pass the `addComma` variable if it was definded.  I understand that you were trying to build more general solution, but here is a place where it's better to keep it simple. I will affect the performance.

Uzlopak:
@samzilverberg 
@sloops77 

Thank you BOTH for your contributions and testing.

mcollina:
@Uzlopak @ivan-tymoshenko could you lgtm this as well?

ivan-tymoshenko:
This should be simplified. See comment https://github.com/fastify/fast-json-stringify/pull/651#pullrequestreview-1691471725. And I would like to see benchmarks, as it affects a hot path.

sloops77:
@ivan-tymoshenko Im not quite following your suggestion. The solution relies on passing state between the parent and nested functions using the pass-by-ref semantics of the properties in the runtimeContext object. 
Changing it to a `defaultAddComma` param would require an additional return value added to buildInnerObject. is that what you would prefer?

Uzlopak:
Would you be ok, if @ivan-tymoshenko directly edits your PR?

mateus4k:
Any updates on this?

ivan-tymoshenko:
This PR will fix the issue:
https://github.com/fastify/fast-json-stringify/pull/665

ivan-tymoshenko:
#642 should be fixed now. This PR can be closed.

PR #653, opened 2023-10-20
deps: use json-schema-ref-resolver

ivan-tymoshenko:
https://github.com/fastify/json-schema-ref-resolver


test/ref.test.js
```diff
@@ -1377,7 +1377,7 @@ test('Bad key', t => {
       })
       t.fail('Should throw')
     } catch (err) {
-      t.equal(err.message, 'Cannot find reference "extrenal#/definitions/projectId"')
+      t.equal(err.message, 'Schema with id "extrenal" is not found.')
```
Eomm, line 1380: the error message is less specific

Eomm (commented):
I would mark this PR a breaking
I think we should battle-test the module before landing it - maybe test it in a real project

Ref: https://github.com/fastify/json-schema-ref-resolver/issues/1

test/ref.test.js
ivan-tymoshenko: not a problem to fix that

ivan-tymoshenko:
> I would mark this PR a breaking I think we should battle-test the module before landing it - maybe test it in a real project
> 
> Ref: [fastify/json-schema-ref-resolver#1](https://github.com/fastify/json-schema-ref-resolver/issues/1)

I don't see how it's a breacking change. Tests are passing. This package was developed from the [ref-resolver.js](https://github.com/fastify/fast-json-stringify/blob/master/lib/ref-resolver.js) module. If you look at implementation of the methods that are used, you will see that it's the same code.

test/ref.test.js
ivan-tymoshenko: https://github.com/fastify/json-schema-ref-resolver/pull/4

Eomm (approved):
> The code that is used inside of fast-json-stringify wasn't changed.

Thks, did see it at first

mcollina (approved):
lgtm

mcollina:
@ivan-tymoshenko could you resolve the conflicts?

Uzlopak (approved):
LGTM

PR #665, opened 2023-11-25
fix: update internal schema merging

### LINKED ISSUE #642: Schema serialization of anyOf/oneOf properties are parsed incorrectly

### LINKED ISSUE #648: Recursive validation schema compiles but fails at runtime
### Prerequisites

### Fastify version

4.23.2

### Plugin version

_No response_

### Node.js version

18.12.1

### Operating system

Windows

11

### Description

The following error is thrown when validating a response using a response schema, when trying to return a response:

```
{
  "statusCode": 500,
  "error": "Internal Server Error",
  "message": "can't resolve reference #/definitions/Type from id merged_45bedfdc-0c19-4649-8739-66c2f819bb2e"
}
```

This is unexpected: it's my understanding that schemas should be validated on server startup.
The error also does not throw if the response does not include the problematic parts of the schema.

This is most likely an AJV bug, but I was not able to reproduce it trivially. I would appreciate help with isolating the Fastify-AJV interaction to reproduce this without the Fastify server at all.

### Steps to Reproduce

Full sample code:
```ts
import Fastify from "fastify";

const schema = {
  $schema: "http://json-schema.org/draft-07/schema#",
  title: "Schema Response",
  type: "object",
  required: [],
  properties: {
    types: {
      description: "A list of types.",
      type: "object",
      additionalProperties: {
        $ref: "#/definitions/Type",
      },
    },
  },
  definitions: {
    Type: {
      title: "Type",
      description: "Types track the valid representations of values as JSON",
      oneOf: [
        {
          description: "A named type",
          type: "object",
          required: ["name", "type"],
          properties: {
            name: {
              description:
                "The name can refer to a primitive type or a scalar type",
              type: "string",
            },
            type: {
              type: "string",
              enum: ["named"],
            },
          },
        },
        {
          description: "A nullable type",
          type: "object",
          required: ["type", "underlying_type"],
          properties: {
            type: {
              type: "string",
              enum: ["nullable"],
            },
            underlying_type: {
              description: "The type of the non-null inhabitants of this type",
              allOf: [
                {
                  $ref: "#/definitions/Type",
                },
              ],
            },
          },
        },
      ],
    },
  },
};

const data = {
  types: {
    Int: {
      type: "named",
      name: "Int",
    },
    NullableString: {
      type: "nullable",
      underlying_type: {
        type: "named",
        name: "String",
      },
    },
  },
};

async function start_server() {
  const server = Fastify({
    logger: true,
  });

  server.get(
    "/",
    {
      schema: {
        response: {
          200: schema,
        },
      },
    },
    () => {
      return data;
    }
  );

  try {
    await server.listen({ port: 3030 });
  } catch (error) {
    server.log.error(error);
    process.exit(1);
  }
}

start_server();
```
Fastify version: 4.23.2
To reproduce the error, run the server, and perform a GET request against it.

### Expected Behavior

The response should be validated successfully, _or_ it should fail to validate, but _not_ fail to resolve the validation schema, except on server startup.

Note, the following using only AJV does _not_ error:
```ts
const schema = { /* same as above */ }
const data = { /* same as above */ }

const validate = ajv.compile(schema);
const valid = validate(data);
if (!valid) {
  console.log("validation failed");
  console.log(validate.errors);
} else {
  console.log("validation succeeded");
}
```

I strongly suspect this is an AJV issue, but as the above did not repro the problem I would appreciate any help in isolating the issue.

### Workarounds

I've observed two ways to avoid this issue:
1. use `$recursiveRef` instead of `$ref`. I don't believe this is the correct use case for `$recursiveRef`, but it does make the problem go away
2. specify an `$id` on the root, then qualify `$refs` with the id before the `#`

### Possibly related issues
[Using discriminator in a definition causes failure to compile if oneOf contains a reference fastify/fastify#1899](https://github.com/ajv-validator/ajv/issues/1899)
[Is it possible to create a recursive schema definition? fastify/fastify#659](https://github.com/fastify/help/issues/659)

ivan-tymoshenko:
This PR adds a proper schema merging for different cases:
- root schema + oneOf/anyOf subschema 
- root schema + all allOf subschemas
- root schema + then schema; root schema + else schema

Fix #648
Fix #642

### DIFF
--- a/index.js
+++ b/index.js
@@ -93,7 +93,6 @@ function build (schema, options) {
     functionsCounter: 0,
     functionsNamesBySchema: new Map(),
     options,
-    wrapObjects: true,
     refResolver: new RefResolver(),
     rootSchemaId: schema.$id || `__fjs_root_${schemaIdCounter++}`,
     validatorSchemasIds: new Set()
@@ -361,10 +360,8 @@ function buildInnerObject (context, location) {
 
   code += `
     let addComma = false
-    let json = '${context.wrapObjects ? '{' : ''}'
+    let json = '{'
   `
-  const wrapObjects = context.wrapObjects
-  context.wrapObjects = true
 
   if (schema.properties) {
     for (const key of Object.keys(schema.properties)) {
@@ -408,9 +405,8 @@ function buildInnerObject (context, location) {
     code += buildExtraObjectPropertiesSerializer(context, location)
   }
 
-  context.wrapObjects = wrapObjects
   code += `
-  return json${context.wrapObjects ? ' + \'}\'' : ''}
+    return json + '}'
   `
   return code
 }
--- a/index.js
+++ b/index.js
@@ -2,8 +2,6 @@
 
 /* eslint no-prototype-builtins: 0 */
 
-const merge = require('@fastify/deepmerge')()
-const clone = require('rfdc')({ proto: true })
 const { RefResolver } = require('json-schema-ref-resolver')
 
 const validate = require('./lib/schema-validator')
@@ -32,6 +30,8 @@ const addComma = '!addComma && (addComma = true) || (json += \',\')'
 
 let schemaIdCounter = 0
 
+const mergedSchemaRef = Symbol('fjs-merged-schema-ref')
+
 function isValidSchema (schema, name) {
   if (!validate(schema)) {
     if (name) {
@@ -46,13 +46,15 @@ function isValidSchema (schema, name) {
   }
 }
 
-function resolveRef (context, location, ref) {
+function resolveRef (context, location) {
+  const ref = location.schema.$ref
+
   let hashIndex = ref.indexOf('#')
   if (hashIndex === -1) {
     hashIndex = ref.length
   }
 
-  const schemaId = ref.slice(0, hashIndex) || location.getOriginSchemaId()
+  const schemaId = ref.slice(0, hashIndex) || location.schemaId
   const jsonPointer = ref.slice(hashIndex) || '#'
 
   const schema = context.refResolver.getSchema(schemaId, jsonPointer)
@@ -62,10 +64,15 @@ function resolveRef (context, location, ref) {
 
   const newLocation = new Location(schema, schemaId, jsonPointer)
   if (schema.$ref !== undefined) {
-    return resolveRef(context, newLocation, schema.$ref)
+    return resolveRef(context, newLocation)
   }
 
   return newLocation
+}
+
+function getMergedLocation (context, mergedSchemaId) {
+  const mergedSchema = context.refResolver.getSchema(mergedSchemaId, '#')
+  return new Location(mergedSchema, mergedSchemaId, '#')
 }
 
 function getSchemaId (schema, rootSchemaId) {
@@ -329,7 +336,7 @@ function buildInnerObject (context, location) {
       }
       let propertyLocation = propertiesLocation.getPropertyLocation(key)
       if (propertyLocation.schema.$ref) {
-        propertyLocation = resolveRef(context, location, propertyLocation.schema.$ref)
+        propertyLocation = resolveRef(context, propertyLocation)
       }
 
       const sanitizedKey = JSON.stringify(key)
@@ -362,7 +369,7 @@ function buildInnerObject (context, location) {
     for (const key of Object.keys(schema.properties)) {
       let propertyLocation = propertiesLocation.getPropertyLocation(key)
       if (propertyLocation.schema.$ref) {
-        propertyLocation = resolveRef(context, location, propertyLocation.schema.$ref)
+        propertyLocation = resolveRef(context, propertyLocation)
       }
 
       const sanitizedKey = JSON.stringify(key)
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -407,107 +407,137 @@ function buildInnerObject (context, location) {
...
+function mergeSchemas (context, locations) {
+  const mergedSchema = {}
 
-    if (allOfSchema.properties !== undefined) {
-      if (mergedSchema.properties === undefined) {
-        mergedSchema.properties = {}
-      }
-      Object.assign(mergedSchema.properties, allOfSchema.properties)
-    }
+  for (let location of locations) {
```
gurgunday, line 440: Won't a simple for loop be faster than iterators?

It was the case last time I checked

index.js
```diff
@@ -407,107 +407,137 @@ function buildInnerObject (context, location) {
...
-      mergedSchema.oneOf.push(...allOfSchema.oneOf)
-    }
+    for (const key in schema) {
+      const value = schema[key]
 
-    if (allOfSchema.anyOf !== undefined) {
-      if (mergedSchema.anyOf === undefined) {
-        mergedSchema.anyOf = []
+      if (key === '$id') continue
+      if (key === 'allOf') {
```
gurgunday, line 466: `switch` *might* be faster here because of all the string comparisons

mcollina (approved):
lgtm

index.js
ivan-tymoshenko: This kind of change doesn't create any perf difference unless it places in a _very_ _very_ high-loaded bottleneck (not this library). This library could be optimized by reducing the amount of time it clones the schemas.

Feel free to run benchmarks and prove me wrong.

index.js
ivan-tymoshenko: Ditto

index.js
```diff
@@ -407,107 +407,137 @@ function buildInnerObject (context, location) {
...
 
-    if (allOfSchema.allOf !== undefined) {
-      mergeAllOfSchema(context, location, allOfSchema, mergedSchema)
+function resolveRelativeRefs (schema, schemaId) {
+  if (schema.$ref?.charAt(0) === '#') {
+    schema.$ref = schemaId + schema.$ref
+  }
+  for (const subSchema of Object.values(schema)) {
+    if (
+      typeof subSchema === 'object' &&
```
gurgunday, line 531: Not familiar so just asking, we are sure there can't be a null value right?

```suggestion
      subSchema != null &&
```

ivan-tymoshenko: It shouldn't be, but this is a good catch. Will add a check.

index.js
gurgunday: I see, no problem I'll take your word 😄

ivan-tymoshenko:
Need a proper review for this one. This is more difficult than it might seem.

package.json
```diff
@@ -48,13 +48,13 @@
...
   "dependencies": {
-    "@fastify/deepmerge": "^1.0.0",
     "ajv": "^8.10.0",
     "ajv-formats": "^2.1.1",
     "fast-deep-equal": "^3.1.3",
     "fast-uri": "^2.1.0",
     "rfdc": "^1.2.0",
-    "json-schema-ref-resolver": "^1.0.1"
+    "json-schema-ref-resolver": "^1.0.1",
+    "json-schema-merge-allof": "^0.8.1"
```
mcollina, line 57: The use of https://www.npmjs.com/package/json-schema-merge-allof does not inspire me a lot of faith as it seems an unmaintained module. Is there an alternative? Can we fork or embed its code in here?

ivan-tymoshenko: Unfortunatelly this is the best solution I could find. We can write our own solution, but I don't have time for this right now. Anyway it should be better that our currect merge logic.

ivan-tymoshenko:
Waiting for https://github.com/fastify/merge-json-schemas release.

test/allof.test.js
```diff
@@ -553,50 +553,165 @@ test('allof with local anchor reference', (t) => {
...
 
   const schema = {
     allOf: [
       { type: 'string' },
       { type: 'number' }
     ]
   }
-  t.throws(() => build(schema), new Error('allOf schemas have different type values'))
+  try {
+    build(schema)
```
Uzlopak, line 618:
```suggestion
    build(schema)
    t.fail('Should have thrown')
```

index.js
```diff
@@ -400,139 +405,65 @@ function buildInnerObject (context, location) {
...
 
-    if (allOfSchema.additionalProperties !== undefined) {
-      if (mergedSchema.additionalProperties === undefined) {
-        mergedSchema.additionalProperties = {}
-      }
-      Object.assign(mergedSchema.additionalProperties, allOfSchema.additionalProperties)
-    }
+  if (
+    schema.$id !== undefined &&
+    schema.$id.charAt(0) !== '#'
```
Uzlopak, line 443: What if schema.$id is an empty string?

index.js
```diff
@@ -848,81 +779,193 @@ function buildConstSerializer (location, input) {
...
+  let schemaRef = location.getSchemaRef()
+  if (schemaRef.startsWith(context.rootSchemaId)) {
+    schemaRef = schemaRef.replace(context.rootSchemaId, '')
   }
 
-  if (schema.allOf) {
-    mergeAllOfSchema(context, location, schema, clone(schema))
-    schema = location.schema
+  code += `
+    else throw new TypeError(\`The value of '${schemaRef}' does not match schema definition.\`)
```
Uzlopak, line 861:
```suggestion
    else throw new TypeError(\`The value of '${schemaRef}' does not match the schema definition.\`)
```
?

test/if-then-else.test.js
```diff
@@ -249,15 +249,7 @@ const alphabetOutput = JSON.stringify({
...
-const noElseGreetingOutput = JSON.stringify({
-  kind: 'greeting',
-  foo: 'FOO',
-  bar: 42,
-  hi: 'HI',
-  hello: 45,
-  a: 'A',
-  b: 35
-})
+const noElseGreetingOutput = JSON.stringify({})
```
Uzlopak, line 252: Why did you change this?

test/allof.test.js
```diff
@@ -553,50 +553,165 @@ test('allof with local anchor reference', (t) => {
...
+  } catch (error) {
+    t.ok(error instanceof Error)
+    t.equal(error.message, 'Failed to merge "format" keyword schemas.')
+    t.same(error.schemas, ['date', 'time'])
+  }
 })
 
-test('allOf: throw Error if nullable mismatch /1', (t) => {
+test('recursive nested allOfs', (t) => {
```
Uzlopak, line 644: Why did you change these two allOf tests?

Uzlopak:
Tbh. This is a quite complex PR. I dont think that I could approve it with the full confidence that I understood everything. I read over it, and tbh. I did not make a deep dive. Just a raw review, if everything makes still kind of sense.

Also I dont want to stall to merge this, just because I am not confident enough to approve it.

For my approval I just need the confidence, that we didnt break the api contract and if we break it, it should be an informed decision.As we have a good code coverage in this project, it would enough for me, if no old tests were chaged, and if they were changed, that they dont reflect a significant behavioural change.

I would like to know why some of the tests got changed/removed in the allof.test.js, as it seems the only place where the behaviour changed significantly. If those questions are answered, I am  happy to approve it.

index.js
ivan-tymoshenko: It shouldn't be an empty string. If user explicitly add a schema $id field with empty string value, it's kind of his problem. Anyway, this code will prefix all nested schemas' $refs with an empty string then.

index.js
ivan-tymoshenko: I didn't add this error. If you want to update the error message for this error, you can create another PR.

test/if-then-else.test.js
ivan-tymoshenko: Because [the test](https://github.com/fastify/fast-json-stringify/blob/69a34a2d3d4670b8641e1ecd9df9458f30819986/test/if-then-else.test.js#L311) was wrong. 

Here is a schema: https://github.com/fastify/fast-json-stringify/blob/69a34a2d3d4670b8641e1ecd9df9458f30819986/test/if-then-else.test.js#L167

Here is an input data: https://github.com/fastify/fast-json-stringify/blob/69a34a2d3d4670b8641e1ecd9df9458f30819986/test/if-then-else.test.js#L209

As you can see the `if` condition doesn't fullfil as `greetingInput.kind` equal `greeting` and not a `foobar`, so it should use `else` schema instead of `then`. And as `else` schema is missing the result is an empty object.

test/allof.test.js
ivan-tymoshenko: I didn't change the test here. I removed `allOf: throw Error if nullable mismatch /1` test and added a couple of new tests. I dropped the `allOf: throw Error if nullable mismatch /1` because now the merging algoritm is more accurate and can resolve the situation when two schemas has different `nullable` keyword values.

If one schema has `nullable` = true and another one has `nullable` = false the merged schema will have `nullable` = false. Now `fast-json-stringify` just throws an error in this case.

Uzlopak (approved):
Thank you for the explaination. 
LGTM

Pezmc:
@Uzlopak I'm currently hitting this bug (oneOf results in a trailing comma) and can confirm that this PR resolves it. The latest release of `fast-json-strinify` was 5.10.0 three weeks ago and does not include this fix, how can I help cut a new release?

PR #674, opened 2024-01-10
Set strictSchema by default to make silent errors less likely.

### LINKED ISSUE #673: lack of strictSchema in ajv options results in silent typos
### Prerequisites

### Issue

`strictSchema` is [set to false in the ajv options](https://github.com/fastify/fast-json-stringify/blob/9b30d17d87c11628944bc9f33c491f6df6765653/lib/validator.js#L12) (and overrides any user-provided `ajv:` options to the constructor, so it cannot be set to true by the user either).

This makes it very easily to add a schema with an error that is silently ignored.

In my case I made a typo with `additionalProperties` in a fastify response schema, which meant that additional properties were not removed from a response on serialisation that should have been.

It would seem like having strictSchema:true would be a better default? (which would be  a breaking change needing a major version I suppose), or at least allow the user to set this.

I tested setting `strictSchema:true` and running the tests, and the only additional keyword used internally seems to be `kind`[1],[2], which is only used in tests.

[1]: https://github.com/fastify/fast-json-stringify/blob/9b30d17d87c11628944bc9f33c491f6df6765653/test/anyof.test.js#L206
[2]: https://github.com/fastify/fast-json-stringify/blob/9b30d17d87c11628944bc9f33c491f6df6765653/test/array.test.js#L371

autopulated:
Also allow AJV options to be overriden, and use that in tests that require additional keywords in schemas.

Fixes #673

#### Checklist

(no documentation added yet, will add docs if this is a welcome change, this will probably need to be in a future major version)

### DIFF
--- a/test/anyof.test.js
+++ b/test/anyof.test.js
@@ -218,7 +218,7 @@ test('symbol value in schema', (t) => {
     required: ['value']
   }
 
-  const stringify = build(schema, { ajv: { strictSchema: false } })
+  const stringify = build(schema)
   t.equal(stringify({ value: 'foo' }), '{"value":"foo"}')
   t.equal(stringify({ value: 'bar' }), '{"value":"bar"}')
   t.equal(stringify({ value: 'baz' }), '{"value":"baz"}')
--- a/test/anyof.test.js
+++ b/test/anyof.test.js
@@ -218,7 +218,7 @@ test('symbol value in schema', (t) => {
     required: ['value']
   }
 
-  const stringify = build(schema)
+  const stringify = build(schema, { ajv: { strictSchema: false } })
   t.equal(stringify({ value: 'foo' }), '{"value":"foo"}')
   t.equal(stringify({ value: 'bar' }), '{"value":"bar"}')
   t.equal(stringify({ value: 'baz' }), '{"value":"baz"}')
--- a/test/array.test.js
+++ b/test/array.test.js
@@ -397,7 +397,7 @@ test('object array with anyOf and symbol', (t) => {
       required: ['name', 'option']
     }
   }
-  const stringify = build(schema)
+  const stringify = build(schema, { ajv: { strictSchema: false } })
   const value = stringify([
     { name: 'name-0', option: 'Foo' },
     { name: 'name-1', option: 'Bar' }
--- a/test/array.test.js
+++ b/test/array.test.js
@@ -397,7 +397,7 @@ test('object array with anyOf and symbol', (t) => {
       required: ['name', 'option']
     }
   }
-  const stringify = build(schema, { ajv: { strictSchema: false } })
+  const stringify = build(schema)
   const value = stringify([
     { name: 'name-0', option: 'Foo' },
     { name: 'name-1', option: 'Bar' }
--- a/index.js
+++ b/index.js
@@ -5,6 +5,9 @@
 const merge = require('@fastify/deepmerge')()
 const clone = require('rfdc')({ proto: true })
 const { RefResolver } = require('json-schema-ref-resolver')
+const Ajv = require('ajv')
+const fastUri = require('fast-uri')
+const ajvFormats = require('ajv-formats')
 
 const validate = require('./lib/schema-validator')
 const Serializer = require('./lib/serializer')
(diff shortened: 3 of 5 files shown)


test/anyof.test.js
```diff
@@ -218,7 +218,7 @@ test('symbol value in schema', (t) => {
     required: ['value']
   }
 
-  const stringify = build(schema)
+  const stringify = build(schema, { ajv: { strictSchema: false } })
```
mcollina, line 221: why this?

test/array.test.js
```diff
@@ -397,7 +397,7 @@ test('object array with anyOf and symbol', (t) => {
       required: ['name', 'option']
     }
   }
-  const stringify = build(schema)
+  const stringify = build(schema, { ajv: { strictSchema: false } })
```
mcollina, line 400: why this?

mcollina (commented):
Can you add a test for the error that will be made less silent?

Uzlopak (changes requested):
please dont modify existing tests, as this implies, that this is a breaking change. Please implement a new test which shows that you can override strictSchema with a custom value.

We should discuss first, if we want to set strictSchema to true by default or not. That it was not overridable is the real issue. We could activate the strictSchema option once in the development phase of a project and ensure that the schemas are properly implemented and then turn it off again. 

Probably turning strictSchema and validateSchema was set to false by default was decided for performance reasons (maybe should have added a comment on why we set it to false). 

Lets discuss it first before we potentially break or slow down implementations ;)

test/anyof.test.js
Uzlopak: because strictSchema is via this PR set to true by default...

mcollina: But is that needed for that test to pass?

autopulated:
This is a breaking change!

Uzlopak:
Can you, @mcollina,  please clarify on my thoughts in my blocker?

I think the most important change is to make ajv settings be able to override the default values in a non breaking way. 
Changing strictSchema by default to true on the other hand can be a semver-minor or a semver-major. Also it has performance implications. 

It is also raises questions about the right default configuration. 

If we set strictSchema to true by default, maybe nobody will take the extra mile to make his implementation "production ready" by checking if strictSchema is a performance blocker and can be set to false in production. On the other hand, we say explicitly that schemas are to treated like production code. So it is the responsibility of the developer to write proper code and also potentially test it for edge cases and security reasons. 

Anyhow it would need to document it somewhere that strictSchema is by default set to true/false and what implications this has, and what this means for performance and testing responsibility by the developer.

gurgunday:
This is indeed a breaking change and I enable strict true every time I use Fastify, so I support this change

gurgunday:
Maybe those tests can instead be modified to use schemas that pass the strict check

autopulated:
> Can you add a test for the error that will be made less silent?

In the process of trying to do this, I see that the error will only be thrown on quite complex schemas, and is only thrown on serialisation, not on the building of the serialiser (the anyOf and array tests just happened to be complex enough)

So to actually achieve the desired result in most cases here the change isn't this simple one, but instead the `build-schema-validator.js`-generated `schema-validator.js` would have to respect this option.

So close this PR I guess?

autopulated:
@Uzlopak If I have time then I'll look at what needs to be done to add this validation, but since it's beyond the simple fix that I initially thought, I can't promise anything: I need to first grok more about how fast-json-stringify actually works internally.

Uzlopak:
@autopulated 
I considered making the ajv options overridable actually a good change. I

autopulated:
@Uzlopak The problem is that the AJV options there don't have the effect you think they do - they don't seem to be applied when building schemas for serialisers, and instead are only applied when actually doing serialisation of complex schemas. e.g. even with strictSchema: true, this will not throw:

autopulated:
(re-opening for further discussion, now with an implementation that actually behaves as originally intended)

So, previously AJV was only being used via Validator when this was necessary to resolve how to serialise an if-else, anyOf, or oneOf clause in a schema (at the time of serialisation). It was never invoked at the time of building the serialiser, resulting the behaviour mentioned above.

In order to actually validate the schema when building the serialiser, it is necessary to explicitly call `ajv.compile` on the schema when building the serialiser.

Additionally, it isn't desirable to use the validator's existing AJV instance to do this because that instance does not have all of the externally referenced schemas added to it (and probably should not have these added). So, I've just created a new AJV instance to use only for validating the schema, and initialised it with all of the referenced schemas.

Issues with this current version:

 * some of the tests are failing:
   * some due to external references that cannot be resolved. From a brief look it looks like these are being caused by a conflict between schemas internal $id values vs the keys used in options.schema
   * Also the const nullable test is failing
 * I don't know if this is actually desirable, considering the overhead of constructing another AJV instance and adding all the schemas to it, however all of the overhead should be at build time, with no serialisation overhead.

autopulated:
@ivan-tymoshenko There is `lib/schema-validator.js`, (no -strict version?), but this validates the schema against the json schema metaschema.

The strictSchema errors, like unknown keywords, are not actually errors in the JSON schema, (I understand that this is a deliberate choice, to maximise compatibility when accepting schemas from other people), rather the AJV strictSchema option is most useful to catch programmer errors (like typos) when someone is writing *their own* schemas (which I think is a very useful case where fast-json-stringify is being used via fastify, with schemas defined by the programmer for their own routes/models).

As the schema-validator.js is a compiled AJV validator, the AJV instance is not available at runtime to actually do the AJV strictSchema validation of the serialisation schema (enabling strictSchema option here would only check that the json schema metaschema itself passes strictSchema). 

Since the AJV instance that will be used at runtime for serialising oneOf/if else is created by Validator, I re-used the other options that validator passes to it to be consistent. This makes sure that errors can never occur only if a schema is used within oneOf/if else but be silently ignored elsewhere.

I did consider modifying the isValidSchema method, which calls schema-validator, to add the check adjacent to the schema-validator call, however since any referenced schemas have not been loaded at that point, it would require extensive modification.

I don't have any strong opinions here, and happy to implement that.

Before proceeding further though, I'd like to know what to do about the failing tests - const nullable seems to be deliberate but also be completely incompatible with AJV stictSchema, so maybe this is just dead in the water anyway?

Tony133:

PR #675, opened 2024-01-27
perf: optimize required props serialization

ivan-tymoshenko:
It removes a check before adding a comma if there are required props. Which gives you around 8-10% perf boost if object has at least one required property.

_Before:_
image

_After:_
image


Uzlopak:
awesome

mcollina (approved):
love it

Issue #680, opened 2024-02-13
When `asString` receives `undefined` an exception is thrown


kyleaupton:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.24.3

### Plugin version

5.9.1

### Node.js version

20.2.0

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Ubuntu 22.04.1

### Description

Not entirely sure this is a bug, could be intended behavior.

I have an array of stings, however one of the elements is `undefined`.

```
['test1', 'test2', 'test3', undefined]
```

If I return that from a fastify endpoint, `asString` throws an exception:

```
TypeError: Cannot read properties of undefined (reading 'toString')
    at Serializer.asString (/repo/node_modules/fast-json-stringify/lib/serializer.js:113:19)
    at anonymous1 (eval at build (/repo/node_modules/fast-json-stringify/index.js:190:23), <anonymous>:17:28)
    at anonymous0 (eval at build (/repo/node_modules/fast-json-stringify/index.js:190:23), <anonymous>:52:19)
    at serialize (/repo/node_modules/fastify/lib/reply.js:896:12)
    at preSerializationHookEnd (/repo/node_modules/fastify/lib/reply.js:516:17)
    at preSerializationHook (/repo/node_modules/fastify/lib/reply.js:500:5)
    at Reply.send (/repo/node_modules/fastify/lib/reply.js:204:7)
    at /repo/node_modules/fastify/lib/wrapThenable.js:25:15
```

Possibly related:
https://github.com/fastify/fast-json-stringify/issues/218
https://github.com/fastify/fast-json-stringify/pull/84

### Steps to Reproduce

```typescript
export default async (app: FastifyInstance) => {
  app.route({
    method: 'GET',
    url: '/test',
    schema: {
      response: {
        200: {
          type: 'object',
          properties: {
            status: { type: 'string' },
            message: { type: 'string' },
            testArray: {
              type: 'array',
              items: { type: 'string' }
            }
          }
        }
      }
    },
    handler: async () => {
      return {
        status: 'ok',
        message: 'This is a test route',
        testArray: ['test1', 'test2', 'test3', undefined]
      }
    }
  })
}
```

### Expected Behavior

It seems that `asString` will default the input to an empty string if its `null`.

[...]

PR #681, opened 2024-02-13
fix: coerce undefined to string (#680)

kyleaupton:
#### Checklist

On my end all test pass, however coverage is at 96% which is causing an error.

```
Suites:   49 passed, 49 of 49 completed
Asserts:  818 passed, of 818
Time:   4s
ERROR: Coverage for functions (96%) does not meet global threshold (100%)
-------------------------|---------|----------|---------|---------|-------------------
File                     | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s
-------------------------|---------|----------|---------|---------|-------------------
All files                |   98.54 |    98.16 |      96 |   98.69 |
 fast-json-stringify     |   99.75 |    98.21 |     100 |   99.75 |
  index.js               |   99.75 |    98.21 |     100 |   99.75 | 508
 fast-json-stringify/lib |   95.03 |    98.05 |    91.3 |   95.58 |
  location.js            |     100 |      100 |     100 |     100 |
  merge-schemas.js       |     100 |      100 |     100 |     100 |
  serializer.js          |   98.85 |      100 |    90.9 |   98.79 | 165
  standalone.js          |     100 |      100 |     100 |     100 |
  validator.js           |   83.33 |     91.3 |   85.71 |   85.71 | 36,86-90
-------------------------|---------|----------|---------|---------|-------------------
```

### DIFF
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -100,7 +100,7 @@ module.exports = class Serializer {
 
   asString (str) {
     if (typeof str !== 'string') {
-      if (str === null) {
+      if (str === null || str === undefined) {
         return '""'
       }
       if (str instanceof Date) {
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -100,7 +100,7 @@ module.exports = class Serializer {
 
   asString (str) {
     if (typeof str !== 'string') {
-      if (str === null || str === undefined) {
+      if (str == null) {
         return '""'
       }
       if (str instanceof Date) {
--- a/test/array.test.js
+++ b/test/array.test.js
@@ -195,6 +195,34 @@ buildTest({
   '@data': ['test']
 })
 
+test('array of strings with null and undefined elements coerced to empty string', (t) => {
+  t.plan(1)
+
+  const schema = {
+    type: 'object',
+    properties: {
+      foo: {
+        type: 'array',
+        items: {
+          type: 'string'
+        }
+      }
+    }
+  }
+
+  const stringify = build(schema)
+  const result = stringify({
+    foo: [
+      'foo',
+      'bar',
+      null,
+      undefined
+    ]
+  })
+
+  t.equal(result, '{"foo":["foo","bar","",""]}')
+})
+
 test('invalid items throw', (t) => {
   t.plan(1)
   const schema = {


Uzlopak (approved):
LGTM

ivan-tymoshenko (changes requested):
Hmm, this is a tricky one. When fjs receives an object with a property set to `undefined`, it skips the property. I guess, the logic here might be somehow similar. Although skipping array's items might not be the best idea.

What should fjs do if it receives an array with undefined property? Possible solutions I see:
- skip the item
- leave the item equal to the undefined
- throw an error (not a bad solution btw)

@mcollina wdyt?

@kyleaupton thanks for your PR, I will block it until we get a consistent answer on what to do in this case. If we decide to coerse it to the empty string it would require to also make some changes for all other types + a few use cases.

Uzlopak:
Feels wrong to skip Elements.

ivan-tymoshenko:
Yeah, this is not an easy one.

mcollina:
@ivan-tymoshenko this issue is about strings, casting a undefined to an empty string seems ok to me. Dropping the value seems not good, or keeping it undefined.

If we pass an number to a type string, it will be casted to a string. Why undefined should be different?

lib/serializer.js
```diff
@@ -100,7 +100,7 @@ module.exports = class Serializer {
 
   asString (str) {
     if (typeof str !== 'string') {
-      if (str === null) {
+      if (str === null || str === undefined) {
```
gurgunday, line 103: nit:

```suggestion
      if (str == null) {
```

Uzlopak: is it faster?

gurgunday:
I like this change for the same reasons as @mcollina

lib/serializer.js
gurgunday: It has its own bytecode instruction, so before JIT, it should be

ivan-tymoshenko:
> If we pass an number to a type string, it will be casted to a string. Why undefined should be different?

Because if you have an object schema with a string property and pass an object with property equal to undefined it will skip the property instead of coercing it to the empty string.

ivan-tymoshenko:
> this issue is about strings, casting a undefined to an empty string

This issue can't be only about strings anyway. It should be about all types.

Uzlopak:
Countering your argument
JSON.stringify({ a: undefined}) // {}
vs.
JSON.stringify([undefined]) // [null]

ivan-tymoshenko:
Since when did we start use JSON.stringify as a rule to go?

Uzlopak:
I use it now as an argument.

lib/serializer.js
kyleaupton: Oh wow seems to be a good bit quicker, this SO post outlines some numbers. Thanks for the input, I've learned something new!

https://stackoverflow.com/questions/7283577/speed-of-comparing-to-null-vs-undefined-in-javascript

```
➜  ~ node
Welcome to Node.js v20.5.1.
Type ".help" for more information.
> undefined == null
true
> undefined === null
false
```

Eomm (approved):
Since we are casting `null` to `""`, it makes sense to do the same for `undefined`.

In the context of `JSON.stringify`, both `null` and `undefined` result in `null`, so maintaining consistency in our handling makes sense, so both serialize as `string`.

gurgunday (approved):

gurgunday (changes requested):
So I agree that all functions need to conform to this behavior (nullish values get coerced to the falsy value of the corresponding type) if we're going to merge this:

https://github.com/fastify/fast-json-stringify/blob/dfb627e2545c2d5b56261218dacf00348762b5d7/lib/serializer.js#L53-L62

Tony133:

PR #698, opened 2024-03-19
perf: avoid unnecessary operation for improve performance

cesco69:
This PR has some microptimization for reduce the overhead of function call that can replaced with more faster inline control 

Side note: the code become a bit ugly but seems more faster

PS. someone can run benchmark on github (master vs PR)? I have noticed my PC is a bit unstable on benchmark results

Master
```
short string............................................. x 21,651,330 ops/sec ±1.82% (185 runs sampled)
unsafe short string...................................... x 890,279,050 ops/sec ±1.42% (181 runs sampled)
short string with double quote........................... x 12,229,862 ops/sec ±2.59% (190 runs sampled)
long string without double quotes........................ x 15,138 ops/sec ±0.37% (190 runs sampled)
unsafe long string without double quotes................. x 1,022,578,589 ops/sec ±0.39% (186 runs sampled)
long string.............................................. x 15,864 ops/sec ±0.47% (189 runs sampled)
unsafe long string....................................... x 1,054,094,454 ops/sec ±0.47% (190 runs sampled)
number................................................... x 1,060,238,499 ops/sec ±0.43% (190 runs sampled)
integer.................................................. x 219,863,056 ops/sec ±0.41% (189 runs sampled)
```
PR
```
short string............................................. x 23,801,382 ops/sec ±0.55% (189 runs sampled)
unsafe short string...................................... x 1,049,266,290 ops/sec ±0.35% (191 runs sampled)
short string with double quote........................... x 12,629,236 ops/sec ±0.43% (191 runs sampled)
long string without double quotes........................ x 15,461 ops/sec ±0.36% (190 runs sampled)
unsafe long string without double quotes................. x 1,060,363,294 ops/sec ±0.41% (191 runs sampled)
long string.............................................. x 16,106 ops/sec ±0.35% (191 runs sampled)
unsafe long string....................................... x 1,047,641,170 ops/sec ±0.37% (191 runs sampled)
number................................................... x 1,075,211,555 ops/sec ±0.43% (191 runs sampled)
integer.................................................. x 1,050,029,012 ops/sec ±0.44% (189 runs sampled)
```

#### Checklist

### DIFF
--- a/test/fix-604.test.js
+++ b/test/fix-604.test.js
@@ -20,5 +20,5 @@ const render = fjs(schema)
 try {
   render(input)
 } catch (err) {
-  t.equal(err.message, 'The value cannot be converted to a number.')
+  t.equal(err.message, 'The value "phone" cannot be converted to a number.')
 }
--- a/test/fix-604.test.js
+++ b/test/fix-604.test.js
@@ -20,5 +20,5 @@ const render = fjs(schema)
 try {
   render(input)
 } catch (err) {
-  t.equal(err.message, 'The value "phone" cannot be converted to a number.')
+  t.equal(err.message, 'The value cannot be converted to a number.')
 }
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -24,41 +24,33 @@ module.exports = class Serializer {
   }
 
   asInteger (i) {
-    if (typeof i === 'number') {
-      if (Number.isInteger(i)) {
-        return '' + i
-      }
-      // check if number is Infinity or NaN
-      // eslint-disable-next-line no-self-compare
-      if (i === Infinity || i === -Infinity || i !== i) {
-        throw new Error(`The value "${i}" cannot be converted to an integer.`)
-      }
-      return this.parseInteger(i)
-    } else if (i === null) {
+    if (i === null) {
       return '0'
     } else if (typeof i === 'bigint') {
       return i.toString()
     } else {
       /* eslint no-undef: "off" */
-      const integer = this.parseInteger(i)
-      if (Number.isFinite(integer)) {
-        return '' + integer
-      } else {
-        throw new Error(`The value "${i}" cannot be converted to an integer.`)
+      i = this.parseInteger(i)
+      // check if number is Infinity or NaN
+      // eslint-disable-next-line no-self-compare
+      if (i === Infinity || i === -Infinity || i !== i) {
+        throw new Error('The value cannot be converted to an integer.')
       }
+      return '' + i
     }
   }
 
   asNumber (i) {
-    const num = Number(i)
+    // fast cast to number
+    i = +i
     // check if number is NaN
     // eslint-disable-next-line no-self-compare
-    if (num !== num) {
-      throw new Error(`The value "${i}" cannot be converted to a number.`)
-    } else if (!Number.isFinite(num)) {
+    if (i !== i) {
+      throw new Error('The value cannot be converted to a number.')
+    } else if (i === Infinity || i === -Infinity) {
       return 'null'
     } else {
-      return '' + num
+      return '' + i
     }
   }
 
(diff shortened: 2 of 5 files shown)


test/fix-604.test.js
```diff
@@ -20,5 +20,5 @@ const render = fjs(schema)
 try {
   render(input)
 } catch (err) {
-  t.equal(err.message, 'The value "phone" cannot be converted to a number.')
+  t.equal(err.message, 'The value cannot be converted to a number.')
```
ivan-tymoshenko, line 23: This is a completelly useless error.

ivan-tymoshenko (changes requested):

test/fix-604.test.js
Uzlopak: actually avoids a potential reflection attack.

cesco69: After casting or parsing, I have used the same variable to avoid allocating a new one. However, the issue is that the original value is lost and message can't be specific. This is a side effect of my PR, if is a problem, we can close! :(

gurgunday: I like the flow changes and the use of language syntax rather than function calls (eg, `+` instead of `Number()`)

Since we're micro optimizing, feels like it should still be a little faster even with the allocation

Can you not add it back and see if there is a gain?

cesco69: On my PC the difference is negligible

cesco69: > actually avoids a potential reflection attack.

also a XSS if the error is send into the response

cesco69: @ivan-tymoshenko @gurgunday  I have reverted the error message https://github.com/fastify/fast-json-stringify/pull/698/commits/ba9de7d0769f98c46e3ea273e49d65e4406fe56a

Uzlopak:
@cesco69 

Why closed?

cesco69:
@Uzlopak I thought you don't want to merge anymore :D

gurgunday:
I want to check the V8 bytecode to see if a switch typeof uses string comparison or not, `typeof exp === "{type}"` doesn't but `switch` might

Uzlopak:
The problem is, that we have always maendering interests, so it is kind of hard to always get things done in a timely manner. So just dont think we are not interested, we are just busy, like you ;).

gurgunday:
Seems like I cannot get the `TestTypeOf` bytecode without using `===`; we instead get the good ol' `TestEqualStrict` bytecode for each `case`, which means that string comparison is taking place, there is no way that it can be faster

```js
const a = (b) => {
  if (typeof b === "function") {
    return b;
  }
};
```

```bash
[generated bytecode for function: a (0x328e80b5b9f1 <SharedFunctionInfo a>)]
Bytecode length: 11
Parameter count 2
Register count 0
Frame size 0
Bytecode age: 0
 3174 S> 0x328e80b5ccae @    0 : 0b 03             Ldar a0
         0x328e80b5ccb0 @    2 : 20 06             TestTypeOf #6
         0x328e80b5ccb2 @    4 : 99 05             JumpIfFalse [5] (0x328e80b5ccb7 @ 9)
 3209 S> 0x328e80b5ccb4 @    6 : 0b 03             Ldar a0
 3218 S> 0x328e80b5ccb6 @    8 : a9                Return
         0x328e80b5ccb7 @    9 : 0e                LdaUndefined
 3223 S> 0x328e80b5ccb8 @   10 : a9                Return
Constant pool (size = 0)
Handler Table (size = 0)
Source Position Table (size = 12)
0x328e80b5ccc1 <ByteArray[12]>
```

```js
const a = (b) => {
  switch (typeof b) {
    case "function":
      return b;
  }
};
```

```bash
[generated bytecode for function: a (0x1ded2941b9f1 <SharedFunctionInfo a>)]
Bytecode length: 18
Parameter count 2
Register count 1
Frame size 8
Bytecode age: 0
 3174 S> 0x1ded2941ccc6 @    0 : 0b 03             Ldar a0
         0x1ded2941ccc8 @    2 : 56                TypeOf
         0x1ded2941ccc9 @    3 : c4                Star0
         0x1ded2941ccca @    4 : 13 00             LdaConstant [0]
         0x1ded2941cccc @    6 : 6c fa 00          TestEqualStrict r0, [0]
         0x1ded2941cccf @    9 : 98 04             JumpIfTrue [4] (0x1ded2941ccd3 @ 13)
         0x1ded2941ccd1 @   11 : 8a 05             Jump [5] (0x1ded2941ccd6 @ 16)
 3221 S> 0x1ded2941ccd3 @   13 : 0b 03             Ldar a0
 3230 S> 0x1ded2941ccd5 @   15 : a9                Return
         0x1ded2941ccd6 @   16 : 0e                LdaUndefined
 3235 S> 0x1ded2941ccd7 @   17 : a9                Return
Constant pool (size = 1)
0x1ded2941cc79: [FixedArray] in OldSpace
 - map: 0x0624cc240211 <Map(FIXED_ARRAY_TYPE)>
 - length: 1
           0: 0x0624cc246351 <String[8]: #function>
Handler Table (size = 0)
Source Position Table (size = 12)
0x1ded2941ccd9 <ByteArray[12]>
```

gurgunday (changes requested):
https://github.com/fastify/fast-json-stringify/pull/698#issuecomment-2049615981

The rest of the PR (`Number` -> `+`, `isFinite` -> `!== Infinity || !== -Infinity`) LGTM

gurgunday (approved):

mcollina (approved):
lgtm

ivan-tymoshenko (approved):
lgtm

cesco69:
@mcollina seem all "ok" for the merge!

PR #717, opened 2024-04-29
workflows: update and pin node versions

gurgunday:


mcollina:
windows on node 22 must be skipped for now

.github/workflows/ci.yml
```diff
@@ -17,7 +17,8 @@ on:
 
 jobs:
   test:
-    uses: fastify/workflows/.github/workflows/plugins-ci.yml@v4.1.0
+    uses: fastify/workflows/.github/workflows/plugins-ci.yml@v4.2.1
```
climba03003, line 20: I believe we can release `v4`, because the `node` version is now configurable.

gurgunday: We’d have to configure every plugin manually, so not sure

I’m pinning libraries right now as they are fastify independent

gurgunday: Yes it does, but the default tests for 18, 20, 21

main branches still test for 14, 16, so they would all need manual modification, no?

gurgunday: But yeah not saying we cannot do it

climba03003: Eventually, we need to update all of them back to `v4` tag.
We can do this earlier if we release.

gurgunday: Yeah actually could be good as even next versions currently have hardcoded versions... which in not ideal

PR #734, opened 2024-08-15
Handle items: [{ $ref }]

rtkevin:
Need to resolveRef here else buildArrayTypeCondition is given undefined item.type and returns undefined, which is falsy so throws the Item at i does not match schema definition error.

#### Checklist

added test that fails in current master and does not fail in patch.


index.js
```diff
@@ -564,9 +564,14 @@ function buildArray (context, location) {
...
-      const item = itemsSchema[i]
+      let item = itemsSchema[i]
+      let itemLocation = itemsLocation.getPropertyLocation(i)
+      if (item.$ref) {
+        itemLocation = resolveRef(context, itemLocation)
+        item = itemLocation.schema
+      }
       functionCode += `value = obj[${i}]`
-      const tmpRes = buildValue(context, itemsLocation.getPropertyLocation(i), 'value')
+      const tmpRes = buildValue(context, itemLocation, 'value')
```
climba03003, line 574: Shouldn't `buildValue` itself will resolve `$ref`?

rtkevin: Maybe, I only looked into the code enough to fix the problem with minimal changes.  But my problem wasn't with buildValue, it was that buildArrayTypeCondition was given item.type which was undefined since item was still {$ref }.

kalvenschraut:
@climba03003 @ivan-tymoshenko  
Any thoughts or path forward for this issue? We are currently maintaining a patch with this fix. Open to really any direction as we provided our test case where we were experiencing the problem. LMK if you need anything more from us.

index.js
kalvenschraut: buildValue can resolve it but the code needs item to be resolved outside of buildValue for the buildArrayTypeCondition call below. I started looking into this a bit more after the latest release as we currently are maintaining a patch of this fix.

climba03003 (approved):

gurgunday (approved):

climba03003 (approved):

PR #763, opened 2025-02-15
Add JSON accelerator to benchmarks

mcollina:


benchmark/bench-cmp-lib.js
```diff
@@ -1,4 +1,4 @@
-'use strict'
+'use strictcreateAccelerator'
```
Fdawgs, line 1: Can't find any documentation around this, is this just an error?

climba03003 (approved):

PR #774, opened 2025-06-11
refactor: use built-in wellformed operation

climba03003:
From my local benchmark, it vary too large to conclude it has improvement or regression.
#### Checklist

- [ ] run `npm run test` and `npm run benchmark`
- [ ] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

PR #788, opened 2025-09-23
test: isWellFormed (PR 774) break the lib

cesco69:
`isWellFormed` don't works like the regex `/[\u0000-\u001f\u0022\u005c\ud800-\udfff]/`

This commit must be reverted
https://github.com/fastify/fast-json-stringify/commit/7f48cb6b21e41ccc37fbad09338d6c7e5f107ab3

This PR https://github.com/fastify/fast-json-stringify/pull/774 just pass the test, because no test match the condition >40 chars <5000

I have added a test (serialize medium string) that enter in this condition
```
# Subtest: serialize short string
ok 362 - serialize short string

# Subtest: serialize medium string
not ok 363 - serialize medium string

# Subtest: serialize long string
ok 364 - serialize long string
```

The regex `[\u0000-\u001f\u0022\u005c\ud800-\udfff]/` check for:
1. ASCII (U+0000–U+001F)
3. Char `"`
5. Char `\`
4. UTF-16 (\uD800–\uDFFF)

`isWellFormed()` only check case 4

These are case need escape but isWellFormed says are good!
```js
'"'.isWellFormed()
true

'\\'.isWellFormed()
true

'\n'.isWellFormed()
true
```

### DIFF
--- a/test/string.test.js
+++ b/test/string.test.js
@@ -34,6 +34,21 @@ test('serialize short string', (t) => {
   t.assert.equal(JSON.parse(output), input)
 })
 
+test('serialize medium string', (t) => {
+  t.plan(2)
+
+  const schema = {
+    type: 'string'
+  }
+
+  const input = new Array(2e4).fill('\x00').join('')
+  const stringify = build(schema)
+  const output = stringify(input)
+
+  t.assert.equal(output, `"${new Array(150).fill('\\u0000').join('')}"`)
+  t.assert.equal(JSON.parse(output), input)
+})
+
 test('serialize long string', (t) => {
   t.plan(2)
 


nigrosimone:
isWellFormed return true for string need escaping!
```js
const str = 'abc\ndef';
str.isWellFormed()
true
JSON.stringify(str)
'"abc\\ndef"'
```

gurgunday (commented):
Seems to be true

gurgunday:
@Uzlopak wdyt?

Uzlopak:
@gurgunday 

Well, @cesco69  ist right. 

According to MDN [isWellFormed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/isWellFormed) just gives true if the string does not contain lone surrogates. Thats what @cesco69 means with only case 4 is covered.

So to fix this we have two options:

1. revert to the old regex,
1. use isWellFormed in combination with a shorter regex, like this pseudocode
```js
const ASCII_ESCAPE = /[\u0000-\u001f\u0022\u005c]/u

if (str.isWellFormed() && ASCII_ESCAPE.test(str) === false) {
  // we need a bigger boat
}
```

PR #791, opened 2025-09-27
fix: initialize point variable in asString method

nigrosimone:
it is not necessary


Uzlopak (changes requested):
I think I did this on purpose to make it possible for v8 to make the variable monomorphic or so.

PR #792, opened 2025-09-27
fix: isWellFormed

### LINKED ISSUE #793: Newlines in serialized bodies
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

Working through a regression in 6.1.0 where serialized bodies containing newline characters causing the body to be invalid json. Overriding to 6.0.1 seems to fix the issue. Coming up with a minimal test case. 

reproduction: https://gist.github.com/bcomnes/915f1593aaed9fa8608eb454cdac6937

Overriding back to fast-json-stringify 6.0.1 and it works. Trying to return with a return schema causes newline to show up in the response body.

### LINKED ISSUE #794: 6.1.0: Breaks Fastify error serialization for messages which contain quotes
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.6.1

### Plugin version

6.1.0

### Node.js version

23.11.0

### Operating system

macOS

15.61.

### Description

After updating `fast-json-stringify` to 6.1.0, error messages which contain double-quotes (`"`) are not correctly serialized to JSON when using a custom Validator Compiler which returns a `FastifyValidationResult` that contains an `errors` array.

This was introduced in #774 when the error serializer code was switched to use `isWellFormed`, rather than evaluating against a regular expression. The regular expression which was present before:

```
/[\u0000-\u001f\u0022\u005c\ud800-\udfff]/
```

Detected the following characters:
- Control characters
- Quotation marks
- Backslashes

The switch to only use `isWellFormed` removes the backslash and quotation detection.

### Link to code that reproduces the bug

https://github.com/Senney/fastify-fast-json-stringify-bug

### Expected Behavior

Validation error messages which contain quotation marks should pass through `JSON.stringify` as expected, or be escaped through other means.

nigrosimone:
#### Checklist

see https://github.com/fastify/fast-json-stringify/pull/792#issuecomment-3342412747

fix https://github.com/fastify/fast-json-stringify/issues/794
fix https://github.com/fastify/fast-json-stringify/issues/793

### DIFF
--- a/benchmark/bench.js
+++ b/benchmark/bench.js
@@ -41,6 +41,13 @@ for (let i = 0; i < SHORT_ARRAY_SIZE; i++) {
 
 const benchmarks = [
   {
+    name: 'empty string',
+    schema: {
+      type: 'string'
+    },
+    input: ''
+  },
+  {
     name: 'short string',
     schema: {
       type: 'string'
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -95,13 +95,7 @@ module.exports = class Serializer {
 
     if (len === 0) {
       return '""'
-    }
-
-    if (!str.isWellFormed()) {
-      return JSON.stringify(str)
-    }
-
-    if (len < 42) {
+    } else if (len < 42) {
       // magically escape strings for json
       // relying on their charCodeAt
       // everything below 32 needs JSON.stringify()
@@ -120,13 +114,13 @@ module.exports = class Serializer {
           last === -1 && (last = 0)
           result += str.slice(last, i) + '\\'
           last = i
-        } else if (point < 32) {
+        } else if (point < 32 || (point >= 0xd800 && point <= 0xdfff)) {
           // The current character is non-printable characters or a surrogate.
           return JSON.stringify(str)
         }
       }
       return (last === -1 && ('"' + str + '"')) || ('"' + result + str.slice(last) + '"')
-    } else if (len < 5000 && ASCII_ESCAPE.test(str) === false) {
+    } else if (len < 5000 && str.isWellFormed() && ASCII_ESCAPE.test(str) === false) {
       // Only use the regular expression for shorter input. The overhead is otherwise too much.
       return '"' + str + '"'
     } else {
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -1,5 +1,8 @@
 'use strict'
 
+// eslint-disable-next-line
+const ASCII_ESCAPE = /[\u0000-\u001f\u0022\u005c]/
+
 module.exports = class Serializer {
   constructor (options) {
     switch (options && options.rounding) {
@@ -89,6 +92,7 @@ module.exports = class Serializer {
 
   asString (str) {
     const len = str.length
+
     if (len === 0) {
       return '""'
     } else if (len < 42) {
@@ -116,7 +120,7 @@ module.exports = class Serializer {
         }
       }
       return (last === -1 && ('"' + str + '"')) || ('"' + result + str.slice(last) + '"')
-    } else if (len < 5000 && str.isWellFormed()) {
+    } else if (len < 5000 && str.isWellFormed() && ASCII_ESCAPE.test(str) === false) {
       // Only use the regular expression for shorter input. The overhead is otherwise too much.
       return '"' + str + '"'
     } else {
--- a/lib/serializer.js
+++ b/lib/serializer.js
@@ -95,7 +95,13 @@ module.exports = class Serializer {
 
     if (len === 0) {
       return '""'
-    } else if (len < 42) {
+    }
+
+    if (!str.isWellFormed()) {
+      return JSON.stringify(str)
+    }
+
+    if (len < 42) {
       // magically escape strings for json
       // relying on their charCodeAt
       // everything below 32 needs JSON.stringify()
@@ -114,13 +120,13 @@ module.exports = class Serializer {
           last === -1 && (last = 0)
           result += str.slice(last, i) + '\\'
           last = i
-        } else if (point < 32 || (point >= 0xD800 && point <= 0xDFFF)) {
+        } else if (point < 32) {
           // The current character is non-printable characters or a surrogate.
           return JSON.stringify(str)
         }
       }
       return (last === -1 && ('"' + str + '"')) || ('"' + result + str.slice(last) + '"')
-    } else if (len < 5000 && str.isWellFormed() && ASCII_ESCAPE.test(str) === false) {
+    } else if (len < 5000 && ASCII_ESCAPE.test(str) === false) {
       // Only use the regular expression for shorter input. The overhead is otherwise too much.
       return '"' + str + '"'
     } else {
--- a/test/string.test.js
+++ b/test/string.test.js
@@ -34,6 +34,21 @@ test('serialize short string', (t) => {
   t.assert.equal(JSON.parse(output), input)
 })
 
+test('serialize medium string', (t) => {
+  t.plan(2)
+
+  const schema = {
+    type: 'string'
+  }
+
+  const input = new Array(150).fill('\x00').join('')
+  const stringify = build(schema)
+  const output = stringify(input)
+
+  t.assert.equal(output, `"${new Array(150).fill('\\u0000').join('')}"`)
+  t.assert.equal(JSON.parse(output), input)
+})
+
 test('serialize long string', (t) => {
   t.plan(2)
 


Uzlopak:
This is imho not acceptable for fairness reasons. as the test is basically from @cesco69. You should try to give him atleast the credentials for the test.

IMHO checkout his PR by doing `gh pr checkout 788`, then create a new branch from this, as you have his commit in the git commit historie, add your solution as another commit on top of his commit.

Uzlopak:
Ah, ok. :)

Yeah, i thought so too.  I guess we need to revert it completely

nigrosimone:
@Uzlopak I found a better combination, it seems to perform well in benchmarks
master
```
short string............................................. x 23,010,004 ops/sec ±0.58% (190 runs sampled)
unsafe short string...................................... x 1,062,255,820 ops/sec ±0.37% (191 runs sampled)
short string with double quote........................... x 13,749,470 ops/sec ±0.37% (191 runs sampled)
long string without double quotes........................ x 15,661 ops/sec ±0.31% (192 runs sampled)
unsafe long string without double quotes................. x 1,062,491,836 ops/sec ±0.37% (191 runs sampled)
long string.............................................. x 16,092 ops/sec ±0.32% (192 runs sampled)
unsafe long string....................................... x 1,060,550,353 ops/sec ±0.31% (192 runs sampled)
```

PR
```
short string............................................. x 23,078,714 ops/sec ±1.87% (178 runs sampled)
unsafe short string...................................... x 119,577,216 ops/sec ±5.05% (160 runs sampled)
short string with double quote........................... x 16,375,591 ops/sec ±2.20% (179 runs sampled)
long string without double quotes........................ x 44,909 ops/sec ±1.13% (186 runs sampled)
unsafe long string without double quotes................. x 116,764,398 ops/sec ±3.85% (164 runs sampled)
long string.............................................. x 33,744 ops/sec ±5.16% (163 runs sampled)
unsafe long string....................................... x 92,154,003 ops/sec ±5.08% (153 runs sampled)
```

 ```js
 asString (str) {
    const len = str.length

    if (len === 0) {
      return '""'
    }

    if (!str.isWellFormed()) {
      return JSON.stringify(str)
    }

    if (len < 42) {
      // magically escape strings for json
      // relying on their charCodeAt
      // everything below 32 needs JSON.stringify()
      // every string that contain surrogate needs JSON.stringify()
      // 34 and 92 happens all the time, so we
      // have a fast case for them
      let result = ''
      let last = -1
      let point = 255
      for (let i = 0; i < len; i++) {
        point = str.charCodeAt(i)
        if (
          point === 0x22 || // '"'
          point === 0x5c // '\'
        ) {
          last === -1 && (last = 0)
          result += str.slice(last, i) + '\\'
          last = i
        } else if (point < 32) {
          // The current character is non-printable characters or a surrogate.
          return JSON.stringify(str)
        }
      }
      return (last === -1 && ('"' + str + '"')) || ('"' + result + str.slice(last) + '"')
    } else if (len < 5000 && ASCII_ESCAPE.test(str) === false) {
      // Only use the regular expression for shorter input. The overhead is otherwise too much.
      return '"' + str + '"'
    } else {
      return JSON.stringify(str)
    }
  }
```

**Review by gurgunday — APPROVED** [review]
> Hi @Uzlopak, @cesco69 is my account at work :D we are the same person

Lol, good one cesco

**Comment by nigrosimone** [comment]
I have made a benchmark for test only `asString` over different type of string (short, long, with and without surrogate ect)

TLTR: On node 24 isWellFormed with a short regex (ASCII_ESCAPE) seem faster. 

On my PC (cpu: Intel(R) Core(TM) i7-1065G7 CPU @ 1.30GHz)
```
runtime: node 24.0.2 (x64-win32)
asStringNew +1.51…+1.52x faster than asStringOld

runtime: node 22.13.0 (x64-win32)
asStringNew +1.92…+2.12x faster than asStringOld

runtime: node 20.14.0 (x64-win32)
asStringNew +1.55…+1.9x faster than asStringOld
```

`asStringNew` offers substantial performance gains over the old implementation, especially for longer or escape-heavy strings.
image


see online the benchmark https://bolt.new/~/mitata-vyo1nyyz (NOTE: it run in a webcontainer, and asStringNew is slower)
```js
import { run, bench, summary, do_not_optimize } from "mitata";

/**
 * NEW IMPLEMENTATION OF asString
 * 
 * Uses String.prototype.isWellFormed() to skip regex check for surrogate pairs
 */
const ASCII_ESCAPE = /[\u0000-\u001f\u0022\u005c]/;
const asStringNew = (str) => {
  const len = str.length;

  if (len === 0) {
    return '""';
  } else if (len < 42) {
    // magically escape strings for json
    // relying on their charCodeAt
    // everything below 32 needs JSON.stringify()
    // every string that contain surrogate needs JSON.stringify()
    // 34 and 92 happens all the time, so we
    // have a fast case for them
    let result = "";
    let last = -1;
    let point = 255;
    for (let i = 0; i < len; i++) {
      point = str.charCodeAt(i);
      if (
        point === 0x22 || // '"'
        point === 0x5c // '\'
      ) {
        last === -1 && (last = 0);
        result += str.slice(last, i) + "\\";
        last = i;
      } else if (point < 32 || (point >= 0xd800 && point <= 0xdfff)) {
        // The current character is non-printable characters or a surrogate.
        return JSON.stringify(str);
      }
    }
    return (
      (last === -1 && '"' + str + '"') || '"' + result + str.slice(last) + '"'
    );
  } else if (len < 5000 && str.isWellFormed() && ASCII_ESCAPE.test(str) === false) {
    // Only use the regular expression for shorter input. The overhead is otherwise too much.
    return '"' + str + '"';
  } else {
    return JSON.stringify(str);
  }
};

/**
 * OLD IMPLEMENTATION OF asString
 * 
 * Uses a regex to check for surrogate pairs
 */
const STR_ESCAPE = /[\u0000-\u001f\u0022\u005c\ud800-\udfff]/;
const asStringOld = (str) => {
  const len = str.length;
  if (len === 0) {
    return '""';
  } else if (len < 42) {
    // magically escape strings for json
    // relying on their charCodeAt
    // everything below 32 needs JSON.stringify()
    // every string that contain surrogate needs JSON.stringify()
    // 34 and 92 happens all the time, so we
    // have a fast case for them
    let result = "";
    let last = -1;
    let point = 255;
    for (let i = 0; i < len; i++) {
      point = str.charCodeAt(i);
      if (
        point === 0x22 || // '"'
        point === 0x5c // '\'
      ) {
        last === -1 && (last = 0);
        result += str.slice(last, i) + "\\";
        last = i;
      } else if (point < 32 || (point >= 0xd800 && point <= 0xdfff)) {
        // The current character is non-printable characters or a surrogate.
        return JSON.stringify(str);
      }
    }
    return (
      (last === -1 && '"' + str + '"') || '"' + result + str.slice(last) + '"'
    );
  } else if (len < 5000 && STR_ESCAPE.test(str) === false) {
    // Only use the regular expression for shorter input. The overhead is otherwise too much.
    return '"' + str + '"';
  } else {
    return JSON.stringify(str);
  }
};

/**
 * BENCHMARKS CASES
 */
const shortString = 'a'.repeat(10)
const longString = 'a'.repeat(1000)
const shortDirtyString = shortString + '\n'
const longDirtyString = longString + '\n'
const surrogate = '\uD800';
const escapeChar = '"';
const shortSurrogateString = 'abc' + surrogate;
const shortQuoteString = 'abc' + escapeChar;
const longSurrogateString = 'abc' + surrogate;
const longQuoteString = 'abc' + escapeChar;

const args = { 
  str: [
    '',
    shortString, longString, 
    shortDirtyString, longDirtyString,
    shortSurrogateString, shortQuoteString,
    longSurrogateString, longQuoteString
  ] 
};

/**
 * BENCHMARKS IMPLEMENTATIONS
 */
summary(() => {
  bench("asStringNew", function* xNew(state) {
    const str = state.get('str');
    yield () => do_not_optimize(asStringNew(str));
  }).gc("inner").args(args);

  bench("asStringOld", function* xOld(state) {
    const str = state.get('str');
    yield () => do_not_optimize(asStringOld(str));
  }).gc("inner").args(args);
});

await run({
  format: "mitata",
  colors: true,
  throw: true,
});
```

gurgunday (approved):

mcollina:
@nigrosimone can you please include a revert of the revert of @climba03003 original commits? I've added tests for all listed regressions

Uzlopak (approved):
RSLGTM

To be honest, I cant follow the benchmarks. mitata is not the best benchmark out there. Actually tinybench is afaik the best benchmark out there, which is the reason it is used for benchmarks in nodejs.

If you post that master has 15.000 ops and this PR has 12.000 ops then this PR is slower. But as I said I cant read the mitata benchmarks. Or lets say, I dont want to dig deep into them.

Disclaimer: I am also maintainer of tinybench.

climba03003 (approved):
LGTM. Thanks for fixing the mistake.

mcollina (approved):
lgtm

gurgunday (approved):

cesco69:
Guys, this implementation seem slower...

gurgunday:
Oh I thought it was fixing something else actually, then yeah maybe let's keep it as is?

Uzlopak:
I am closing this, as this part of the code made already too much confusion and friction. 

We should first make some good benchmarks for strings, with various cases. E.g. I think one of the tests is repeated NUL-character (\u0000), which is basically pointless, because it should have the same performance like a string with length of 1 with one NUL character. So we should make some smart test cases and then balance the performance. 

Also using tinybench, as it is more correct than any other benchmark engine right now on the market. E.g. one of the tables  in benchmark.js is mathematically wrong. mitata is programmatically not correct.

cesco69:
Side note,
I’d like to propose the implementation of `JSON.isWellFormed()`  to TC39: https://es.discourse.group/t/json-iswellformed-str-string-boolean 

It will probably be very difficult for the committee to approve such a niche idea.
@mcollina since you’re a TSC member, you have the power!

PR #796, opened 2025-09-29
docs: remove flatstr suggestion

cesco69:
it's obsolete and unecesssary in modern nodejs


gurgunday (approved):
Seems true, but I didn't verify myself:

https://github.com/davidmarkclements/flatstr/issues/10#issue-860266996

PR #802, opened 2025-10-08
perf: avoid addComma when not necessary

cesco69:
Just a small performance improvements when an object contains only one key or an array only one item

Side note: the benchmark don't cover this case

### DIFF
--- a/index.js
+++ b/index.js
@@ -369,7 +369,8 @@ function buildInnerObject (context, location) {
   code += 'let json = JSON_STR_BEGIN_OBJECT\n'
 
   let addComma = ''
-  if (!hasRequiredProperties && (propertiesKeys.length > 1 || (schema.patternProperties || schema.additionalProperties))) {
+  const hasComma = !hasRequiredProperties && (propertiesKeys.length > 1 || (schema.patternProperties || schema.additionalProperties))
+  if (hasComma) {
     code += 'let addComma = false\n'
     addComma = '!addComma && (addComma = true) || (json += JSON_STR_COMMA)'
   }
@@ -387,14 +388,14 @@ function buildInnerObject (context, location) {
     code += `
       value = obj[${sanitizedKey}]
       if (value !== undefined) {
-        ${addComma}
+        ${i > 0 ? addComma : (hasComma ? 'addComma = true' : '')}
         json += ${JSON.stringify(sanitizedKey + ':')}
         ${buildValue(context, propertyLocation, 'value')}
       }`
 
     if (defaultValue !== undefined) {
       code += ` else {
-        ${addComma}
+        ${i > 0 ? addComma : (hasComma ? 'addComma = true' : '')}
         json += ${JSON.stringify(sanitizedKey + ':' + JSON.stringify(defaultValue))}
       }
       `
--- a/index.js
+++ b/index.js
@@ -369,7 +369,7 @@ function buildInnerObject (context, location) {
   code += 'let json = JSON_STR_BEGIN_OBJECT\n'
 
   let addComma = ''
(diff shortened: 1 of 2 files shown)


gurgunday (approved):

index.js
```diff
@@ -369,7 +369,7 @@ function buildInnerObject (context, location) {
   code += 'let json = JSON_STR_BEGIN_OBJECT\n'
 
   let addComma = ''
-  if (!hasRequiredProperties) {
+  if (!hasRequiredProperties && (propertiesKeys.length > 1 || (schema.patternProperties || schema.additionalProperties))) {
```
Uzlopak, line 372: additionalProperties is afaik default true, so you need to check if addtionalProperties is explicitly not set to false

Uzlopak (changes requested):
The approach is good, but additionalProperties can break your neck

index.js
cesco69: > additionalProperties is afaik default true
```js
console.log(schema.additionalProperties)
if (!hasRequiredProperties && (propertiesKeys.length > 1 || (schema.patternProperties || schema.additionalProperties))) {
```
log
```js
undefined
```

Uzlopak:
https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#rfc.section.10.3.2.3


> [10.3.2.3.](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#rfc.section.10.3.2.3) [additionalProperties](https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#additionalProperties)
> The value of "additionalProperties" MUST be a valid JSON Schema.
> 
> The behavior of this keyword depends on the presence and annotation results of "properties" and "patternProperties" within the same schema object. Validation with "additionalProperties" applies only to the child values of instance names that do not appear in the annotation results of either "properties" or "patternProperties".
> 
> For all such properties, validation succeeds if the child instance validates against the "additionalProperties" schema.
> 
> The annotation result of this keyword is the set of instance property names validated by this keyword's subschema.
> 
> **Omitting this keyword has the same assertion behavior as an empty schema.**
> 
> Implementations MAY choose to implement or optimize this keyword in another way that produces the same effect, such as by directly checking the names in "properties" and the patterns in "patternProperties" against the instance property set. Implementations that do not support annotation collection MUST do so.

And what does empty schema do?

https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-00#rfc.section.4.3.2

> The boolean schema values "true" and "false" are trivial schemas that always produce themselves as assertion results, regardless of the instance value. They never produce annotation results.
> 
> These boolean schemas exist to clarify schema author intent and facilitate schema processing optimizations. They behave identically to the following schema objects (where "not" is part of the subschema application vocabulary defined in this document).
> 
> true:
> **Always passes validation, as if the empty schema {}**
> false:
> Always fails validation, as if the schema { "not": {} }
> While the empty schema object is unambiguous, there are many possible equivalents to the "false" schema. Using the boolean values ensures that the intent is clear to both human readers and implementations.

In result, this means, that if additionalProperties is not set `additionalProperties: undefined`, it is equivalent to empty schema `additionalProperties: {}`, which is equivalent to setting it to true `additionalProperties: true`. 

So, to be compliant with json schema, you have to check that it is explicitly not false.

index.js
Uzlopak: This is not a qualified analysis to disprove my assessment, see my comment below.

cesco69:
done https://github.com/fastify/fast-json-stringify/pull/802/commits/a1fade334c05a666ef77ded5816535bc9196fe14

PR #803, opened 2025-10-09
feat: new benchmark with tinybench

cesco69:
TLTR: the current benchmark is not reliable
see https://github.com/fastify/fast-json-stringify/pull/801#issuecomment-3384792312 

In this PR have migrated the current benchmark based on benchmark.js ([abbandoned](https://github.com/bestiejs/benchmark.js), last update 7 years ago) to [tinybench](https://github.com/tinylibs/tinybench). 
I have also preserved the same output

PR (tinybench)
```
short string............................................. x 11,148,323 ops/sec ±96.84% (449942 runs sampled)
unsafe short string...................................... x 17,983,965 ops/sec ±0.41% (1559790 runs sampled)
short string with double quote........................... x 8,865,937 ops/sec ±7.80% (668640 runs sampled)
long string without double quotes........................ x 9,749 ops/sec ±35.67% (471 runs sampled)
```

MASTER (benchmark.js)
```
short string............................................. x 25,309,923 ops/sec ±2.39% (180 runs sampled)
unsafe short string...................................... x 95,582,481 ops/sec ±6.63% (157 runs sampled)
short string with double quote........................... x 16,337,301 ops/sec ±2.08% (182 runs sampled)
long string without double quotes........................ x 52,829 ops/sec ±2.25% (182 runs sampled)
```

### Benchmark compare: PR vs PR
image

### Benchmark compare: MASTER VS MASTER 
image

### Raw result patch-10 VS patch-10
```
Checking out "patch-10"
Execute "npm run bench"

> fast-json-stringify@6.1.1 bench
> node --expose-gc ./benchmark/bench.js

short string............................................. x 11,706,183 ops/sec ±0.51% (1490282 runs sampled)
unsafe short string...................................... x 13,448,873 ops/sec ±0.39% (1751476 runs sampled)
short string with double quote........................... x 9,955,676 ops/sec ±1.22% (1042841 runs sampled)
long string without double quotes........................ x 58,338 ops/sec ±2.10% (5234 runs sampled)
unsafe long string without double quotes................. x 14,098,104 ops/sec ±0.70% (1838793 runs sampled)
long string.............................................. x 32,270 ops/sec ±2.72% (2562 runs sampled)
unsafe long string....................................... x 15,952,407 ops/sec ±0.35% (2090631 runs sampled)
number................................................... x 16,597,008 ops/sec ±0.33% (2172467 runs sampled)
integer.................................................. x 12,844,829 ops/sec ±0.45% (1659085 runs sampled)
[...]

PR #809, opened 2026-01-12
chore: use default time value (100 ms is too low)

cesco69:
With https://github.com/fastify/fast-json-stringify/pull/803 I have misunderstanding the settings of `time` into tinybench. `time: 100` is 100 ms and is too low for a benchmark, the default time is `1000 ms`. This PR use the deafult value


mcollina (approved):
lgtm

PR #811, opened 2026-01-14
perf(serializer): avoid addComma when not necessary

cesco69:
Just a small performance improvements when an object contains only one key or an array only one item, the fjs code avoid to check to `addComma` 

Currently this schema
```js
fjs({
    title: 'additionalProperties',
    type: 'object',
    properties: {
      foo: {
        type: 'string'
      }
    },
    additionalProperties: false
})
```

produce this fjs code
```js
function anonymous0 (input) {
      const obj = (input && typeof input.toJSON === 'function')
    ? input.toJSON()
    : input

      let json = JSON_STR_BEGIN_OBJECT
let addComma = false

      const value_foo_0 = obj["foo"]
      if (value_foo_0 !== undefined) {
        !addComma && (addComma = true) || (json += JSON_STR_COMMA)
        json += "\"foo\":"

        if (typeof value_foo_0 !== 'string') {
          if (value_foo_0 === null) {
            json += JSON_STR_EMPTY_STRING
          } else if (value_foo_0 instanceof Date) {
            json += JSON_STR_QUOTE + value_foo_0.toISOString() + JSON_STR_QUOTE
          } else if (value_foo_0 instanceof RegExp) {
            json += asString(value_foo_0.source)
          } else {
            json += asString(value_foo_0.toString())
          }
        } else {
          json += asString(value_foo_0)
        }

      }

    return json + JSON_STR_END_OBJECT

    }
```
whith this PR
```JS
function anonymous0 (input) {
      const obj = (input && typeof input.toJSON === 'function')
    ? input.toJSON()
    : input

      let json = JSON_STR_BEGIN_OBJECT

      const value_foo_0 = obj["foo"]
      if (value_foo_0 !== undefined) {

        json += "\"foo\":"

        if (typeof value_foo_0 !== 'string') {
          if (value_foo_0 === null) {
            json += JSON_STR_EMPTY_STRING
          } else if (value_foo_0 instanceof Date) {
            json += JSON_STR_QUOTE + value_foo_0.toISOString() + JSON_STR_QUOTE
          } else if (value_foo_0 instanceof RegExp) {
            json += asString(value_foo_0.source)
          } else {
            json += asString(value_foo_0.toString())
          }
        } else {
          json += asString(value_foo_0)
        }

      }

    return json + JSON_STR_END_OBJECT

    }
```

Diff
image

Side note: the benchmark don't cover this case

### DIFF
--- a/index.js
+++ b/index.js
@@ -372,12 +372,14 @@ function buildInnerObject (context, location) {
   code += 'let json = JSON_STR_BEGIN_OBJECT\n'
 
   let addComma = ''
-  if (!hasRequiredProperties) {
+  const needComma = !hasRequiredProperties && (propertiesKeys.size > 1 || (schema.patternProperties || schema.additionalProperties !== false))
+  if (needComma) {
     code += 'let addComma = false\n'
     addComma = '!addComma && (addComma = true) || (json += JSON_STR_COMMA)'
   }
 
   let counterValue = 0
+  let i = 0
   for (const key of propertiesKeys) {
     let propertyLocation = propertiesLocation.getPropertyLocation(key)
     if (propertyLocation.schema.$ref) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #815, opened 2026-01-16
chore: remove RunKit (is dead)

cesco69:
Maybe https://stackblitz.com/ is a valid alternative


mcollina:
Stackblitz is not running Node.js, but their private fork.

PR #818, opened 2026-01-18
benchmark: change date format from 'date' to 'datetime'

nigrosimone:
Ohter libs serialize in "datetime" format. 
Also, the "dat"e format is slower in fjs


mcollina (approved):
lgtm

PR #823, opened 2026-01-29
chore: update tinybench to v6 and devDependencies

nigrosimone:
update tinybench to v6 and also update all devDependencies


gurgunday (approved):
lgtm

PR #837, opened 2026-04-25
refactor(types): migrate from tsd to tstyche

Tony133:
Proposal:

- migrate from `tsd` to `tstyche`

PR of Reference: 
  - https://github.com/fastify/fastify/pull/6532
  - https://github.com/fastify/fastify/pull/6525

### DIFF
--- a/types/index.tst.ts
+++ b/types/index.tst.ts
@@ -1,7 +1,7 @@
 // eslint-disable-next-line @typescript-eslint/no-unused-vars -- Test using this disabled, see https://github.com/fastify/fast-json-stringify/pull/683
 import Ajv from 'ajv'
 import build, { restore, Schema, validLargeArrayMechanisms } from '..'
-import { expectError, expectType } from 'tsd'
+import { expect } from 'tstyche'
 
 // Number schemas
 build({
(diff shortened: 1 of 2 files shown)


types/index.tst.ts
```diff
@@ -191,11 +190,12 @@ const debugCompiled = build({
     }
   }
 }, { mode: 'debug' })
-expectType<ReturnType<typeof build>>(build.restore(debugCompiled))
-expectType<ReturnType<typeof build>>(restore(debugCompiled))
 
-expectType<string[]>(build.validLargeArrayMechanisms)
-expectType<string[]>(validLargeArrayMechanisms)
+expect(build.restore(debugCompiled)).type.toBe<ReturnType<typeof build>>()
+expect(restore(debugCompiled)).type.toBe<ReturnType<typeof build>>()
```
mrazauskas: For overloaded functions (like `build()`), `ReturnType` picks only the last signature ([reference](https://www.typescriptlang.org/docs/handbook/utility-types.html#returntypetype)). That is not obvious for a reader. I would suggest comparing with that call signature directly:

```suggestion
expect(build.restore(debugCompiled)).type.toBe(build({} as Schema))
expect(restore(debugCompiled)).type.toBe(build({} as Schema))
```

mrazauskas:
lgtm

gurgunday (approved):
lgtm

PR #838, opened 2026-04-30
fix: extract reusable functions for non-recursive external $ref schemas

mcollina:
## Problem

When a non-recursive schema registered via external `$ref` (with a `$id`) is referenced multiple times, `fast-json-stringify` inlines the full serialization code at every reference point instead of extracting a reusable function.

This causes exponential code generation in schemas with repeated `$ref` targets, especially when combined with `anyOf`/`oneOf` wrapping — a single serializer can generate 48+ MB of code.

## Root Cause

In `buildObject()` and `buildArray()`, the function extraction path only triggers when:
1. `recursivePaths.has(fullPath)` — schema references itself (circular)
2. `buildingSet.has(schema)` — schema is currently being built (re-entrant)

For non-recursive external schemas, neither condition is true, so the code falls through to the inline path. The `functionsNamesBySchema` cache is never populated.

## Fix

Add `schemaId !== ''` as a third condition to trigger function extraction in both `buildObject()` and `buildArray()`. This reuses the same mechanism already used for recursive schemas: extract into a named function on first encounter, then reuse via `functionsNamesBySchema` on subsequent encounters.

## Tests

Added 5 new tests in `test/issue-836.test.js`:
- External `$ref` schema reuse (direct references)
- External `$ref` schema reuse with `anyOf` wrapping
- External `$ref` schema reuse in array items
- External array `$ref` schema reuse
- Regression guard: inline anonymous schemas still inlined normally

All 473 existing tests pass, TypeScript types pass.

### DIFF
--- a/index.js
+++ b/index.js
@@ -568,7 +568,7 @@ function buildObject (context, location, input) {
   const jsonPointer = location.jsonPointer || ''
   const fullPath = `${schemaId}#${jsonPointer}`
 
-  if (context.recursivePaths.has(fullPath) || context.buildingSet.has(schema)) {
+  if (context.recursivePaths.has(fullPath) || context.buildingSet.has(schema) || schemaId !== '') {
     const functionName = generateFuncName(context)
     context.functionsNamesBySchema.set(schema, functionName)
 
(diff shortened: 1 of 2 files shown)


jsumners (commented):
I do not have enough knowledge of this code base to properly review. It looks like a simple change, but I'm not sure.

Eomm (commented):
Checked here https://github.com/fastify/fast-json-stringify/pull/836 and it was legit 👍🏼 

Thanks for the test

Eomm (approved):

ivan-tymoshenko (approved):
Lgtm

PR #841, opened 2026-05-19
fix: avoid stray comma when required is set but properties is empty

imalbi:
## Problem

A response schema with `required` set but no `properties` map (e.g. what `z.toJSONSchema` produces for `z.record(enum, V)`) serializes with a stray leading comma:

```js
const schema = {
  type: 'object',
  properties: {
    obj: {
      type: 'object',
      propertyNames: { type: 'string', enum: ['a', 'b'] },
      additionalProperties: { type: 'number' },
      required: ['a', 'b']
    }
  }
}
build(schema)({ obj: { a: 1, b: 2 } })
// {"obj":{,"a":1,"b":2}}
```

HTTP 200 from Fastify, but `JSON.parse` fails client-side.

## Root cause

The optimization at `index.js:389` skips the runtime comma check when `requiredProperties.length > 0`, on the assumption that the first declared property will anchor the comma logic. When `properties` is empty (or absent), the loop body never runs and `addComma` is set to unconditional immediately after, so the `additionalProperties` branch at `index.js:484` emits a separator before its first entry.

## Fix

Tighten the condition to `propertiesKeys.length > 0 && requiredProperties.includes(propertiesKeys[0])`. The sort just above (`index.js:367-373`) moves required-and-declared keys to the front, so this captures exactly when the "first declared property anchors the comma" premise actually holds. It falls through to the existing runtime `addComma` flag path in three cases that the old condition handled incorrectly:

1. `properties` is empty (record-style schema, only `additionalProperties`)
2. `required` only names keys that are not in `properties`
3. (Trivially) `requiredProperties` is empty — already handled before

No behaviour change for any schema where at least one declared property is required.

## Tests

Three regression tests covering the canonical bug shapes:

- `test/additionalProperties.test.js`: `required + additionalProperties` with no `properties` map (the original z.record-style trigger).
- `test/additionalProperties.test.js`: `required` key not in `properties`, with `additionalProperties: true`.
- `test/patternProperties.test.js`: same shape with `patternProperties` (symmetric branch at `index.js:484`).

Each test fails on the respective baseline (`main` for #1, the first proposed fix for #2 and #3) and passes after this patch. Full suite: 476/476 green.

## Benchmark

Ran `npm run benchmark` 3 times on `main` and 3 times on the fix branch. All scenario ranges overlap; absolute mean deltas ≤ 1.1%, within natural inter-run variance (±0.7% to ±2.0%). The modified branch is not exercised by any standard scenario, so no runtime perf change is expected by construction.

After tightening the guard further (to also handle `required` keys not in `properties`), re-ran 3 vs 3 between the first fix and the tightened version. Absolute mean deltas ≤ 1.7%, still within inter-run variance. By transitivity, the tightened fix is also performance-neutral vs `main`.

## Related

- #642 / #651: same class of bug (extra-comma in object serialization), different branch.
- #530: proposed dropping `required` support entirely. This fix is a smaller targeted patch in the same area while that broader change is on hold.

---

## Checklist 

### DIFF
--- a/index.js
+++ b/index.js
@@ -386,23 +386,9 @@ function buildInnerObject (context, location, objVar) {
   const localUid = context.uid++
   let addComma = ''
 
-  // The skip-leading-comma optimization is only safe when at least one
-  // declared property is also required: that guarantees the first emitted
-  // entry has no preceding separator. `propertiesKeys` is sorted with
-  // required keys first, so `requiredProperties.includes(propertiesKeys[0])`
-  // is the cheapest way to express the condition.
-  //
-  // Falls through to the generic runtime-`addComma` branch when:
-  //   * `requiredProperties` is empty (no anchor needed), or
-  //   * `propertiesKeys` is empty (record-style schema with only
-  //     `additionalProperties`), or
-  //   * `required` only names keys that are not in `properties` (so no
-  //     declared property is guaranteed to emit first).
-  // Otherwise the additionalProperties / patternProperties serializer
-  // would write a separator with no preceding property, producing
-  // `{ ,"k":v,... }`.
+  // propertiesKeys is sorted required-first; the guard checks [0] is required because
+  // otherwise additionalProperties/patternProperties would emit a stray `{ ,"k":v }`.
   if (propertiesKeys.length > 0 && requiredProperties.includes(propertiesKeys[0])) {
-
     // The first property is required, so we don't need a comma.
     // For the subsequent properties, we can blindly add a comma.
 
--- a/index.js
+++ b/index.js
@@ -386,16 +386,22 @@ function buildInnerObject (context, location, objVar) {
   const localUid = context.uid++
   let addComma = ''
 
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -386,9 +386,22 @@ function buildInnerObject (context, location, objVar) {
...
+  //
+  // Falls through to the generic runtime-`addComma` branch when:
+  //   * `requiredProperties` is empty (no anchor needed), or
+  //   * `propertiesKeys` is empty (record-style schema with only
+  //     `additionalProperties`), or
+  //   * `required` only names keys that are not in `properties` (so no
+  //     declared property is guaranteed to emit first).
+  // Otherwise the additionalProperties / patternProperties serializer
+  // would write a separator with no preceding property, producing
+  // `{ ,"k":v,... }`.
```
mcollina: This comment is gargantuan. Shorten it.

imalbi: done

mcollina (approved):
lgtm

PR #845, opened 2026-06-11
Fix percent-encoded refs in ajv validation paths

### LINKED ISSUE #740: Error using stringify on schema with uri-encoding definition property containing anyOf, allOf or oneOf
### Prerequisites

### Fastify version

Latest version

### Plugin version

6.0.0

### Node.js version

20.1.0

### Operating system

Linux

22.04.4 LTS

### Description

While using stringify I started getting "can't resolve reference" error.
It seems that when you have a definition property using a uri-encoding like '%3' or '%22', and this property has fields contaning oneOf, allOf or anyOf, it returns the error.

The following code can be run on the fast-json-stringify test suite to confirm the error.

```javascritpt
'use strict'

const { test } = require('tap')
const build = require('..')

process.env.TZ = 'UTC'

test('ref internal - properties', (t) => {
  t.plan(1)

  const schema = {
    title: 'object with $ref',
    definitions: {
      'Some%3Cloremipsum%3E': {
          additionalProperties: {
            oneOf: [
              { type: 'string' },
              { type: 'number' },
              { type: 'object' },
              { type: 'null' }
            ]
          },
          type: 'object'
      }
    },
    type: 'object',
    properties: {
      obj: {
        $ref: '#/definitions/Some%3Cloremipsum%3E'
      }
    }
  }

  const object = {
    obj: {
      str: 'test'
    }
  }

  const output = stringify(object)

  JSON.parse(output)
  t.pass()

  t.equal(output, '{"obj":{"str":"test"}}')
})
```

The error exhibited is

` FAIL  test/anyof.test.js 1 failed of 1 1.183s
 ✖ ref internal - properties > can't resolve reference
   #/definitions/Some%3Cloremipsum%3E from id __fjs_root_0
    node_modules/ajv/lib/vocabularies/core/ref.ts                            
     16     const {root} = env                                               
     17     if (($ref === "#" || $ref === "#/") && baseId === root.baseId) 
    return callRootRef()
     18     const schOrEnv = resolveRef.call(self, root, baseId, $ref)       
     19     if (schOrEnv === undefined) throw new 
    MissingRefError(it.opts.uriResolver, baseId, $ref)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛                              
     20     if (schOrEnv instanceof SchemaEnv) return callValidate(schOrEnv) 
     21     return inlineRefSchema(schOrEnv)                                 
     22                                                                      
     23     function callRootRef(): void {                                   
    missingRef: __fjs_root_0#/definitions/Some%3Cloremipsum%3E
    missingSchema: __fjs_root_0
    tapCaught: testFunctionThrow
    Object.code (node_modules/ajv/lib/vocabularies/core/ref.ts:19:39)
    keywordCode (node_modules/ajv/lib/compile/validate/index.ts:532:9)
    <anonymous> (node_modules/ajv/lib/compile/validate/index.ts:228:21)
    CodeGen.code (node_modules/ajv/lib/compile/codegen/index.ts:545:33)
    CodeGen.block (node_modules/ajv/lib/compile/codegen/index.ts:700:20)
    schemaKeywords (node_modules/ajv/lib/compile/validate/index.ts:228:9)
    typeAndKeywords (node_modules/ajv/lib/compile/validate/index.ts:161:3)
    subSchemaObjCode (node_modules/ajv/lib/compile/validate/index.ts:147:3)
    subschemaCode (node_modules/ajv/lib/compile/validate/index.ts:124:7)
    KeywordCxt.subschema 
        (node_modules/ajv/lib/compile/validate/index.ts:500:5)
`

Removing the '%' on the definition property OR removing oneOf like in the following code returns no error.
```javascript
'Some%3Cloremipsum%3E': {
    additionalProperties: {
         type: 'string' ,
    },
    type: 'object'
}
```

### Link to code that reproduces the bug

_No response_

### Expected Behavior

_No response_

morgan-coded:
`fast-json-stringify` could compile a schema with percent-encoded definition keys, then crash on the first `stringify()` when a validation branch went through ajv.
The serializer resolves those JSON-pointer keys literally, while ajv decodes URI fragments before lookup. This keeps the existing literal-key behavior and re-encodes `%` only for refs handed to ajv, including the schema clone used by standalone mode.
I tested the issue repro plus the new branch and standalone regressions. The unit suite passes with 478 tests, tstyche passes 12/12, and lint plus `git diff --check` are clean.
Closes #740

### DIFF
--- a/test/issue-740.test.js
+++ b/test/issue-740.test.js
@@ -1,11 +1,10 @@
 'use strict'
 
-const { test, after } = require('node:test')
+const { test } = require('node:test')
 const fs = require('fs')
+const os = require('os')
 const path = require('path')
 const build = require('..')
-
-process.env.TZ = 'UTC'
 
 // https://github.com/fastify/fast-json-stringify/issues/740
 // Schema keys containing percent-encoded sequences (e.g. '%3C') are resolved
@@ -109,11 +108,17 @@ test('ref to a percent-encoded definition key with oneOf in standalone mode', as
 
   const code = build(buildIssueSchema(oneOfComposition), { mode: 'standalone' })
 
-  const destination = path.resolve('test/fixtures', 'standalone-issue-740.js')
+  // The standalone output does `require('fast-json-stringify/lib/...')`, so the
+  // generated file must live where that package resolves. Use a per-run temp
+  // dir and link node_modules into it so nothing is written into the repo tree.
+  const dir = await fs.promises.mkdtemp(path.join(os.tmpdir(), 'fjs-issue-740-'))
 
-  after(async () => {
-    await fs.promises.rm(destination, { force: true })
+  t.after(async () => {
+    await fs.promises.rm(dir, { recursive: true, force: true })
   })
+
+  await fs.promises.symlink(path.resolve(__dirname, '..', 'node_modules'), path.join(dir, 'node_modules'), 'junction')
+  const destination = path.join(dir, 'standalone-issue-740.js')
 
   await fs.promises.writeFile(destination, code)
   const standalone = require(destination)
--- a/test/issue-740.test.js
+++ b/test/issue-740.test.js
@@ -1,0 +1,122 @@
+'use strict'
+
+const { test, after } = require('node:test')
+const fs = require('fs')
+const path = require('path')
+const build = require('..')
+
+process.env.TZ = 'UTC'
+
+// https://github.com/fastify/fast-json-stringify/issues/740
+// Schema keys containing percent-encoded sequences (e.g. '%3C') are resolved
+// literally by fast-json-stringify, but the refs handed to ajv were not
+// re-encoded, so ajv percent-decoded them and failed to find the keys.
+
+function buildIssueSchema (composition) {
+  return {
+    title: 'object with $ref',
+    definitions: {
+      'Some%3Cloremipsum%3E': {
+        additionalProperties: composition,
+        type: 'object'
+      }
+    },
+    type: 'object',
+    properties: {
+      obj: {
+        $ref: '#/definitions/Some%3Cloremipsum%3E'
+      }
+    }
+  }
+}
+
+const oneOfComposition = {
+  oneOf: [
+    { type: 'string' },
+    { type: 'number' },
+    { type: 'object' },
+    { type: 'null' }
+  ]
+}
+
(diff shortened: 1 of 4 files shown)


test/issue-740.test.js
```diff
@@ -0,0 +1,120 @@
...
+  t.assert.equal(stringify({ 'weird%name': 'value' }), '{"weird%name":"value"}')
+  t.assert.equal(stringify({ 'weird%name': 42 }), '{"weird%name":42}')
+})
+
+test('ref to a percent-encoded definition key with oneOf in standalone mode', async (t) => {
+  t.plan(1)
+
+  const code = build(buildIssueSchema(oneOfComposition), { mode: 'standalone' })
+
+  const destination = path.resolve('test/fixtures', 'standalone-issue-740.js')
```
mcollina: Can you use a temp folder instead?

mcollina (changes requested):

test/issue-740.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+const { test } = require('node:test')
+const fs = require('fs')
+const path = require('path')
+const build = require('..')
+
+// https://github.com/fastify/fast-json-stringify/issues/740
+// Schema keys containing percent-encoded sequences (e.g. '%3C') are resolved
+// literally by fast-json-stringify, but the refs handed to ajv were not
+// re-encoded, so ajv percent-decoded them and failed to find the keys.
```
Eomm, line 12: I want to validate this

Eomm (changes requested):

test/issue-740.test.js
morgan-coded: I checked the Ajv behavior behind that note. The schema key is literal (`Some%3Cloremipsum%3E`), but Ajv decodes the ref fragment while resolving it, so `%3C`/`%3E` can become `<`/`>` and miss the literal key.

That’s why the fix goes through `encodeRefFragment`: it changes `%` to `%25` before Ajv sees the ref, then Ajv’s decode step lands back on the literal `%` key. The focused issue/standalone run passes 10/10, covering `oneOf`, `anyOf`, `if/then/else`, and generated standalone output.

mcollina: @Eomm PTAL

PR #852, opened 2026-07-08
perf: faster object property and date/time serialization

mcollina:
- Fold commas (and the leading `{` when a required property comes first) into the property key literals, so each property is serialized with a single string append instead of two or three.
- Format `date-time`, `date` and `time` manually instead of via `toISOString` and intermediate `Date` allocations (~4.5x, ~16x and ~17x respectively). Expanded years and invalid dates fall back to the previous path.
- Replace the bound `asInteger` with a closure.

Benchmarks (Node 24): formatted time +1605%, formatted date +1493%, date-time +355%, objects +7-15%, arrays of objects +15%. No regressions.

Also fixes invalid JSON output (`{,"b":2}`) when `required` lists a key not declared in `properties` and the first declared property is undefined; regression test added.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,6 +16,7 @@ let largeArrayMechanism = 'default'
 const serializerFns = `
 const {
   asString,
+  asInteger,
   asNumber,
   asBoolean,
   asDateTime,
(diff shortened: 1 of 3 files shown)


Tony133:
there are some conflicts in the `index.js`

nigrosimone:
Perhaps the readme.md needs to be updated as well https://github.com/fastify/fast-json-stringify#benchmarks

PR #853, opened 2026-07-15
perf: optimize Ajv refs for external schema branches

RelevantShannon:
Redirect buildOneOf's Ajv ref to a registered ref target when possible.

buildOneOf currently validates oneOf/anyOf branches through location.getSchemaRef(), which points back into the generated root schema even when the branch is only an external ref already registered with Ajv. This lets those branches validate against the exact external ref target instead, so Ajv can use the already-registered schema path rather than resolving through the root document.

The helper is intentionally guarded:
- only pure  branches are redirected
- inline branches and refs with sibling keywords are left unchanged
- unresolved refs are left unchanged
- JSON-pointer and supported fragment refs are redirected only when the resolver can resolve the exact target
- relative root  cases are only redirected when the fragment matches the registered schema id shape

This also applies the same validator-ref selection to if schemas, which use the same runtime validator path.

Verification:
- npm run lint
- git diff --check
- node --test test/oneof.test.js
- node --test test/oneof.test.js test/anyof.test.js test/if-then-else.test.js
- npm run test:unit

npm run test:unit passed locally with 488 tests and 0 failures. The local run printed Node inspector auto-attach warnings, but exited successfully.

### DIFF
--- a/index.js
+++ b/index.js
@@ -100,6 +100,51 @@ function getSafeSchemaRef (context, location) {
   return schemaRef
 }
 
+function getValidatorSchemaRef (context, location) {
+  let schemaRef = location.getSchemaRef()
+  const schema = location.schema
+
+  // A schema that is only a registered $ref can validate against that target
+  // instead of a JSON-pointer path through the root schema.
+  if (
+    typeof schema.$ref !== 'string' ||
+    Object.keys(schema).length !== 1
+  ) {
+    return schemaRef
+  }
+
+  const ref = schema.$ref
+  const hashIndex = ref.indexOf('#')
+  const refSchemaId = hashIndex === -1 ? ref : ref.slice(0, hashIndex)
+  const refJsonPointer = hashIndex === -1 ? '' : ref.slice(hashIndex)
+  const isBareRef = refJsonPointer === '' || refJsonPointer === '#'
+  const isNamedFragmentRef = /^#[A-Za-z_][-A-Za-z0-9._]*$/.test(refJsonPointer)
+  const isJsonPointerRef = refJsonPointer.startsWith('#/')
+
+  if (
+    (isBareRef || isNamedFragmentRef || isJsonPointerRef) &&
+    refSchemaId &&
+    context.refResolver.hasSchema(refSchemaId)
+  ) {
+    const refSchema = context.refResolver.getSchema(refSchemaId)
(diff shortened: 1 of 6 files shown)


mcollina (approved):
lgtm

Fdawgs (commented):
Any supporting benchmarks for this optimisation?

RelevantShannon:
Added a benchmark covering the affected path: cold `fast-json-stringify` compile plus first `stringify()` for a `oneOf` field whose branches are external registered `$ref` schemas.

Local before/after using the same benchmark workload:

| Branch | Result |
| --- | ---: |
| `main` | 462.0 ops/sec |
| this PR | 1,529.3 ops/sec |

That works out to roughly:

- `3.31x` faster
- `+231%` throughput
- `~69.8%` less time per compile + first stringify
- `2.16ms` -> `0.65ms` per operation

I also ran the full benchmark suite locally after adding the benchmark, and the new case reported:

```text
oneOf external refs compile and first stringify x 1,584 ops/sec ±1.09% (1534 runs sampled)
```

Tony133 (approved):
lgtm

index.js
```diff
@@ -100,6 +100,51 @@ function getSafeSchemaRef (context, location) {
...
+  ) {
+    return schemaRef
+  }
+
+  const ref = schema.$ref
+  const hashIndex = ref.indexOf('#')
+  const refSchemaId = hashIndex === -1 ? ref : ref.slice(0, hashIndex)
+  const refJsonPointer = hashIndex === -1 ? '' : ref.slice(hashIndex)
+  const isBareRef = refJsonPointer === '' || refJsonPointer === '#'
+  const isNamedFragmentRef = /^#[A-Za-z_][-A-Za-z0-9._]*$/.test(refJsonPointer)
```
Fdawgs: The regex pattern could be cached at module level ala https://github.com/fastify/fast-uri/blob/5e31a6b0f6408f79d7c23e8490b092aba4d873c5/lib/schemes.js#L4, which we do everywhere to save the GC some bother.

Likewise could be cut down to `/^#[a-z_][-\w._]*$/i` to simplify it.

RelevantShannon: Yep, sounds good. I've gone ahead and done that.

Fdawgs (approved):

PR #854, opened 2026-07-17
feat: inline validators in standalone output

RelevantShannon:
Add an opt-in mode that embeds Ajv's generated validators into fast-json-stringify standalone output.

Standalone mode currently restores an Ajv-backed validator from serialized schema state. Schemas using `anyOf`, `oneOf`, or `if/then/else` therefore still compile their branch validators at runtime. The new `inlineValidators` option asks Ajv's standalone compiler for the exact fragment validators used by the generated serializer and includes them in the same output module.

The inline path is intentionally scoped:
- it is only enabled for standalone mode when `inlineValidators` is true
- only schema fragments referenced by generated validator calls are emitted
- schemas that do not use Ajv-backed branches emit no validator code
- omitting or disabling the option preserves the existing state-restoration path
- generated validator code is forced to CommonJS while other caller-provided Ajv code options are preserved

This also makes the internal `fjs_type` keyword code-generating so Date and other `toJSON` values remain compatible with Ajv standalone output. External schema refs and built-in formats are included through Ajv's normal standalone dependency handling.

The output remains one generated module, but still uses fast-json-stringify's standalone serializer dependency and may reference small Ajv runtime helpers when required by the generated validation code.

Verification:
- npm run lint
- git diff --check
- node --test test/standalone-mode.test.js
- npm test

npm test passed locally with 504 tests, 100% coverage, and 0 failures.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,7 @@ const stringify = fastJson(mySchema, {
 - `schema`: external schemas references by $ref property. [More details](#ref)
 - `ajv`: [ajv v8 instance's settings](https://ajv.js.org/options.html) for those properties that require `ajv`. [More details](#anyof)
 - `rounding`: setup how the `integer` types will be rounded when not integers. [More details](#integer)
+- `inlineValidators`: when using standalone mode, embed Ajv-generated validator functions in the output instead of compiling schemas at runtime. [More details](#standalone)
 - `largeArrayMechanism`: set the mechanism that should be used to handle large
 (by default `20000` or more items) arrays. [More details](#largearrays)
 - `compileValidators`: when `true`, the `ajv` validators used by `anyOf`, `oneOf` and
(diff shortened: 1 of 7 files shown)


mcollina:
This is awesome. Why is it still a draft?

RelevantShannon:
> This is awesome. Why is it still a draft?

Ah sorry, it was a Friday, and I wasn't sure if it was ready yet. I think it is :D

mcollina (approved):
lgtm

mcollina:
This needs conflicts to be resolved

Tony133:
@RelevantShannon As soon as you have time, can you resolve conflicts?

RelevantShannon:
@Tony133, conflicts resolved.

Tony133 (approved):

PR #860, opened 2026-07-29
fix: escape string values for date/date-time/time formats

mcollina:
## Problem

`asDateTime`, `asDate` and `asTime` concatenate string inputs straight between quotes:

```js
if (typeof date === 'string') {
  return '"' + date + '"'
}
```

So a string value containing `"`, `\`, a control character or a lone surrogate produces output that is not valid JSON:

```js
const stringify = fastJson({
  type: 'object',
  properties: { ts: { type: 'string', format: 'date-time' } }
})

stringify({ ts: 'a"b' })   // {"ts":"a"b"}  -> JSON.parse throws
```

Plain `type: string` fields already go through `asString`, so whether a string got escaped depended on `format`, which is surprising.

## Fix

- Hoist the escaping routine to module scope as `asString(str)`; the class method delegates to it. This is needed because the generated code destructures the serializer methods and calls them unbound, so `this.asString` isn't reachable from `asDateTime` & co.
- Route the string branch of `asDateTime` / `asDate` / `asTime` through it.
- `Date` inputs keep the plain concatenation — `toISOString()` output can never need escaping.
- `format: 'unsafe'` is untouched; it remains the documented way to opt out of escaping.

## Notes

- Serializing a date-format **string** now costs the same as serializing a regular string of the same length (roughly 10ns -> 37ns per field in a tight loop for a 24-char ISO timestamp). Unavoidable if the output has to be valid JSON. **`Date` values are unaffected.**
- Values are still not reformatted or validated as dates, only escaped. README note updated to say so.

## Test

New case in `test/date.test.js` asserts all three formats round-trip quotes, backslashes, control characters and lone surrogates identically to `JSON.stringify`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TmhA1oTx5qUWc6UmEgWuar


mcollina:
The real reason for this is so we stop receiving reports.

Fdawgs (approved):

PR #864, opened 2026-08-07
Add option to precompile AJV schemas to avoid lazy-compilation hit on first request

rkday-pro:
I've been tracking down an issue in my Fastify application where calls to a certain API take 3-4 seconds on the first hit after application startup, and then are fine.

The problem is that that API has a complicated schema with various oneOf entries, so on the first hit to that route we're doing a lot of AJV schema compilation.

This PR lets me move that work off the first-request path, and pass in an option to precompile those schemas at startup. By default, the behaviour is unchanged.


mcollina (approved):
lgtm

ilteoood (approved):

PR #867, opened 2026-08-08
fix: apply sibling keywords when a schema resolves a $ref

### LINKED ISSUE #866: Sibling required keyword is dropped when a property uses $ref to an external schema
**What's wrong**

When a property both references an external schema with `$ref` and adds its own
sibling keywords (like `required`), the serializer throws away those sibling
keywords. An object that is missing a required field is accepted and serialized
as valid instead of raising a validation error, so incomplete data passes
through silently.

**Where**

`index.js`, in `buildValue`:
https://github.com/fastify/fast-json-stringify/blob/7fad3c2d6305/index.js#L1300-L1303

```javascript
if (schema.$ref) {
  location = resolveRef(context, location)
  schema = location.schema
}
```

Resolving the reference replaces the local schema wholesale, so any sibling
keywords on the referencing schema (for example `required`) are gone before the
serializer is generated. The `getValidatorSchemaRef` guard added recently
(`index.js:113`, treats a schema with more than one key as non-pure) correctly
protects validator reference selection in `buildOneOf`/`buildIfThenElse`, but it
does not stop `buildValue` from dereferencing and discarding siblings.

**How it manifests**

1. Register an external `Foo` schema that requires `value`.
2. Create a root schema with a property `x` that uses `$ref: Foo` plus a sibling
   `required: ['extra']`.
3. Build the serializer with `Foo` registered as an external schema.
4. Serialize `{ x: { value: 1, extra: "x" } }` and then `{ x: { value: 1 } }`.

Both inputs serialize successfully, even though the second is missing the
sibling-required `extra`.

**Expected behavior**

The sibling `required` constraint should still apply. When a `$ref` has sibling
keywords, resolve the reference as the base schema and then apply the sibling
constraints on top, keeping the direct-dereference optimization only for a
schema that contains `$ref` alone.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs. Full analysis: https://app.ito.ai/share/29ad67b0-500d-45d1-9ffd-df265504fb5f.

lazerg:
## Problem

A property that both references another schema and adds its own keywords loses those keywords. `resolveRef` swaps the whole location for the reference target, so anything sitting next to `$ref` is gone before any code is generated:

```js
const stringify = build({
  type: 'object',
  properties: {
    x: { $ref: 'Foo#', required: ['extra'] }
  }
}, {
  schema: {
    Foo: {
      $id: 'Foo',
      type: 'object',
      properties: { value: { type: 'integer' }, extra: { type: 'string' } },
      required: ['value']
    }
  }
})

stringify({ x: { value: 1 } })
// ACTUAL:   {"x":{"value":1}}
// EXPECTED: throws '"extra" is required!'
```

`value` is enforced because it comes from `Foo`, `extra` is not, so incomplete data serializes as if it were valid.

## Fix

A `$ref` that carries siblings now gets the resolved target and those siblings merged through `mergeLocations`, the same path `allOf` already takes. Everything else keeps the direct dereference it always had:

- siblings that cannot change the output (`title`, `description`, `$comment`, `examples`, `deprecated`, `readOnly`, `writeOnly`) are skipped, so annotated refs still share one serializer
- when the sibling keywords and the target genuinely conflict (say a sibling `type` the target contradicts) the merge is dropped and the plain target is used, which is what happens today

Merged refs are cached by content rather than by object identity, since a merge clones its input and a recursive `$ref` would otherwise be merged again on every level.

## Test

Two cases in `test/ref.test.js`, internal and external `$ref` with a sibling `required`: the sibling constraint is enforced and the target's own `required` still is. Both fail on the current main.

Fixes #866

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@
 /* eslint no-prototype-builtins: 0 */
 
 const { RefResolver } = require('json-schema-ref-resolver')
+const { MergeError } = require('@fastify/merge-json-schemas')
 
 const Serializer = require('./lib/serializer')
 const Validator = require('./lib/validator')
@@ -119,7 +120,8 @@ function resolveRef (context, location) {
     mergedSchemaId = `__fjs_merged_${schemaIdCounter++}`
     try {
       mergeLocations(context, mergedSchemaId, [newLocation, siblingLocation])
-    } catch {
+    } catch (err) {
+      if (!(err instanceof MergeError)) throw err
       return newLocation
     }
     context.mergedRefsIds.set(mergedSchemaKey, mergedSchemaId)
--- a/index.js
+++ b/index.js
@@ -42,6 +42,19 @@ const validLargeArrayMechanisms = new Set([
   'json-stringify'
 ])
 
+// Keywords that do not change the output, so a $ref carrying only these
+// can still be dereferenced directly instead of merged with its target.
+const IGNORED_REF_SIBLING_KEYWORDS = new Set([
+  '$ref',
+  '$comment',
+  'title',
+  'description',
+  'examples',
+  'deprecated',
+  'readOnly',
+  'writeOnly'
+])
+
 let schemaIdCounter = 0
 
 function isValidSchema (schema, name) {
@@ -59,7 +72,8 @@ function isValidSchema (schema, name) {
 }
 
 function resolveRef (context, location) {
-  const ref = location.schema.$ref
+  const refSchema = location.schema
+  const ref = refSchema.$ref
 
   let hashIndex = ref.indexOf('#')
   if (hashIndex === -1) {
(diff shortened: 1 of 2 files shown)


lazerg:
Hey @Tony133, could you take a look at this fix for sibling keywords not applying once a schema resolves a $ref? It's been open a month with no feedback.

index.js
```diff
@@ -74,12 +88,44 @@ function resolveRef (context, location) {
...
+  // Keyed by content, not by identity: a merged schema is cloned on every
+  // merge, so a recursive $ref would otherwise be merged again on each level.
+  const mergedSchemaKey = newLocation.getSchemaRef() + JSON.stringify(siblingLocation.schema)
+
+  let mergedSchemaId = context.mergedRefsIds.get(mergedSchemaKey)
+  if (mergedSchemaId === undefined) {
+    mergedSchemaId = `__fjs_merged_${schemaIdCounter++}`
+    try {
+      mergeLocations(context, mergedSchemaId, [newLocation, siblingLocation])
+    } catch {
```
Tony133: Catch only MergeError here, rethrow everything else.

index.js
```diff
@@ -74,12 +88,44 @@ function resolveRef (context, location) {
...
+    return newLocation
+  }
+
+  const siblingLocation = new Location(
+    cloneOriginSchema(context, siblingSchema, location.schemaId),
+    location.schemaId,
+    location.jsonPointer
+  )
+
+  // Keyed by content, not by identity: a merged schema is cloned on every
```
Tony133, line 114: This comment describes the recursive case but no test covers it. Please add one, e.g. a `Node` schema with `next: { $ref: 'Node#', required: ['v'] }`.

Tony133 (changes requested):
Thanks, and sorry for the delay. The fix looks correct.

Two small changes before merging, see inline comments.

index.js
lazerg: Fixed in 84e2bea. The catch now tests `instanceof MergeError` and rethrows everything else. `mergeLocations` also calls `resolveRef` and `addSchema`, so a missing reference was swallowed here before.

index.js
lazerg: Added in 73d0a22 as `ref external - recursive sibling keywords`. I used the external-schema form of your `Node` example, because the plain internal one passes even without the cache and proves nothing. The new test overflows the stack when I drop the content key.

Tony133:
CI is red

lazerg:
CI failed on the 100% coverage gate, not on a failing test. The catch in `resolveRef` that falls back to the plain `$ref` target had no test behind it. I added two: one where a sibling keyword conflicts with the target, one where the merge throws an error that must not be swallowed. Fixed in b19f087.

Tony133 (approved):
lgtm

PR #868, opened 2026-08-09
perf: fold codegen literals and inline nested object serialization

mcollina:
## What

Speeds up the generated serializers, mainly for object shapes (the common fastify response case), without touching date/time serialization. Three changes, all in `index.js`:

1. **Codegen-time literal folding.** The object-open brace, comma, and property key were appended as separate statements (`json += JSON_STR_COMMA`, then `json += '"key":'`). They are now folded into a single pre-computed literal per property (`json += ',"key":'`). For objects with no guaranteed first property, a ternary on the comma flag picks between the `'{"k":'` and `',"k":'` variants, and a guard emits the bare `{` only when no property was emitted at all. `const` values fold all the way into a single literal.

2. **Nested-object inlining.** A non-recursive object schema that is not reached through a `$ref` is now serialized inline into the parent's `json` accumulator instead of through a generated function that builds and returns an intermediate string. Each schema object is inlined at most once (`context.inlinedSchemas`), so shared schema references cannot grow code size; recursive and `$ref`-shared schemas keep the function path with its reuse semantics.

3. **Small call-overhead wins.** `asInteger` is a plain function in the generated preamble instead of a `.bind()`-ed method, and booleans emit `(v ? 'true' : 'false')` instead of a function call.

A deliberate non-change: the fold only ever merges *compile-time constants*. Emitting `json += lit + call()` measures ~20-25% **slower** than two separate `+=` statements in hot loops (it leaves V8's string-append fast path), so that shape is never generated — this is documented in a comment, and is also why the array item loops keep their original form.

## Benchmarks

The host was noisy, so numbers come from an interleaved harness: baseline and patched builds in one process, measured round-robin with alternating order, ops/sec computed from user CPU time, reporting the median of paired ratios. Cases whose generated code is byte-identical to baseline are excluded (their true delta is zero).

| Case | Delta |
|---|---|
| Nested objects (2 levels) | +23% |
| Array of nested objects | +22% |
| Array of 3-property objects | +17% |
| Object with 15 mixed properties | +17% |
| Object with const string property | +11% |
| Simple object (no required) | +10% |
| Simple object with required | +7% |
| Scalar roots, leaf arrays, long strings, date/time formats | unchanged codegen, 0 within noise |

The repo suite (`benchmark/bench.js`) shows the same direction (e.g. `simple object` 5.70M → 7.55M ops/sec) but with much higher run-to-run variance.

## Verification

- All 491 tests pass; lint clean; TypeScript type tests pass.
- Old-vs-new output-equivalence fuzz over edge shapes (defaults containing quotes, nullable nested objects, `patternProperties`, `additionalProperties`, `$ref`/`$id`, multi-type arrays, keys needing escaping) produces byte-identical JSON.
- Root-level `date-time`/`date`/`time` schemas generate byte-identical code to baseline.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01J2ax464sB6TGcmKpAt3H7z


ivan-tymoshenko (approved):
Lgtm

index.js
```diff
@@ -26,7 +26,23 @@ const {
   asUnsafeString
 } = serializer
 
-const asInteger = serializer.asInteger.bind(serializer)
+const parseInteger = serializer.parseInteger
```
Eomm, line 29: can't we move this declaration with the deconstruction above?

index.js
```diff
@@ -26,7 +26,23 @@ const {
...
+// call is measurably faster than a bound method call in hot paths.
+function asInteger (i) {
+  if (Number.isInteger(i)) {
+    return '' + i
+  } else if (typeof i === 'bigint') {
+    return i.toString()
+  }
+  const integer = parseInteger(i)
+  // check if number is Infinity or NaN
+  if (integer === Infinity || integer === -Infinity || integer !== integer) {
```
ilteoood, line 41: instead of checking for `=== Infinity`, could we use `Number.isFinite`?

Tony133:
I think the tests need to be reviewed regarding this change

PR #870, opened 2026-08-14
fix: apply sibling keywords when dereferencing $ref

### LINKED ISSUE #866: Sibling required keyword is dropped when a property uses $ref to an external schema

haoku123:
Fixes #866

When a schema contains `$ref` alongside sibling keywords (e.g. `{ $ref: 'Foo', required: ['extra'] }`), `buildValue` resolved the reference and replaced the local schema wholesale, silently discarding the siblings. Objects missing a sibling-required field were serialized as valid.

**Solution**

Added `resolveRefLocation()`: schemas containing `$ref` alone keep the direct-dereference fast path; schemas with siblings merge the resolved reference with the sibling keywords using the same `mergeLocations` path as `allOf` (required arrays are unioned by `@fastify/merge-json-schemas`). All eager `resolveRef` call sites in `buildInnerObject` and `buildArray` now route through it, so properties, array items and tuple items behave consistently.

**Tests**

- `ref external with sibling required keyword keeps both constraints`: missing sibling-required field throws.
- `ref external with sibling required keyword keeps ref constraints`: missing ref-required field still throws.

Full suite: 493/493 pass, lint clean.


Tony133:
Duplicate see #867, closing

PR #878, opened 2026-09-05
fix: escape schema values in generated code

mcollina:
## Summary

- stop embedding schema references directly into generated comments, template literals, validator calls, and regular-expression literals
- encode complete generated error literals, including required-property errors
- escape JSON Pointer tokens and URI-fragment characters in generated validator references
- reject active schema identifiers containing block-comment terminators when Ajv source generation is enabled
- add regression coverage across normal, debug/restore, standalone, and inline-validator modes

## Rationale

The README requires schema definitions to be trusted application code. This change preserves that trust model while ensuring schema-derived strings cannot accidentally change generated JavaScript or resolve to the wrong validator path.

## Validation

- `npm test` — 514 tests passed; 100% statement, branch, function, and line coverage
- `npm run lint`
- targeted tests on Node.js 20, 22, 24, and 26
- `npm pack --dry-run`
- `git diff --check`


index.js
```diff
@@ -371,7 +371,7 @@ function buildExtraObjectPropertiesSerializer (context, location, addComma, objV
       const propertyLocation = patternPropertiesLocation.getPropertyLocation(propertyKey)
 
       code += `
-        if (/${propertyKey.replace(/\\*\//g, '\\/')}/.test(key)) {
+        if (new RegExp(${JSON.stringify(propertyKey)}).test(key)) {
```
Tony133, line 455: This is inside the `for (const key of Object.keys(...))` loop, so the `RegExp` gets constructed once per key per pattern on every serialization. Was that intentional, or worth hoisting it to module scope next to the generated functions?

gurgunday (approved):
lgtm

PR #879, opened 2026-09-06
fix: drop unmatched properties when additionalProperties is false

MaxFreedomPollard:
## Problem

A schema that combines `patternProperties` with `additionalProperties: false` serializes every property the patterns did not match, instead of ignoring it.

```js

const stringify = build({
  type: 'object',
  properties: { nickname: { type: 'string' } },
  patternProperties: { num: { type: 'number' } },
  additionalProperties: false
})

stringify({ nickname: 'nick', matchnum: 3, leaked: 'secret' })
// ACTUAL:   {"nickname":"nick","matchnum":3,"leaked":"secret"}
// EXPECTED: {"nickname":"nick","matchnum":3}
```

Deleting the `additionalProperties: false` line from that same schema produces the correct `{"nickname":"nick","matchnum":3}`, so writing the stricter setting is what widens the output. README "Additional properties" states the intended rule: "If *additionalProperties* is not present or is set to `false`, every property that is not explicitly listed in the *properties* and *patternProperties* objects will be ignored".

This matters beyond a formatting difference. `additionalProperties: false` is the way a response schema is written to guarantee that only declared fields reach the client, and here it is the one spelling that does not hold.

## Root cause

`buildExtraObjectPropertiesSerializer` gates its `additionalProperties` branch on `additionalPropertiesSchema !== undefined` (`index.js:387` on `main`). A boolean `false` schema is not `undefined`, so it passes that guard, and it is not `=== true`, so it lands in the else branch that calls `buildValue(context, propertyLocation, 'value')` on it. `buildValue` serializes any boolean schema as `json += JSON.stringify(${input})` (`index.js:1306`), which emits the property verbatim.

`buildInnerObject` calls the function only when `schema.patternProperties || schema.additionalProperties` is truthy (`index.js:532`). With `additionalProperties: false` and no `patternProperties` that guard is false and the function never runs, which is why the leak requires a sibling `patternProperties` to surface and why `additionalProperties: false` on its own has always been correct.

Here is the generated loop body for the schema above, before the fix. The final three lines are the emitted `false` branch:

```js
for (const key of Object.keys(obj)) {
  if (key === "nickname" || obj[key] === undefined || typeof obj[key] === 'function' || typeof obj[key] === 'symbol') continue
  const value = obj[key]
  if (/num/.test(key)) {
    !addComma_0 && (addComma_0 = true) || (json += JSON_STR_COMMA)
    json += asString(key) + JSON_STR_COLONS
    json += asNumber(value)
    continue
  }
  !addComma_0 && (addComma_0 = true) || (json += JSON_STR_COMMA)
  json += asString(key) + JSON_STR_COLONS
  json += JSON.stringify(value)
}
```

## Fix

Exclude `false` from the guard, so no branch is emitted for it and an unmatched key simply falls off the end of the loop body and is dropped:

```js
if (additionalPropertiesSchema !== undefined && additionalPropertiesSchema !== false) {
```

`additionalProperties: true` still goes through its `JSON.stringify` fast path, object schemas still go through `buildValue`, and `additionalProperties: false` without `patternProperties` is untouched because that path never reached this function.

## Tests

Two cases added to `test/additionalProperties.test.js`, one with a declared `properties` map and one without, each asserting the exact output string and that it round-trips through `JSON.parse`. Both fail on `main` (`'{"str1":"a","leaked":"secret"}' == '{"str1":"a"}'`) and pass with the fix.

Verification run on Node 22.23.2, macOS arm64, from a clone of `main` at e02b5bb:

`npx c8 --100 --all --include=index.js --include='lib/**/*.js' --exclude=lib/schema-validator.js node --test --test-concurrency=2` gives 506/506 passing and holds 100% statements, branches, functions and lines on `index.js` and every file under `lib/`, which the `--100` threshold enforces. `npm run lint` is clean and `npm run test:typescript` passes 14/14 assertions. No type change is needed: `types/index.d.ts:136` already declares `additionalProperties?: Schema | boolean`.

`npm run benchmark` completes with no errors. I am not reporting a delta from it, because no scenario in `benchmark/` uses `patternProperties`, so the changed branch is never generated there and any difference in those numbers would be machine noise rather than an effect of this patch.

No documentation change: README already describes the behaviour this restores.

---

## Checklist


mcollina (approved):
lgtm

gurgunday (approved):
lgtm

Tony133 (approved):

PR #880, opened 2026-09-08
perf: reduce concatenation for optional properties

gurgunday:
* Combines the comma and property name into one string when the schema is compiled

This results in one fewer concatenation per subsequent supplied property, repeated across every object

```text
   Existing benchmark                                           Node 24    Node 26
  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━  ━━━━━━━━━
   object with properties of different types                     +16.1%     +14.1%
  ───────────────────────────────────────────────────────────  ─────────  ─────────
   short array of objects with properties of different types      +1.4%      +8.8%
```


mcollina (approved):
lgtm

Tony133 (approved):
lgtm

PR #884, opened 2026-09-15
fix: apply the else branch when a schema has if and else but no then

MaxFreedomPollard:
## Problem

A schema that pairs `if` with `else` and no `then` has its `if` and `else` silently ignored, so every property declared only under `else` is dropped from the output.

```js

const stringify = build({
  type: 'object',
  properties: {
    kind: { type: 'string' }
  },
  if: {
    type: 'object',
    properties: {
      kind: { type: 'string', enum: ['foobar'] }
    }
  },
  else: {
    type: 'object',
    properties: {
      bar: { type: 'string' }
    }
  }
})

console.log(stringify({ kind: 'other', bar: 'value' }))
// ACTUAL:   {"kind":"other"}
// EXPECTED: {"kind":"other","bar":"value"}
```

Adding a `then` to that same schema is enough to make it work. With `then: { type: 'object', properties: { baz: { type: 'string' } } }` present, the identical input serializes to `{"kind":"other","bar":"value"}`, which isolates the absent `then` as the cause rather than anything about the `else` subschema. `then` is not required for `else` to apply: README.md:394 says the library supports the `if/then/else` jsonschema feature and points at [the ajv documentation](https://ajv.js.org/keywords.html#ifthenelse), and JSON Schema draft-07 section 6.6.3 makes `else` depend only on `if`. The schema is accepted, `build()` succeeds and nothing is reported, so the only symptom is missing fields in the response body.

## Root cause

`buildValue` (index.js:1313) is the only caller of `buildIfThenElse`, and its guard at index.js:1333 requires both keywords:

```js
if (schema.if && schema.then) {
  return buildIfThenElse(context, location, input)
}
```

With `then` absent the whole block is skipped and the schema is serialized as if `if` and `else` were not there, which is why the two-keyword form loses data while the three-keyword form is fine.

The guard cannot simply be widened on its own. `buildIfThenElse` destructures `then` at index.js:1249 and then merges it unconditionally: index.js:1264 builds a `then` location that does not exist, index.js:1271 calls `context.mergedSchemasIds.set(thenSchema, ...)` with `undefined` as the key, and index.js:1273 hands that location to `mergeLocations`. Widening only the guard turns the dropped property into `TypeError: Cannot read properties of undefined (reading '$ref')` at build time.

`build(schema, { mode: 'debug' })` on main for the repro schema emits the serializer below. The surrounding module preamble is left out; the function itself is verbatim, generated indentation included. There is no branch, and `bar` never appears:

```js
      function anonymous0 (input) {
        const obj = (input && typeof input.toJSON === 'function')
    ? input.toJSON()
    : input
  
        let json = ''

        json += JSON_STR_BEGIN_OBJECT

          const value_kind_1 = obj["kind"]
          if (value_kind_1 !== undefined) {
            json += "\"kind\":"
            
        if (typeof value_kind_1 !== 'string') {
          if (value_kind_1 === null) {
            json += JSON_STR_EMPTY_STRING
          } else if (value_kind_1 instanceof Date) {
            json += JSON_STR_QUOTE + value_kind_1.toISOString() + JSON_STR_QUOTE
          } else if (value_kind_1 instanceof RegExp) {
            json += asString(value_kind_1.source)
          } else {
            json += asString(value_kind_1.toString())
          }
        } else {
          json += asString(value_kind_1)
        }
        
          }

    json += JSON_STR_END_OBJECT
  
        return json
      }
```

The emitted code contains zero occurrences of `validator.validate` and zero occurrences of `"bar"`.

## Fix

Two edits in index.js. The guard now accepts either branch keyword:

```js
if (schema.if && (schema.then || schema.else)) {
```

and the `then` merge in `buildIfThenElse` becomes conditional, with the true branch defaulting to the root schema:

```js
let thenMergedLocation = rootLocation
if (thenSchema !== undefined) {
  // existing merge, unchanged
}
```

Defaulting to `rootLocation` is the behaviour draft-07 asks for: when `then` is absent the true branch adds no keywords, so a matching instance serializes with the schema minus `if`/`then`/`else`, which is exactly the location the existing `if (!elseSchema)` path already uses for its false branch.

The merge body is the same code at one more level of indentation. The `if (!elseSchema)` early return at index.js:1279 is untouched and is now only reachable when `then` exists, so the `then`-only and `then` plus `else` shapes generate byte-identical code to main. A schema with `if` and neither branch keyword still falls through as before, and so does `if` plus `else: false`, since `false` is falsy in the widened guard. A truthy but empty `else` (`true` or `{}`) now takes the `buildIfThenElse` path where main skipped it; the serialized output is identical to main, and the generated code gains the `validator.validate` call and a second serializer function.

One further shape does change, and it is the same defect. `if` plus `then: false` plus an `else` object was skipped on main for exactly the reason the repro was, because `then: false` is falsy. It now enters `buildIfThenElse`, and the repro input serializes as `{"kind":"other","bar":"value"}` where main gave `{"kind":"other"}`. Nothing throws, and an instance matching `if` still gives `{"kind":"foobar"}` on both. `else` applies whenever `if` fails, whatever `then` says.

## Tests

One test added to test/if-then-else.test.js, four assertions. It builds the repro schema and covers both directions: an instance that fails `if` must keep the `else` property (`{"kind":"other","bar":"value"}`), and an instance that matches `if` must serialize with the root schema alone (`{"kind":"foobar"}`, with `bar` correctly absent). Each output is checked as an exact string and round-tripped through `JSON.parse`. All 11 existing tests in that file supply a `then`, so this shape had no coverage.

On main the new test fails with `error: '{"kind":"other"}' == '{"kind":"other","bar":"value"}'`. With the fix it passes.

Verified on Node v22.23.2, macOS arm64, against a clone of main at 99bc4e858e5a1d77240e557182204adaa859a046. `npm run test:unit` reports 515 tests, 515 pass, 0 fail in about 3 s, and the c8 table under the `--100` threshold still reports 100% statements, branches, functions and lines for index.js and for all five files under lib/. The new `thenSchema !== undefined` branch is covered in both directions, the existing tests taking the true side and the new test the false side. `npm run lint` exits 0 with no output. `npm run test:typescript` passes 14 assertions.

Three things this does not change. No type change is needed, since the fix is internal to code generation and the public signature is untouched. No documentation change is needed, since README.md:394 already defers to the ajv semantics this restores. `npm run benchmark` completes, but no delta is worth quoting because no benchmark scenario builds an `if`/`else` schema, so any difference would be machine noise.

Standalone mode was checked by hand on the repro schema: with the fix it returns `{"kind":"other","bar":"value"}` for the `else` branch and `{"kind":"foobar"}` for the true branch, but the added test uses the normal build path only.

---

## Checklist

- [ ] documentation is changed or added: not needed, the README already describes the `if/then/else` support this restores and adds no caveat about `then` being mandatory


mcollina (approved):
lgtm

Tony133 (approved):
lgtm

